PR #1514, opened 2014-11-04
Add name to access token response

wjrsimpson:


rcpeters:
@wjrsimpson can you recall which trello ticket this was with?

PR #1759, opened 2015-01-12
Integration tests changes

amontenegro:


amontenegro:
Hey @rcpeters,

I found some test that where failing, so, i fixed them; and I also did the following changes:

1) Took some time to move all tests to the same folder, so, it will be easier to run them in the integration server (When I configure it to run the integration test)
2) Add some system out's to PopulateOAuthSignInCodeIntegrationTest and T2OrcidOAuthApiAuthorizationCodeIntegrationTest, beacuse, randomly, some of their tests fail, so, at the begging of each test Im printing the test name, so, when one of this tests fails, we can see which ones runs first, and, identify if they deleted any info that is still needed by the tests that fails; as soon as I identify the problem I will remove those system out's

Thanks

PR #1761, opened 2015-01-12
When using a client credential scope in a 3 step auth, we should throw a...

amontenegro:
...n error


amontenegro:
https://trello.com/c/1qae5Jm4/1834-asking-for-read-public-using-3-leg-oauth-gives-you-token-with-all-scopes

PR #1767, opened 2015-01-15
Removed unused json config classes

wjrsimpson:


rcpeters:
I wonder how much easier the Jackson upgrade will be now?

PR #1770, opened 2015-01-20
Fix reseting client secret

amontenegro:


amontenegro:
https://trello.com/c/BnxoCWWT/1842-public-client-user-orcid-org-0000-0002-5973-3490-can-t-reset-client-secret-for-app-0zc6kcb3yxh5w8eh

PR #1779, opened 2015-01-23
Updated putCode path in freemarker template and javascript

wjrsimpson:
...to match changes in schema and Java.


wjrsimpson:
https://trello.com/c/bflmUk9s/1844-messaging-change-notifications-xsd-so-that-putcode-is-an-attribute-instead-of-an-element

PR #1785, opened 2015-01-28
use temporary repo for spring security oauth

rcpeters:


rcpeters:
@wjrsimpson Please run unit and integration test for this patch. Merge when done.

PR #1789, opened 2015-01-30
Scope rename and black box tests for notifications

wjrsimpson:


amontenegro:
Hey @wjrsimpson, 

Do we need to run any db update after deploying this to QA and Sandbox? since the card say you have to add that scope to premium members, it looks we need to add the scope to the existing members https://trello.com/c/FgblktKZ/1860-messaging-change-scope-name-and-add-to-premium-members 

So, should we run any query/cli to update the existing clients? 

Thanks

wjrsimpson:
@amontenegro Hi, Angel. There is a database function in notifications_part2.xml that adds the new scope to premium members. It is not as flexible as your script for adding scopes, but it is run automatically by liquibase. So, we do not need to run a db update. W

amontenegro:
Cool! Thanks will!

PR #1792, opened 2015-02-02
Removing code for the old way of doing funding external identifiers

amontenegro:


amontenegro:
https://trello.com/c/OFfWFFgj/1810-remove-unused-funding-external-identifiers-code

PR #1798, opened 2015-02-06
Search and indexing fixes

wjrsimpson:


wjrsimpson:
https://trello.com/c/nB9pkwug/1876-new-records-that-fail-indexing-cause-error-in-public-api

PR #1803, opened 2015-02-10
Fundings where grouped by ext id type, but, they are always grant_type, ...

amontenegro:
...so, we need to group them by url or by value


amontenegro:
https://trello.com/c/95igrsIl/1877-groupings-funding-items-without-grant-numbers-are-grouping-incorrectly

PR #1805, opened 2015-02-12
Mapping funding to profile funding entity

amontenegro:


amontenegro:
We dont need a pull request for this

PR #1806, opened 2015-02-12
Mapping funding to profile funding entity

amontenegro:
Hey @wjrsimpson,

Do you want to take a look before merging this? I also tried to change all references to fundingX in the xsd, so, for example

"funding:fundingExternalIdentifiers" is now "funding:externalIdentifiers"

And 

"funding:fundingContributors" is now "funding:contributors"

Thanks


wjrsimpson:
@amontenegro This looks great!

PR #1824, opened 2015-02-18
Fixing linebreak behavior in Public Profile

hexplus:


rcpeters:
thanks for being so quick!

PR #1825, opened 2015-02-19
Mapping record message and core objects

amontenegro:


amontenegro:
Hey @wjrsimpson @rcpeters 

Please take a look at this: 

I created a class org.orcid.pojo.FundingExternalIdentifier on orcid-core, that one will be used to map the json in the db, so, all messages that comes from the UI (message) or the API (record) should be translated to it.... so org.orcid.pojo.FundingExternalIdentifier have some utility methods that will translate from one object to the other.

Take a look at it and let me know what you think.

Also, just as a note, I found a weird issue... on org.orcid.jaxb.model.record.Funding, I had to change the name of the Amount property, from amount to fundingAmount, because, for a reason that I don't understand yet, when the converter MapperFacadeFactory executes the mapping from ProfileFundingEntity to Funding, it fail to convert the BigDecimal numeric amount (from ProfileFundingEntity) to the amount.content ... the other way it works fine (from Amount to ProfileFundingEntity), and, I reviewed all mappings and naming conventions and they looked ok, so, I dont know why it failed.... and magically, after change the property name, the mapping worked fine.

BTW, there are a couple of tests failing on JpaJaxbEntityAdapterToOrcidProfileTest, but it is something related with works contributors, so, I think it will be better to fix it in other branch.

Thanks

wjrsimpson:
@amontenegro Very nice!
It looks like FundingEntity also has a property called amount, and .byDefault() is overriding your mapping because the property name matches Funding.amount.
If you move classMap.byDefault(); up to line 114, then it should be OK.
(However, maybe FundingEntity.amount still needs to be mapped to something? I don't know fundings well enough to say.)

amontenegro:
@wjrsimpson 

WOW! good catch! I never thought on that.
And, the amount on FundingEntity is not used anymore, we are using the BigDecimal numericAmount since some months ago, so, I just created this card to remove that field https://trello.com/c/WxbJLlCt/1884-remove-the-amount-field-from-the-profilefundingentity-since-we-now-use-the-bigdecimal-numericamount since it is really confusing.

Thanks

amontenegro:
Ok @wjrsimpson I rolled back the name of the amount property.

There is still an issue with JpaJaxbEntityAdapterToOrcidProfileTest but im addressing it in a separate branch

Thanks

amontenegro:
@wjrsimpson Fixed the error, so, I merged it :D

PR #1842, opened 2015-03-03
Fix notification start

wjrsimpson:


wjrsimpson:
https://trello.com/c/ClfAdSde/1868-notification-email-messages-are-not-being-sent

PR #1845, opened 2015-03-03
Fix api validation

wjrsimpson:


wjrsimpson:
https://trello.com/c/zDpb1XgN/1867-add-activities-message-doesn-t-have-the-url-to-actually-add-the-item

PR #1877, opened 2015-03-11
Api return404 on not found

amontenegro:


amontenegro:
https://trello.com/c/bLmd3zXG/1820-api-returns-a-500-error-when-a-record-is-not-found

PR #1878, opened 2015-03-11
Done

amontenegro:


amontenegro:
https://trello.com/c/DyemFbQO/1798-when-creating-a-record-with-the-api-if-claimed-true-claimed-is-set-the-record-immediately-becomes-public

rcpeters:
merge monster

PR #1885, opened 2015-03-13
Change how group profile is loaded

amontenegro:


orcid-core/src/main/java/org/orcid/core/manager/impl/ProfileEntityManagerImpl.java
```diff
@@ -82,7 +84,8 @@
     private JpaJaxbWorkAdapter jpaJaxbWorkAdapter;
 
     @Override
-    public ProfileEntity findByOrcid(String orcid) {
+    @Cacheable(value = "profile-entity", key = "#orcid.concat('-').concat(#lastModified)")
```
amontenegro: Hey @rcpeters,

https://trello.com/c/T4VGtkzU/1702-change-the-way-groupprofile-is-loaded

Please check this, now, when we get the group info from the ClientDetailsEntity class, we just get an orcid string, then, we call this to get the ProfileEntity, which, hopefully will be already cached.

Thanks

amontenegro:
https://trello.com/c/T4VGtkzU/1702-change-the-way-groupprofile-is-loaded

rcpeters:
@amontenegro Sorry, but we shouldn't use the cacheable annotation any longer. See /orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidProfileCacheManagerImpl.java

amontenegro:
@rcpeters 

Done, now we have a ProfileEntityCacheManager that will cache profile entities, and, all classes that needs a profile entity use it instead of directly calling the ProfileEntityManager

Thanks

PR #1889, opened 2015-03-16
The method should be send in a transaction to be able to fetch the lazy ...

amontenegro:
...client details object


amontenegro:
https://trello.com/c/x3TdSs0l/1911-500-error-when-creating-records-for-some-clients

PR #1902, opened 2015-03-20
Fixing issue where grant number was not displayed if the url was empty

amontenegro:


amontenegro:
https://trello.com/c/oaHajefE/1913-grant-numbers-are-being-omitted-when-a-funding-item-is-added-via-the-api

PR #1903, opened 2015-03-23
Handle case when there is no primary email.

wjrsimpson:
Continue to next record if there is any runtime error.


wjrsimpson:
https://trello.com/c/QZ3GJZ1I/1925-messaging-digest-email-sending-fails-when-the-record-has-no-primary-email

PR #1852, opened 2015-03-05
activate Czech translation in production

helix84:
Please merge #1850 first. Thank you.

PR #1904, opened 2015-03-23
Czech translation reviewed

helix84:


helix84:
PR for activation: #1852

PR #1905, opened 2015-03-23
Update put code to path

amontenegro:


amontenegro:
https://trello.com/c/HJ7JNsHj/1921-change-the-put-code-on-the-new-api-so-it-includes-the-path-to-the-activity

amontenegro:
Hey @rcpeters check this out.

So, for the new API, every activity we get will have a put code that looks like "/{orcid}/{activity}/{putCode}".

So far it works for GET, but not for POST/PUT/DELETE, so, for them you should use the actual put code; let me know if you would also want to enable a POST/PUT/DELETE that works with the new format.

Thanks

rcpeters:
:+1:

PR #1920, opened 2015-04-03
fix blank other names, which isn't valid

rcpeters:


rcpeters:
https://trello.com/c/j18aZdFY/1944-web-ui-don-t-allow-saving-empty-other-names-to-db

PR #1928, opened 2015-04-15
Fix error where the chineese language code was not correctly parsed

amontenegro:


amontenegro:
https://trello.com/c/q28XPxYN/1949-work-language-doesn-t-display-in-ui-if-using-zh-cn-or-zh-tw

amontenegro:
Hey @rcpeters 

Please check this one, the first problem was that we where looking for zn, but it should be zh... but then, when we store in the DB, we save it like zh_TW and zh_CN, but the java.util.Locale toString() return zh_tw and zh_cn, so, we have to ways of fixing this...

1) By updating all db entries and updating our language code generator, so, it return everything lower cased
2) By changing the code as in this pull request

Let me know if you are ok with the fix in this pull request, or, you want me to update the database entries.

Thanks

PR #1931, opened 2015-04-16
Fixing error where all fundings display 'Grant number' for the ext id nu...

amontenegro:
...mber, no matter the funding type


amontenegro:
https://trello.com/c/1ppkln4i/1937-award-and-contract-numbers-incorrectly-display-as-grant-numbers

PR #1954, opened 2015-04-30
Oauth skip authorization when token already exists

amontenegro:


amontenegro:
https://trello.com/c/xFYKTXjG/1942-skip-authorization-prompt-and-return-the-authorization-code-if-long-lived-token-exists

PR #1957, opened 2015-05-06
Work publication date fix

hexplus:


hexplus:
https://trello.com/c/lsubzlBV/1817-in-ui-can-add-works-with-a-day-or-month-but-no-year

PR #1960, opened 2015-05-06
Fixing external identifiers add link visibility when adding more EI

hexplus:


hexplus:
https://trello.com/c/xW2NdSHK/1974-peer-review-keep-just-the-last-link-to-add-external-ids

PR #1963, opened 2015-05-06
Fixing show sources option in Fundings - Public Profile

hexplus:


hexplus:
https://trello.com/c/XNqIeFsS/1976-show-sources-for-fundings-in-public-profile-doesn-t-work

PR #1965, opened 2015-05-07
Fixing alternate URL rendering

hexplus:


hexplus:
https://trello.com/c/XgaIvZb3/1914-funding-alternative-urls-added-via-the-api-are-rendered-incorrectly

PR #1967, opened 2015-05-07
Fixing Bulk edit missing title

hexplus:


hexplus:
https://trello.com/c/VxEWXJhc/209-bulk-edit-header-is-missing-from-the-drop-down-in-production

PR #1969, opened 2015-05-07
Fix buttons format

hexplus:


hexplus:
https://trello.com/c/MTOlERXY/1825-email-verification-pop-up-close-and-cancel-buttons-don-t-look-like-buttons

PR #1971, opened 2015-05-07
Fixing test that should not work that way anymore

amontenegro:


amontenegro:
Super small fix on integration test, so, it is now and master and will move to qa in the next release

PR #1975, opened 2015-05-08
Extracting the source id and overwriting to the input orcidprofile be…

nitw-shobhit:
…fore persist
https://trello.com/c/jJukgGLt/1960-do-not-allow-funding-items-to-be-added-with-bad-sources

### DIFF
--- a/orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidProfileManagerImpl.java
+++ b/orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidProfileManagerImpl.java
@@ -423,10 +423,8 @@ public class OrcidProfileManagerImpl implements OrcidProfileManager {
     private void addSourceToWorks(OrcidWorks orcidWorks, String amenderOrcid) {
         if (orcidWorks != null && !orcidWorks.getOrcidWork().isEmpty() && amenderOrcid != null) {
             for (OrcidWork orcidWork : orcidWorks.getOrcidWork()) {
-                if (orcidWork.getSource() == null || StringUtils.isBlank(orcidWork.getSource().retrieveSourcePath())) {
-                    Source source = createSource(amenderOrcid);
-                    orcidWork.setSource(source);
-                }
+                Source source = createSource(amenderOrcid);
+                orcidWork.setSource(source);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidProfileManagerImpl.java
```diff
@@ -423,10 +423,8 @@ private void addSourceToWorks(OrcidProfile orcidProfile, String amenderOrcid) {
     private void addSourceToWorks(OrcidWorks orcidWorks, String amenderOrcid) {
         if (orcidWorks != null && !orcidWorks.getOrcidWork().isEmpty() && amenderOrcid != null) {
             for (OrcidWork orcidWork : orcidWorks.getOrcidWork()) {
-                if (orcidWork.getSource() == null || StringUtils.isBlank(orcidWork.getSource().retrieveSourcePath())) {
```
rcpeters: @nitw-shobhit I'm pretty sure this is going to cause previous valid sources to be overwritten. See OrcidProfileManagerImpl.updateOrcidProfile. @wjrsimpson @amontenegro do you agree?

nitw-shobhit: @rcpeters The source value that we  put in the XML may be valid, but it cannot be any other than the client who is trying to call the API. So it doesn't matter if the source is valid or not. It has to be overwritten with the client ID.
There may be a case that the XML already contains the client id and we try to overwrite it again by the same value, but that doesnt increase the complexity of the code.

rcpeters: @nitw-shobhit Please do the following for 1.2. You'll run into unusual implicit behavior implicit behavior. 

1. Create a sample user with a self added work.

2. Post a new work work with a client.

3. Update the work just posted.

4. Inspect the sources for all works.

rcpeters: @wjrsimpson @amontenegro we need a unit test that covers this situation.

nitw-shobhit: I am trying the scenario.

amontenegro:
Im working on the unit test to verify that the sources are not overwritten after the update

rcpeters:
@nitw-shobhit  How is this ticket going?

nitw-shobhit:
@rcpeters  I will have it in some time. I know the solution but just checking the risks with it.

PR #1978, opened 2015-05-12
Add activity dont change source on existing

amontenegro:


amontenegro:
Hey @rcpeters,

Here is the unit test to check that adding new activities doesn't affect the source of other activities.

Thanks

PR #1987, opened 2015-05-22
update to latest, which parses keys differently

rcpeters:


rcpeters:
@lizkrznarich this should allow any character for bibtex keys except " ,{}=".

PR #1990, opened 2015-05-26
Fix deprecate account issue

amontenegro:


amontenegro:
https://trello.com/c/uqYp024B/1996-please-deprecate-0000-0003-1583-7850-into-0000-0003-0900-4326

PR #1991, opened 2015-05-26
Fixing language files key, it was causing an error in registration form

hexplus:


hexplus:
https://trello.com/c/CXWx6Yeb/1995-error-on-registering-page

PR #1993, opened 2015-05-27
Changing button color in Developer Tools

hexplus:


hexplus:
https://trello.com/c/lUMZeCFE/1998-reset-client-secret-button-is-not-the-right-color

PR #2001, opened 2015-05-29
Comma separated other names fix

nitw-shobhit:
https://trello.com/c/OWfNd3Ig/2004-also-know-as-field-missing-commas-in-public-view


nitw-shobhit:
Weird.. I created another branch and pushed to it.. But how come it is in the same pull request .

PR #2008, opened 2015-06-08
Count unique dois the right way

amontenegro:


amontenegro:
https://trello.com/c/VO0aiaVs/2012-fix-the-way-we-count-unique-dois-on-works

PR #2014, opened 2015-06-11
Moved the 'extraction and format orcid id' part of the code to script and modified to accept userid in various other formats

nitw-shobhit:
https://trello.com/c/pLIvDjfq/1958-let-users-log-in-with-the-full-orcid-id-i-e-http-orcid-org-1234-1111-1111-111x

### DIFF
--- a/orcid-web/src/main/webapp/static/javascript/script.js
+++ b/orcid-web/src/main/webapp/static/javascript/script.js
@@ -397,8 +397,7 @@ $(function() {
                                         {
                                             url : baseUrl + 'signin/auth.json',
                                             type : 'POST',
-                                            data : $('form#loginForm')
-                                                    .serialize(),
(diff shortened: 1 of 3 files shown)


orcid-web/src/main/webapp/static/javascript/script.js
```diff
@@ -397,8 +397,7 @@ $(function() {
                                         {
                                             url : baseUrl + 'signin/auth.json',
                                             type : 'POST',
-                                            data : $('form#loginForm')
-                                                    .serialize(),
+                                            data : 'userId=' + encodeURIComponent( orcidLoginFitler( $('input[name=userId]').val())) + '&password=' + $('input[name=password]').val(),
```
rcpeters: @nitw-shobhit password would also have to be url encoded correct?

nitw-shobhit: Yes. Correct. password can contain escape characters. I ll update and commit

PR #2019, opened 2015-06-15
Ensure client id uri is always plain http in v2 API.

wjrsimpson:


wjrsimpson:
https://trello.com/c/9ElFFenE/2030-client-id-length-error

PR #2021, opened 2015-06-15
Removing duplicated test dependency

amontenegro:


amontenegro:
Just removing a duplicated test dependency that I found in the root pom

PR #2033, opened 2015-06-19
Updated value of property to match with the bean object

nitw-shobhit:
https://trello.com/c/PVCzRj2I/2040-claimwaitperiod-property-value-mismatch-in-staging-persistence-vs-bean-mapping


rcpeters:
This property doesn't need to match. 1 day wait period staging makes testing faster?

nitw-shobhit:
@rcpeters It gets changed to 1 when I debug. With value 10, it worked fine.

rcpeters:
Yes, staging is 1 and production is 10. If you want a different value for your local environment you can overwrite with `-Dorg.orcid.core.claimWaitPeriodDays=10` in your Tomcat VM arguments, but the default staging is value is 1.

PR #2043, opened 2015-06-25
Revert "Updated the minOccur value to 1 in XSD"

rcpeters:
Reverts ORCID/ORCID-Source#2042


rcpeters:
@nitw-shobhit sorry reverting your changes. Although we don't want to new activities without work title, type and identifier there are works in the system without identifiers, which means with these changes we will be returning invalid xml. We have to do server side validation not XSD validation for identifiers.

PR #2045, opened 2015-06-25
Add title identifier validation to work

nitw-shobhit:
https://trello.com/c/w7eIXV3W/2056-2-0-error-can-add-works-without-a-title-or-without-identifier

### DIFF
--- a/orcid-core/src/main/java/org/orcid/core/manager/impl/ProfileWorkManagerImpl.java
+++ b/orcid-core/src/main/java/org/orcid/core/manager/impl/ProfileWorkManagerImpl.java
@@ -21,12 +21,15 @@ import java.util.Date;
 
 import javax.annotation.Resource;
 
+import org.apache.commons.lang3.StringUtils;
 import org.orcid.core.adapter.JpaJaxbWorkAdapter;
+import org.orcid.core.exception.OrcidValidationException;
 import org.orcid.core.manager.OrcidSecurityManager;
 import org.orcid.core.manager.ProfileWorkManager;
 import org.orcid.core.manager.SourceManager;
 import org.orcid.jaxb.model.message.Visibility;
 import org.orcid.jaxb.model.record.Work;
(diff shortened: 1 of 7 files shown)


orcid-core/src/main/java/org/orcid/core/manager/impl/ProfileWorkManagerImpl.java
```diff
@@ -189,6 +193,18 @@ public Work createWork(String orcid, Work work) {
...
 
+    private void validateWork(Work work) {
+        WorkTitle title = work.getWorkTitle();
+        if (title == null || title.getTitle() == null || StringUtils.isEmpty(title.getTitle().getContent())) {
+            throw new OrcidValidationException("Invalid Title: title cannot be null nor emtpy");
+        }
+
+        if (work.getWorkExternalIdentifiers() == null || work.getWorkExternalIdentifiers().getWorkExternalIdentifier() == null
+                || work.getWorkExternalIdentifiers().getWorkExternalIdentifier().isEmpty()) {
+            throw new OrcidValidationException("Invalid work: Works added using message version 1.2_rc5 or greater must contain at least one external identifier");
```
rcpeters: throw a new style exception
throw new OrcidDuplicatedActivityException(localeManager.resolveMessage("api.error.duplicated"));

amontenegro: @nitw-shobhit dont hardcode the exception, add a i18n key and the locale manager to get the error message

nitw-shobhit: @amontenegro I took the code from 1.2. 
Anyway I am gonna check if production has bad work data. If yes then i ll update the code. else we can ignore this change and update the XSD as @rcpeters suggested.

PR #2046, opened 2015-06-25
Modified the error message for the code 9020

nitw-shobhit:
https://trello.com/c/qZKVNp7f/2060-wrong-error-message-for-orcidvalidationexception


orcid-core/src/main/resources/i18n/api_en.properties
```diff
@@ -61,7 +61,7 @@ apiError.9018.userMessage=The ORCID record is locked.
 apiError.9018.moreInfo=http://members.orcid.org/api/api-error-codes
 apiError.9019.userMessage=The put code in the URL did not match that in the body of the message.
 apiError.9019.moreInfo=http://members.orcid.org/api/api-error-codes
-apiError.9020.userMessage=The put code in the URL did not match that in the body of the message.
+apiError.9020.userMessage=Invalid incoming message.
```
rcpeters: It's an improvement, but invalid incoming message isn't specific enough. How will a developer know that to fix? I suspect we need more then one of these.

wjrsimpson: (We do also include the exception message in developerMessage in the error response, though of course that will be in English. The userMessage is for the client to display to the user, so I suspect we might not want too much info in there unless it is something the user can fix themselves. I also think it might be helpful to our support team if the message said something to indicate that the user should complain to the client application support to team, not to ORCID!)

nitw-shobhit: The amount of error messages and Custom Exceptions is certainly not enough.

nitw-shobhit: @rcpeters A question. For the other issue regarding 2.0 Work validation, I created a couple of custom exceptions and associated them with 2 new error messages. Would you have rather preferred using the existing exception with a new messg? I need to fix another issue somewhat similar and I am in a dilemma now.

rcpeters: I would like to see as much detail as possible to be provided to the API user. Custom exceptions with unique 9000 codes sounds good to me.

PR #2047, opened 2015-06-25
Added a check when client adds an extra primary email

nitw-shobhit:
https://trello.com/c/p671ueAU/2005-client-apps-can-overwrite-user-privacy-setting-for-email-addresses

### DIFF
--- a/orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidJaxbCopyManagerImpl.java
+++ b/orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidJaxbCopyManagerImpl.java
@@ -342,9 +342,14 @@ public class OrcidJaxbCopyManagerImpl implements OrcidJaxbCopyManager {
 				}
 			}
 		}
+		//Set primary = false for each remaining new email.
 		if(updatedContactDetails != null) {
-			allEmails.addAll(updatedContactDetails.getEmail());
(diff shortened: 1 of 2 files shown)


orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidJaxbCopyManagerImpl.java
```diff
@@ -342,9 +342,14 @@ private void copyUpdatedEmails(ContactDetails existingContactDetails, ContactDet
 				}
 			}
 		}
+		//Set primary = false for each remaining new email.
 		if(updatedContactDetails != null) {
-			allEmails.addAll(updatedContactDetails.getEmail());
+			for(Email newEmail : updatedContactDetails.getEmail()) {
```
rcpeters: Did we test when a client creates a ORICD?

nitw-shobhit: @rcpeters This code is called only when updating the bio. Is there a particular flow that you have in mind? I can try.

rcpeters: Wanted to make sure we tested a client creating a record. http://members.orcid.org/api/tutorial-create-new-record-using-curl-12-and-earlier

PR #2048, opened 2015-06-25
Setting the migrated value to true and the added to profile date when…

amontenegro:
… requiered


amontenegro:
https://trello.com/c/Xv6v65yY/2065-when-adding-updating-a-work-using-api-1-2-the-migrated-field-in-profile-funding-remains-empty

PR #2064, opened 2015-07-06
Remove user registration dependency on solr being up.

rcpeters:


orcid-core/src/main/java/org/orcid/core/manager/impl/RegistrationManagerImpl.java
```diff
@@ -114,8 +114,6 @@ public OrcidProfile createMinimalRegistration(OrcidProfile orcidProfile) {
         List<Email> emails = minimalProfile.getOrcidBio().getContactDetails().getEmail();
         for(Email email : emails)
             emailManager.addSourceToEmail(email.getValue(), sourceId);
-        //Index new profile
-        orcidProfileManager.processProfilePendingIndexingInTransaction(orcidProfile.getOrcidIdentifier().getPath());
```
rcpeters: @wjrsimpson with our current system the profile is marked pending and the scheduler will index anyway (if solr is up). I believe these two lines are legacy dependency from when we din't have the concept of "PENDING" and "REINDEX". Please LMK what you think.

PR #2072, opened 2015-07-09
Standardize managers

amontenegro:


amontenegro:
https://trello.com/c/ycteVZ1V/2080-cern-reporting-500-error-when-calling-same-record-twice-with-different-tokens

PR #2074, opened 2015-07-10
fix issue with primary record not being recorded

rcpeters:


amontenegro:
Hey @rcpeters looks good to me!

PR #2117, opened 2015-07-23
Fixing npe error when trying to read the citation type when it was null

amontenegro:


amontenegro:
https://trello.com/c/KMI7I9BA/2110-can-t-add-a-citation-to-an-existing-works-in-the-ui

PR #2121, opened 2015-07-27
Added a property to allow all local URI

nitw-shobhit:
https://trello.com/c/prnCy2MT/2124-can-t-add-localhost-as-a-redirect-uri-with-public-api-clients


rcpeters:
Nice change.

PR #2127, opened 2015-07-30
Changes on relationship

amontenegro:


amontenegro:
https://trello.com/c/Snn6YjRD/2119-set-relationship-to-part-of-for-works-where-external-id-type-is-isbn-and-work-type-is-publication-conference-paper

PR #2155, opened 2015-08-07
Changed return MinimizedWorkEntity to WorkEntity, Updated Mapper

nitw-shobhit:
https://trello.com/c/KJAqTh4x/2163-v2-0-verify-all-activities-in-the-summary-have-a-created-and-modified-date

### DIFF
--- a/orcid-core/src/main/java/org/orcid/core/manager/impl/WorkManagerImpl.java
+++ b/orcid-core/src/main/java/org/orcid/core/manager/impl/WorkManagerImpl.java
@@ -222,8 +222,8 @@ public class WorkManagerImpl implements WorkManager {
     @Override
     @Cacheable(value = "works-summaries", key = "#orcid.concat('-').concat(#lastModified)")
     public List<WorkSummary> getWorksSummaryList(String orcid, long lastModified) {
-        List<WorkEntity> works = workDao.findFullWorks(orcid);        
-        return jpaJaxbWorkAdapter.toWorkSummary(works);
+    	List<MinimizedWorkEntity> works = workDao.findWorks(orcid);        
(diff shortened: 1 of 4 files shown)


orcid-core/src/main/java/org/orcid/core/manager/impl/WorkManagerImpl.java
```diff
@@ -222,8 +222,8 @@ private void setIncomingWorkPrivacy(WorkEntity workEntity, ProfileEntity profile
     @Override
     @Cacheable(value = "works-summaries", key = "#orcid.concat('-').concat(#lastModified)")
     public List<WorkSummary> getWorksSummaryList(String orcid, long lastModified) {
-        List<MinimizedWorkEntity> works = workDao.findWorks(orcid);        
-        return jpaJaxbWorkAdapter.toWorkSummaryFromMinimized(works);
```
rcpeters: Will this start forcing loading the full work?

nitw-shobhit: @rcpeters yup.. I removed MinimizedWorkEntity and put WorkEntity. MinimizedWorkEntity has no createdDate or lastModified fields

rcpeters: Add them. There are some records that would crash the system if we load the all their complete works.

nitw-shobhit: @rcpeters It has those fields. I missed the extended BaseEntity. Reverted and updated just the Mapper.

PR #2167, opened 2015-08-13
Add captcha field to profile

amontenegro:


amontenegro:
https://trello.com/c/IAPjTXZ6/2171-record-if-registration-used-recatcha

PR #2169, opened 2015-08-13
Widget localization changes

hexplus:


hexplus:
https://trello.com/c/9MnbUD9K/2051-internationalize-orcid-public-profile-widget-and-text

PR #2174, opened 2015-08-14
Fix error where we never cleaned the ext id url errors so they appear…

amontenegro:
… after fixing the field


amontenegro:
https://trello.com/c/vivSP8Rp/2173-if-you-enter-text-that-is-too-long-in-a-field-then-correct-it-you-still-get-an-error

PR #2175, opened 2015-08-17
Add an empty currency code, so, we can remove it if we remove the amount

amontenegro:


amontenegro:
https://trello.com/c/PvKyKiDL/2174-can-t-unselect-a-currency-in-funding

PR #2182, opened 2015-08-21
Enable public client to search

amontenegro:


amontenegro:
https://trello.com/c/alWM09h0/2170-public-search-with-a-token-doesn-t-work-for-public-clients

PR #2202, opened 2015-08-28
Language map only with locales in api

amontenegro:


amontenegro:
https://trello.com/c/0wpdZ5Y0/2186-invalid-language-codes-on-translated-titles

PR #2209, opened 2015-09-01
Cors and json p filters

amontenegro:


orcid-core/src/main/java/org/orcid/core/web/filters/CrossDomainWebManger.java
```diff
@@ -0,0 +1,118 @@
...
+    private String allowedFunctions;    
+    
+    private List<String> domainsRegex;  
+    
+    public boolean allowed(HttpServletRequest request) throws MalformedURLException {
+        URL url = new URL(request.getRequestURL().toString());
+        String path = url.getPath();        
+        
+        //Check origin header
+        if(!PojoUtil.isEmpty(request.getHeader("origin"))) {
```
amontenegro: Hey @rcpeters please look at this, Im trying to know the origin of the request, however, when working in the local host I dont get the "origin" header, i get the "referer" which contains the same info, the url the request comes from.

amontenegro:
https://trello.com/c/K0ixYybp/2203-fork-jsonpcallbackfilter-and-corsfilter-for-webapp

orcid-core/src/main/java/org/orcid/core/web/filters/CrossDomainWebManger.java
rcpeters: I would more likely make it a && check. If origin is null and referrer is localhost.

amontenegro: referer

Hey @rcpeters Im looking at the headers in each server (qa, sandbox and prod) and it looks like none of them include the origin header (See the image); all contains the host and the referrer headers, but not the origin.

Is that something we need to configure in our servers?

rcpeters: You can give your local machine a domain name using the hosts file and https://github.com/ORCID/shibboleth_nginx_vagrant

amontenegro:
@rcpeters fixed, you can now merge

orcid-core/src/main/java/org/orcid/core/web/filters/CrossDomainWebManger.java
```diff
@@ -0,0 +1,122 @@
...
+    private boolean validatePath(String path) {
+        for(String allowedPath : allowedPaths.split(",")) {
+            if(path.startsWith("/" + allowedPath)) {
+                return true;
+            }
+        }
+        return false;
+    }
+    
+    private boolean validateFunction(String path) {
```
rcpeters: Any path with that function?

amontenegro: Well, so far we have the public path, which will work for 

public_widgets
/public/group/{groupId}

Do you want me to restrict it more? Like allowing only public_widgets and /public/

rcpeters: For paths I rather have one clear regex `(^/public/|^/public_widgets/|^/userStatus.json)`. It should use a compiled regex.

amontenegro: Ok, got it, working on it

orcid-core/src/main/java/org/orcid/core/web/filters/CrossDomainWebManger.java
```diff
@@ -0,0 +1,105 @@
...
+import javax.servlet.http.HttpServletRequest;
+
+import org.orcid.pojo.ajaxForm.PojoUtil;
+import org.springframework.beans.factory.annotation.Value;
+
+public class CrossDomainWebManger {
+
+    private static final String LOCALHOST = "localhost";
+    
+    Pattern p = Pattern.compile(".*/public/.*|.*/public_widgets/.*|.*/userStatus\\.json");
```
amontenegro: Ok, to validate the path we now use a regex

rcpeters: `.*` is different then `^`?

rcpeters: "(^/public/|^/public_widgets/|^/userStatus\\.json).*"

amontenegro: It will not work for localhost calls like: /orcid-web/userStatus.json or /orcid-web/public_widgets/XXX, thats why I changed the regex

rcpeters: Add a static method to OrcidUrlManager that returns the path minus the context path. You'll also what to test with NGINX.
`request.getRequestURI().substring(request.getContextPath().length())`

amontenegro: Fixed, however, didnt test on NGINX because I couldn't start the vagrant for (https://github.com/ORCID/shibboleth_nginx_vagrant) on my machine.

PR #2214, opened 2015-09-04
Custom validation for notifications

wjrsimpson:


wjrsimpson:
https://trello.com/c/IJpeqrD1/2209-orcid-records-says-there-are-messages-but-displays-none

PR #2221, opened 2015-09-10
Added new column to profile table and code to get the ipaddress

nitw-shobhit:
https://trello.com/c/YHWGacA4/2148-log-ip-address-of-user-on-create-or-login-to-the-db

### DIFF
--- a/orcid-web/src/main/java/org/orcid/frontend/web/controllers/RegistrationController.java
+++ b/orcid-web/src/main/java/org/orcid/frontend/web/controllers/RegistrationController.java
@@ -90,6 +90,7 @@ import org.orcid.pojo.ajaxForm.PojoUtil;
 import org.orcid.pojo.ajaxForm.Registration;
 import org.orcid.pojo.ajaxForm.Text;
 import org.orcid.utils.DateUtils;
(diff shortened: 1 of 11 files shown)


orcid-web/src/main/java/org/orcid/frontend/web/controllers/RegistrationController.java
```diff
@@ -311,8 +311,18 @@ public static OrcidProfile toProfile(Registration reg) {
 
         profile.setPassword(reg.getPassword().getValue());
 
+        profile.setUserLastIp(getIpAddress(request));
         return profile;
     }
+
```
rcpeters: Looks good, but make a OrcidRequestUtil.java under /orcid-utils/src/main/java/org/orcid/utils

rcpeters:
@Laura-ORCID Do we want to include privacy policy changes?

orcid-web/src/main/java/org/orcid/frontend/web/controllers/RegistrationController.java
nitw-shobhit: @rcpeters done

PR #2230, opened 2015-09-16
Internal sso

amontenegro:


orcid-core/src/main/java/org/orcid/core/manager/impl/InternalSSOManagerImpl.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Override
+    public void deleteToken(String orcid, HttpServletRequest request, HttpServletResponse response) {
+        // Delete the DB row
+        internalSSODao.delete(orcid);
+        // Delete the cookie
+        if (request.getCookies() != null) {
+            for (Cookie cookie : request.getCookies()) {
+                if (cookie.getName().equals(COOKIE_NAME)) {
+                    cookie.setMaxAge(0);
```
rcpeters: Setting max age isn't reliable. You should also set the value to nothing.

orcid-core/src/main/java/org/orcid/core/manager/InternalSSOManager.java
```diff
@@ -0,0 +1,60 @@
...
+     * */
+    void deleteToken(String orcid, HttpServletRequest request, HttpServletResponse response);
+    
+    /**
+     * Returns true if the given token is still valid
+     * 
+     * @param orcid
+     * @param encryptedToken
+     * */
+    boolean verifyToken(String orcid, String encryptedToken);
```
rcpeters: encryptedToken? Isn't the token just a RandomString?

amontenegro: haha yeah, it should be just token! Fixed

orcid-core/src/main/java/org/orcid/core/manager/impl/InternalSSOManagerImpl.java
amontenegro: Fixed

rcpeters:
merge monster

amontenegro:
@rcpeters fixed!

PR #2232, opened 2015-09-17
Statistics cache manager

nitw-shobhit:
https://trello.com/c/JaaZU45Q/2211-write-statistics-caching-to-use-cache-manger-instead-of-cacheable

### DIFF
--- a/orcid-core/src/main/java/org/orcid/core/manager/impl/StatisticsCacheManagerImpl.java
+++ b/orcid-core/src/main/java/org/orcid/core/manager/impl/StatisticsCacheManagerImpl.java
@@ -1,0 +1,141 @@
+/**
+ * =============================================================================
+ *
+ * ORCID (R) Open Source
+ * http://orcid.org
+ *
+ * Copyright (c) 2012-2014 ORCID, Inc.
+ * Licensed under an MIT-Style License (MIT)
+ * http://orcid.org/open-source-license
+ *
+ * This copyright and license information (including a link to the full license)
+ * shall be included in its entirety in all copies or substantial portion of
+ * the software.
+ *
+ * =============================================================================
+ */
+package org.orcid.core.manager.impl;
+
+import java.text.NumberFormat;
+import java.util.Locale;
+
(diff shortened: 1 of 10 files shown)


orcid-core/src/main/java/org/orcid/core/manager/impl/StatisticsCacheManagerImpl.java
```diff
@@ -0,0 +1,141 @@
...
+    LockerObjectsManager lockers = new LockerObjectsManager();
+
+	@Override
+	public StatisticsSummary retrieve() {
+        StatisticsSummary statisticsSummary = toStatisticsSummary(statisticsCache.get("statisticsValues"));
+        StatisticKeyEntity statisticKeyEntity = toStatisticsKey(statisticsCache.get("statisticsLatestKey"));
+        StatisticKeyEntity statisticKeyEntityLatest = statisticsManager.getLatestKey();
+        if(needsFresh(statisticsSummary, statisticKeyEntity, statisticKeyEntityLatest)) {
+        	try {
+                synchronized (lockers.obtainLock("Statistics")) {
```
rcpeters: You're locking on a sting? Where did that idea come from?

nitw-shobhit: @rcpeters unlike the profile table, which locks a specific row(using orcid value), a table lock is required here. So I used the string "Statistics" as the key.

orcid-model/src/main/java/org/orcid/jaxb/model/statistics/StatisticsTimeline.java
```diff
@@ -32,8 +33,13 @@
  */
 @XmlRootElement(name = "statistics-timeline")
 @XmlAccessorType(XmlAccessType.FIELD)
-public class StatisticsTimeline {
-    @XmlElement(name = "statistic-name")
+public class StatisticsTimeline implements Serializable {
```
rcpeters: Good cache

PR #2254, opened 2015-09-23
Refreshing the cache buster always on dev

amontenegro:
@rcpeters real change is on BaseController.java, the change on OauthAuthorizationPageTest is removing a black box test that dont work anymore because of the recaptcha on oauth register page


rcpeters:
:+1:

PR #2263, opened 2015-09-24
fork the filters

rcpeters:


orcid-web/src/main/resources/orcid-frontend-security.xml
```diff
@@ -296,7 +296,7 @@
 
 	<bean id="corsFilter" class="org.orcid.core.web.filters.CorsFilter" />
```
rcpeters: @amontenegro this one is missing.

PR #2277, opened 2015-09-29
Sso cookie

amontenegro:


amontenegro:
Hey @rcpeters,

In this pull request I included the following changes: 

1) Remove the cookie check on userStatus, so, it doesnt matter if the cookie is not there, it will not logout the user
2) The cookie will be ignored on localhost
3) Cookie is set on login, register and switch user 

Thanks

PR #2297, opened 2015-10-05
Researcher urls v2

amontenegro:


amontenegro:
Hey @wjrsimpson @rcpeters!

This is the API for researcher urls, I think is good enough to start testing, let me know if you see any issues or question

Thanks

PR #2301, opened 2015-10-06
Throw 409 on notifications POST when user have profile.send_member_up…

amontenegro:
…date_requests = t.


amontenegro:
https://trello.com/c/K2utCYgY/2294-messaging-return-409-from-api-if-user-has-opted-out-of-permission-notifications

PR #2303, opened 2015-10-06
Service announcement 2015

rcpeters:


rcpeters:
Profile event to send service message. However the message itself still needs to be added (copied old message for now).

amontenegro:
Looks good!

PR #2307, opened 2015-10-06
Set the string value of the int variable

amontenegro:


amontenegro:
https://trello.com/c/psCfyiLV/2298-notifications-link-that-doesn-t-resolve-extra-brackets

rcpeters:
ha! good find.

PR #2310, opened 2015-10-07
Dont send notifications when profile.send_member_update_requests = f

amontenegro:


orcid-core/src/main/java/org/orcid/core/manager/impl/NotificationManagerImpl.java
```diff
@@ -836,7 +836,7 @@ public Notification createNotification(String orcid, Notification notification)
         if (profile == null) {
             throw OrcidNotFoundException.newInstance(orcid);
         }
-        if(profile.getSendMemberUpdateRequests() != null && profile.getSendMemberUpdateRequests()) {
+        if(profile.getSendMemberUpdateRequests() != null && !profile.getSendMemberUpdateRequests()) {
```
rcpeters: `if (profile.getSendMemberUpdateRequests() == false)`?

amontenegro: But, it the same right? It is just more readable

PR #2321, opened 2015-10-07
A new CLI that will generate a claim link

amontenegro:


amontenegro:
New CLI to generate claim link, this will help us in cases where the user is not getting it

PR #2353, opened 2015-10-13
New ga code

amontenegro:


orcid-web/src/main/webapp/static/javascript/script.js
```diff
@@ -129,24 +129,29 @@ var OrcidCookie = new function() {
...
         return clientGroupName + ' - ' + clientName
     };
     this.gaPush = function(trackArray) {
         if (gaEnabled) {
-            _gaq.push(trackArray);
-            console.log("_gap.push for " + trackArray);
+        	if(typeof trackArray === 'function') {
+        		trackArray();
+        	} else {
+        		if(trackArray[5] == undefined) {
```
rcpeters: what's the documentation for this crazy bit?

orcid-web/src/main/webapp/static/javascript/script.js
```diff
@@ -129,24 +129,29 @@ var OrcidCookie = new function() {
...
-
 var OrcidGA = function() {
     // test and make sure _gaq is working. disconnect.me chrome plugin has
     // caused silent _gaq failures. This check allows us to detect that
     // situation
-    var gaEnabled = false;
-    _gaq.push(function() {
-        gaEnabled = true;
-    });
+    var gaEnabled = true;
```
rcpeters: Um this was a test to make sure GA was loaded. For instance ad blockers.

rcpeters: http://davidwalsh.name/detecting-google-analytics

orcid-web/src/main/webapp/static/javascript/script.js
amontenegro: Look at: 

https://developers.google.com/analytics/devguides/collection/upgrade/reference/gajs-analyticsjs#events

Instead of changing all calls, I keep using the trackArray, but, element at [5] is optional, so, if it is undefined, I just dont send it.

orcid-web/src/main/webapp/static/javascript/script.js
amontenegro: Let me see if the new version have something similar

amontenegro: Ok, am testing http://davidwalsh.name/detecting-google-analytics, I will let you know

orcid-web/src/main/webapp/static/javascript/script.js
```diff
@@ -129,24 +129,31 @@ var OrcidCookie = new function() {
...
     // situation
     var gaEnabled = false;
-    _gaq.push(function() {
-        gaEnabled = true;
-    });
-
+    if(window.ga && ga.create) {
+    	gaEnabled = true;
+    }
+
```
amontenegro: Ok @rcpeters here is the test to check if ga was loaded.

orcid-web/src/main/webapp/static/javascript/script.js
```diff
@@ -129,24 +129,31 @@ var OrcidCookie = new function() {
...
+    
     this.buildClientString = function(clientGroupName, clientName) {
         return clientGroupName + ' - ' + clientName
     };
     this.gaPush = function(trackArray) {
         if (gaEnabled) {
-            _gaq.push(trackArray);
-            console.log("_gap.push for " + trackArray);
+        	if(typeof trackArray === 'function') {
+        		ga(trackArray);
```
amontenegro: I changed this line, so, it uses the ga queue so it executes the callback in the order it was called by the app.

orcid-web/src/main/webapp/static/javascript/script.js
```diff
@@ -129,24 +129,31 @@ var OrcidCookie = new function() {
     };
 };
 
-var _gaq = _gaq || [];
-
 var OrcidGA = function() {
     // test and make sure _gaq is working. disconnect.me chrome plugin has
     // caused silent _gaq failures. This check allows us to detect that
     // situation
     var gaEnabled = false;
```
rcpeters: I'm not 100% sure how I was making sure the GA script was loaded before this code was run....I think we should make a method retest the and reset the gaEnabled variable on load of the GA script.

amontenegro: Hey @rcpeters,

But, according to the link we saw yesterday: 

http://davidwalsh.name/detecting-google-analytics

The way to test that ga was loaded is like this: 

https://github.com/ORCID/ORCID-Source/pull/2353/files#diff-97eba34be435b164fc1d5c488a1edd13R137

So, you want me to make that a function (like onLoad) and call it when the OrcidGA is loaded?

Thanks

PR #2369, opened 2015-10-15
On the account settings page, display all trusted organizations activ…

amontenegro:
…ity, no matter if it have the same scopes, display them all


rcpeters:
It's always good when we can delete code!

PR #2390, opened 2015-10-22
Bring back signout while skipping csrf check

rcpeters:


rcpeters:
@wjrsimpson @amontenegro I'm not 100% on this one. Let's talk.

PR #2391, opened 2015-10-22
To get the form we should use GET

amontenegro:
@rcpeters this will fix the public client page


orcid-web/src/main/java/org/orcid/frontend/web/controllers/DeveloperToolsController.java
```diff
@@ -220,7 +220,7 @@ boolean resetClientSecret(@RequestBody String clientId) {
         return orcidSSOManager.resetClientSecret(clientId);
     }
 
-    @RequestMapping(value = "/get-sso-credentials.json", method = RequestMethod.POST)
```
rcpeters: @wjrsimpson is the the correct way?

wjrsimpson: I think so. Looks good to me.

PR #2397, opened 2015-10-23
Removing unnecessary http calls when the servers are on https

amontenegro:
https://trello.com/c/E8FrOIk8/2373-https-orcid-org-oauth-authorize-redirects-to-http-orcid-org-signin-then-https-orcid-org-signin


orcid-web/src/main/java/org/orcid/frontend/web/controllers/LoginController.java
```diff
@@ -54,8 +54,9 @@ public ModelAndView loginGetHandler(HttpServletRequest request, HttpServletRespo
     @RequestMapping(value = { "/signout"}, method = RequestMethod.GET)
     public ModelAndView signout(HttpServletRequest request, HttpServletResponse response) {
         // in case have come via a link that requires them to be signed out
-        logoutCurrentUser(request, response);
-        ModelAndView mav = new ModelAndView("redirect:/signin");
+        logoutCurrentUser(request, response);    
+        String redirectString = "redirect:" + orcidUrlManager.getBaseUrl()  + "/signin";
```
rcpeters: excellent catch!

PR #2400, opened 2015-10-26
update Czech translation

helix84:
updated Czech translation
fixed some typos in English original


rcpeters:
@helix84 we have a set of translations out. Holding until we get that back.

PR #2404, opened 2015-10-26
Card2278

amontenegro:


amontenegro:
https://trello.com/c/S7V48sKl/2278-store-orcididhash-on-session-or-find-a-way-that-the-method-baseworkspacecontroller-getorcidhash-is-called-just-once-on-login

orcid-core/src/main/java/org/orcid/core/manager/impl/ProfileEntityManagerImpl.java
```diff
@@ -544,6 +550,15 @@ public Visibility getResearcherUrlDefaultVisibility(String orcid) {
     	return jpa2JaxbAdapter.getApplications(tokenDetails);
     }
     
+    @Override
+    @Cacheable(value = "orcid-hash")
```
rcpeters: Why not use a session instead of a cache? I really don't like the cacheable annotation.

amontenegro: Fixed @rcpeters, now we will use the request instead of cache.

PR #2423, opened 2015-10-30
Fix where we log oauth2 redirects

amontenegro:


amontenegro:
Hey @rcpeters, 

There are no errors on the logic, the empty redirects you saw in the logs (OauthConfirmAccessController form.getRedirectUri being sent to client browser: https://XXXX/orcid/callback) happens because we where login a redirect no matter if the form was going to be redirected or not.

So, in example if a user enters an incorrect password, we send the user back to the oauth login page, however, we still write the log entry, which at that point is only the redirect_uri with no params

I "fixed" the code so it log only when the user will be redirected, we might want to do a patch release with this to get ride of a lot of incorrect lines in the logs.

After the refactoring this piece of code should look way better.

Thanks

rcpeters:
Shoot! No real errors. Ok merging and pushing out. Last night I was thinking we shouldn't overwrite redirectUri but instead set a new member finalUserRedirectUri when we are actually going to redirect.

PR #2433, opened 2015-11-05
Fix condition where stats doesn't show

rcpeters:


rcpeters:
@amontenegro @wjrsimpson can one of you please review? I would like to patch before tomorrow.

wjrsimpson:
@rcpeters It is certainly a great improvement.

I wonder if there is still very small possibility of some of the stats not being present in the DB for the latest key.

There is no transaction around the following loop in StatisticsGeneratorCronJobImpl.

    // Store statistics on database
    for (String key : statistics.keySet()) {
        statisticsManager.saveStatistic(statisticKey, key, statistics.get(key));
    }

So, just one stat might be present and you might get the latest key due to that. But the other stats would not be there yet.

This might be a rare case where an transaction would be worthwhile. But only around the saving of the stats, and not the calculation of the stats, because that would be a too long transaction.

Hum, in reality though there is only a second or so each week when that could happen, so probably not worrying about.

PR #2434, opened 2015-11-05
Fix shib csrf with latest

wjrsimpson:
I've updated the Shibboleth link page (where you sign in with your ORCID credentials) to use the same ajax as the main sign in page. I've also added code to regenerate the CSRF token after programmatically sigining the user out.

@rcpeters Do you reckon that looks OK?

https://trello.com/c/umCGdRDk/2411-federated-login-linking-page-broken-by-csrf


orcid-web/src/main/java/org/orcid/frontend/web/controllers/BaseController.java
```diff
@@ -353,7 +358,10 @@ protected void logoutCurrentUser(HttpServletRequest request, HttpServletResponse
 		}
 		if (authentication != null && authentication.isAuthenticated()) {
 			new SecurityContextLogoutHandler().logout(request, response, authentication);
```
rcpeters: @wjrsimpson hum I thought this line caused the a new token to be made.? Please verify that wasn't the case.

wjrsimpson: It gets set to null when logout is called. Then it gets generated by CsrfFilter.java on the next request. I think that is why it works when there is redirect.
In the case where there is no redirect, it is set to null, but the old token (associated with the previous session) is still set as a request attribute. It is then rendered in the page header, and used in subsequent ajax calls, but is invalid.

PR #2449, opened 2015-11-13
For GA calls, when a public client is doing the oauth dance, the memb…

amontenegro:
…er name should be 'PubApp'


amontenegro:
Hey @rcpeters,

This is the change we need for track GA actions for the public client.

Thanks

PR #2457, opened 2015-11-17
Rename credit name visibility field on db

amontenegro:


amontenegro:
https://trello.com/c/zdyeakVL/2440-rename-the-credit-name-visibility-field-to-the-more-general-name-names-visibility

PR #2458, opened 2015-11-18
add read limited scopes that works for both person and activities

rcpeters:


amontenegro:
Hey @rcpeters, code looks fine, however, I dont see any script or query that updates the existing tokens, look at this in PROD DB:

select * from oauth2_token_detail where (scope_type like '%/activities/read-limited%' or scope_type like '%/person/read-limited%') and token_expiration > now();

If it is ok to ignore the existing tokens let me know, but, we need to be aware that if we remove /acitivities/read-limited and /person/read-limited from ScopePathType.java those tokens will not work anymore.

rcpeters:
@amontenegro Not removing those other scopes, they will continue to exist for ones issued. @caoyler will be moving client who used those scopes over to using the new scope once it's available.

PR #2462, opened 2015-11-20
fix issue where grouping works and fundings in v2.0_rc1 was listing s…

rcpeters:
…ame groupings keys twice


rcpeters:
@amontenegro please check this out

PR #2468, opened 2015-12-01
Fix scopes behavior

amontenegro:


orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidSecurityManagerImpl.java
```diff
@@ -73,6 +79,14 @@ public void checkVisibility(Filterable filterable) {
             readLimitedScopes = getReadLimitedScopesThatTheClientHas(authorizationRequest, filterable);
         }
 
+        //TODO!!! Is this the right way of doing this?
```
amontenegro: Hey @rcpeters can we take a look at this please?

PR #2469, opened 2015-12-01
Fix scopes behavior

amontenegro:


amontenegro:
@rcpeters @wjrsimpson This is ready to go, this will allow the read-limited or update scopes to read the data where the client is the source.

This will also fix the blackbox tests and the unit tests

PR #2471, opened 2015-12-02
Refresh page if the user status doesn't return a 200 response

rcpeters:


wjrsimpson:
@rcpeters This causes the page to refresh if you sign into ORCID in another tab. The original tab refreshes. But you are still signed in and on the same page in the original tab, so I suspect that is OK. Just an observation, really.

rcpeters:
Yes I'm aware. We'll see if we get any complaints. The other option is to
goto cookie based CSRF.

PR #2472, opened 2015-12-02
RDF redirects should be 303 See Other

stain:
as a person is a non-information resource

See http://www.w3.org/2001/tag/doc/httpRange-14/2007-05-31/HttpRange-14#iddiv2130462376

This was raised as an issue in http://www.w3.org/2015/03/orcid-semantics


amontenegro:

TomDemeranville:
I'm not an admin, but this looks correct to me.

PR #2473, opened 2015-12-02
RDF support: Update jena

stain:
Update to use [Apache Jena](http://jena.apache.org/) 3.0.0

As agreed with @TomDemeranville in [ORCID-RDF telcon](https://gist.github.com/stain/abaf89cf40df2823dde1)

Note that Jena 3.0.0 changed package name from `com.hp.jpl.jena` to `org.apache.jena`.

BTW, why is Jena depended on by both `orcid-api-common` (which uses it from `RDFMessageBodyWriter`) and `orcid-core` (which don't use it)?

This also removes the old loadOntology() methods in favour of using Jena's [schemagen Maven plugin](http://jena.apache.org/documentation/tools/schemagen-maven.html) which adds vocabulary constant classes to `target/generated-sources/jena`. This should thus be easier to extend to use other vocubularies. As an example, see the `FOAF` inner class.


stain:
Apologies, don't merge yet.. I need to figure out the build-helper settings.

stain:
Fixed - now builds well with Maven and Eclipse.

TomDemeranville:
Hi Stian.  Do you also have the changes that alter the 307 responses to 303?

Tom.

stain:
Yes, @TomDemeranville that is in #2472 which should be compatible with this change.

amontenegro:

PR #2474, opened 2015-12-02
RDF: Add Turtle and JSON-LD support

stain:
Adds supports for RDF media types `application/ld+json` ([JSON-LD](http://www.w3.org/TR/json-ld/)) and `application/n-triples` ([N-triples](http://www.w3.org/TR/n-triples/)).

Based on pull request #2473 and from discussion with @TomDemeranville at [ORCID-RDF telcon](https://gist.github.com/stain/abaf89cf40df2823dde1).

Note that `OrcidApiServiceDelegator` and subclasses remains to be updated to include `JSON_LD` and `N_TRIPLES` in the correct `@Produces` for the redirects and `experimental_rdf` paths.
- the `RDFMessageBody` serializes depending on the negotiated `MediaType` it is given. (so does it really need multiple methods there for Turtle, RDF/XML etc, or would a single rdf method do?)

Serializing JSON-LD is slightly more heavyweight to generate, so that's why I set it at q=2.


amontenegro:

PR #2500, opened 2015-12-09
Check if GA is enabled before pushing an event

amontenegro:


amontenegro:
Pull request 2500!!!!

amontenegro:
https://trello.com/c/akOaVxSQ/2468-check-if-google-analitycs-is-enable-just-before-sending-an-event

rcpeters:
@wjrsimpson @amontenegro Please see  https://developers.google.com/analytics/devguides/collection/analyticsjs/?hl=en
GA should always be available. The root issue looks like we also testing for ga.create which may not be true. We should always be pushing to qa regardless of "create". However it also looks like do to how UA has changed we need to call trackArray(); if ga.create isn't there.

wjrsimpson:
@rcpeters Makes sense, cheers.

PR #2502, opened 2015-12-11
Allow custom emails for basic institutions

amontenegro:


amontenegro:
https://trello.com/c/pL0wvQiF/2465-allow-custom-claim-emails-for-basic-institutions-only-if-it-is-easy

PR #2506, opened 2015-12-11
Moving GA load to the top of the page and removing the check for ga.c…

amontenegro:
Hey @rcpeters,

Could you please check if this is ok according to your comment on

https://github.com/ORCID/ORCID-Source/pull/2500#issuecomment-163716830

With this we will load the GA script right after the win-7 hacks, and, we are not checking ga.create anymore since it looks we dont need to

Thanks

### DIFF
--- a/orcid-web/src/main/webapp/static/javascript/script.js
+++ b/orcid-web/src/main/webapp/static/javascript/script.js
@@ -134,7 +134,7 @@ var OrcidGA = function() {
     // caused silent _gaq failures. This check allows us to detect that
     // situation
     var gaEnabled = false;
-    if(window.ga && ga.create) {
+    if(window.ga) {
     	gaEnabled = true;
     }
     
@@ -144,7 +144,7 @@ var OrcidGA = function() {
     this.gaPush = function(trackArray) {
         //Recheck ga is enabled
         if(!gaEnabled) {
-            if(window.ga && ga.create) {
+            if(window.ga) {
                 gaEnabled = true;
             }
         }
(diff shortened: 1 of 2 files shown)


rcpeters:
Reading the code it looks like you are missing the case where ga.create doesn't exist. That means functions pushed to qa may or may not run. Did you test while blocking GA?

amontenegro:
Yes sir! @rcpeters I tested while blocking GA and it worked fine.

rcpeters:
@amontenegro Tested locally, no it doesn't. You need to test trying to load GA but the loading fails.

orcid-web/src/main/webapp/static/javascript/script.js
```diff
@@ -131,42 +131,35 @@ var OrcidCookie = new function() {
...
-            }
-        }
-        
-        if (gaEnabled) {
-        	if(typeof trackArray === 'function') {
-        		ga(trackArray);
-        	} else {
-        		if(trackArray[5] == undefined) {
-                	ga(trackArray[0], trackArray[1], trackArray[2], trackArray[3], trackArray[4]);
+        if(window.ga) {
```
amontenegro: Hey @rcpeters,

Look at these changes: 

1) First we check if ga was loaded (if(window.ga)), usually on dev env we dont load the ga scripts, so window.ga is null; so, we need to check if GA was loaded before proceed
2) Fill the ga array according to the trackArray (line 140, 143 and 145)
3) On line 150 we wait 200 ms, then, we check for ga.create, if it is not defined it means that GA was not loaded, so, we call th trackArray to continue with the normal flow of the site
4) The else statement for 1, if window.ga was undefined, call the trackArray

Please let me know if you see any issues.

Thanks

rcpeters: looks good :+1:

PR #2508, opened 2015-12-11
Redirect if already loged in

amontenegro:
https://trello.com/c/ptJQ9jbA/2471-stored-request-not-being-cleared-after-user-completes-oauth-authorization


amontenegro:
@rcpeters look at this one, now we clear the saved request after the oauth authorization

PR #2510, opened 2015-12-15
Show source id if name is not public

amontenegro:


amontenegro:
https://trello.com/c/eYooyzee/2477-cant-add-affiliations-on-qa-and-sandbox

PR #2513, opened 2015-12-17
Set the names_visibility to public where credit_name is empty or null

amontenegro:


amontenegro:
https://trello.com/c/LFh7XroR/1927-update-db-set-names-visibility-to-public-if-credit-name-is-null-or-empty

PR #2518, opened 2015-12-21
Fix an NPE that heppens if the Oauth authorize page is called without…

amontenegro:
… any param


amontenegro:
https://trello.com/c/7W5feakn/2480-fix-npe-error-filling-oauth-form-in-prod

PR #2535, opened 2016-01-08
Cookie improvements

wjrsimpson:
https://trello.com/c/UIwkPOcd/2522-make-session-cookie-https-only


rcpeters:
Starting out with formatting code makes patches very tedious.....

wjrsimpson:
Ouch, yes. I don't think I was expecting that one to be a patch when I started out.

PR #2539, opened 2016-01-13
Last modified dates

nitw-shobhit:
https://trello.com/c/tL2hGX6x/2515-add-last-modified-dates-to-rc2-xsds-and-model-classes-wherever-it-can-be-propagated


orcid-model/src/main/resources/record_2.0_rc2/activities-2.0_rc2.xsd
```diff
@@ -65,7 +65,7 @@
 					</xs:annotation>
 					<xs:complexType>
 						<xs:sequence>
-							<xs:element ref="common:last-modified-date" minOccurs="1" maxOccurs="1" />
+							<xs:element ref="common:last-modified-date" minOccurs="0" maxOccurs="1" />
```
amontenegro: Hey @nitw-shobhit, why do you set the last-modified-date as required? what happens if the list is empty and the lastModifiedDate object is empty in the java object?

nitw-shobhit: @amontenegro i removed minOccurs=1 and added minOccurs=0, so it is not mandatory..

rcpeters: Hum good questions. What happens if someone has a list and then deletes it.....

rcpeters: For now if the list is empty the last modified is the record creation date.

PR #2541, opened 2016-01-14
Set names visibility to public when create record from API

amontenegro:


amontenegro:
https://trello.com/c/cvpZfTTb/2504-default-names-to-public-when-creating-records

PR #2554, opened 2016-01-22
Fix notifications keys from the NotificationType enum

amontenegro:


amontenegro:
https://trello.com/c/L0IeaX2q/2548-notifications-error-due-to-a-missing-key

PR #2575, opened 2016-01-26
Fixing error where the accept filter was setting the incorrect accept…

amontenegro:
… header to the orcid-internal-api requests

### DIFF
--- a/orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidSecurityManagerImpl.java
+++ b/orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidSecurityManagerImpl.java
@@ -292,12 +292,14 @@ public class OrcidSecurityManagerImpl implements OrcidSecurityManager {
 
     @Override
     public String getClientIdFromAPIRequest() {
-        OAuth2Authentication oAuth2Authentication = getOAuth2Authentication();
-        if (oAuth2Authentication == null) {
-            return null;
+        SecurityContext context = SecurityContextHolder.getContext();
+        Authentication authentication = context.getAuthentication();
+        if (OAuth2Authentication.class.isAssignableFrom(authentication.getClass())) {
+            OAuth2Authentication oAuth2Authentication = (OAuth2Authentication) authentication;
+            OAuth2Request request = oAuth2Authentication.getOAuth2Request();
+            return request.getClientId();
         }
-        OAuth2Request request = oAuth2Authentication.getOAuth2Request();
-        return request.getClientId();
+        return null;
     }
 
     private boolean hasScope(ScopePathType scope) {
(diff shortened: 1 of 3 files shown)


amontenegro:
https://trello.com/c/8WTFjVpY/2554-internal-api-broken

orcid-core/src/main/java/org/orcid/core/web/filters/AcceptFilter.java
```diff
@@ -53,10 +60,13 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
             if (isValidAcceptType(contentType))
                 requestWrapper = new AcceptHeaderRequestWrapper(request, contentType);
             else
-                if (OrcidUrlManager.getPathWithoutContextPath(request).startsWith("/oauth/"))
+                if (OrcidUrlManager.getPathWithoutContextPath(request).startsWith("/oauth/")) {
```
rcpeters: Hey, 
* In what case is path null? 2) 
* why doesn't getPathWithoutContextPath without context work (that a leading questions)?

amontenegro: Hey @rcpeters,

For 1 path should never ever be null, but is a pretty small check we can add to prevent a null pointer exception that will not be logged (We could also add the logging line in case path is null)

For 2, the calls to the internal API are done like this: 

http://qa-5.orcid.org:8080/orcid-internal-api/member-info

So, OrcidUrlManager.getPathWithoutContextPath(request) will return just /member-info, what I want to check is that this is a request to the internal api, so, I need the request URI which is the full "/orcid-internal-api/member-info" and then verify that it starts with the internal api path.

Please let me know if am wrong. 

Thanks

orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidSecurityManagerImpl.java
```diff
@@ -292,12 +292,14 @@ private OAuth2Authentication getOAuth2Authentication() {
 
     @Override
     public String getClientIdFromAPIRequest() {
-        OAuth2Authentication oAuth2Authentication = getOAuth2Authentication();
-        if (oAuth2Authentication == null) {
-            return null;
+        SecurityContext context = SecurityContextHolder.getContext();
```
amontenegro: This method is used to get the client id from the authentication, we are using it in the OrcidExceptionMapper to log the client ID that is doing the request.

But, for the internal API this is not a OAuth2Authentication, so, when we call the method getOAuth2Authentication() it throw an "AccessControlException", and, that prevent us to log the error that is happening.

amontenegro:
@rcpeters dont merge this, I will just send a new pull request with just the change we need.

PR #2576, opened 2016-01-27
Rolling back method getClientIdFromAPIRequest so it dont fail for non…

amontenegro:
… oauth2 authentications


amontenegro:
@rcpeters this is the rollback I need in the OrcidSecurityManager, the method will be rolledback as it was in the previous version:

https://github.com/ORCID/ORCID-Source/blob/cf6d2e54194c0e756387faadc5d9442820ec4016/orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidSecurityManagerImpl.java#L266

Thanks

PR #2590, opened 2016-02-03
Better error message when no api version is specified

amontenegro:


amontenegro:
https://trello.com/c/bby9Wig6/2569-posting-or-putting-without-a-version-number-returns-a-500-error

PR #2592, opened 2016-02-03
change user status check to only run where there is a non-blank csrf …

rcpeters:
…token


wjrsimpson:
@rcpeters Thanks, looks good!

PR #2601, opened 2016-02-08
Put code redemption examples on same domain user auth

rcpeters:


amontenegro:
https://trello.com/c/h5crpfQq/2581-change-examples-on-https-orcid-org-developer-tools-to-redeem-code-on-same-domain-as-issued-auth-issued-from-user

PR #2611, opened 2016-02-12
xsd for bulk work post

rcpeters:
XSD that could represent taking a list of works for posting and returning a list of works or errors in the same order as works posted.


amontenegro:
Hey @rcpeters am looking at this now, I will include my changes here too and then we can send to QA.

Thanks

amontenegro:
Hey @rcpeters, 

Am ok with the XSD, so, I just pushed the java pojo and a marshaling test; please have a quick look, if you are ok, we can merge this and start working on the bulk edit in a new branch

Thanks

PR #2623, opened 2016-02-16
Using versioned delegator for public API. formatting for OrcidExceptionMapper

nitw-shobhit:
https://trello.com/c/KTef0ZYe/2598-public-api-2-0-not-using-service-versioned-delegator


rcpeters:
@nitw-shobhit merge monster

PR #2648, opened 2016-03-02
All enum values on V2_rcX should be lowercase

amontenegro:


amontenegro:
https://trello.com/c/J30tM4x4/2609-create-a-single-whitebox-test-that-sets-up-blackbox-test

amontenegro:
Hey @wjrsimpson could you please look at this?

So, starting on V2_rc1 all enum values should be lowercase on XMLs right?

Thanks

PR #2649, opened 2016-03-02
forced V2.0 rc2 person external identifiers to default to public visi…

TomDemeranville:
…bility


orcid-model/src/main/java/org/orcid/jaxb/model/record_rc2/PersonExternalIdentifier.java
```diff
@@ -119,6 +119,8 @@ public void setPutCode(Long putCode) {
         this.putCode = putCode;
     }
     public Visibility getVisibility() {
+        if (visibility == null)
```
rcpeters: What does it show in the db, it should always be explicit? @wjrsimpson any thoughts?

TomDemeranville: Without this, it defaults to PRIVATE as it's the first value in the Enum.

PR #2655, opened 2016-03-07
fixed null pointer when not adding subject review identifier

TomDemeranville:


TomDemeranville:
Tests pass locally!  Failing on jenkins with context not found.

PR #2658, opened 2016-03-07
Adding a test to check that the peer reviews are visible/hidden depen…

amontenegro:
…ding on client visibility


amontenegro:
The error here will be fixed with this pull request: 

https://github.com/ORCID/ORCID-Source/pull/2657

PR #2661, opened 2016-03-08
replaced odd logic duplicate checking with equals method in ExternalID

TomDemeranville:


orcid-core/src/main/java/org/orcid/core/manager/validator/ActivityValidator.java
```diff
@@ -155,22 +155,6 @@ public static void checkExternalIdentifiersForDuplicates(ExternalIDs newExtIds,
         }
     }
 
-    private static boolean isDupRelationship(ExternalID newId, ExternalID existingId) {
-        return existingId.getRelationship() != null && existingId.getRelationship().equals(Relationship.SELF) && newId.getRelationship() != null
-                && newId.getRelationship().equals(Relationship.SELF);
```
rcpeters: @TomDemeranville this will break. Identifiers that are part of can exist across multiple activities. @wjrsimpson @amontenegro can you help Tom.

TomDemeranville:
Sat in the hospital and bored, looked at commit and realised I left some
system.out.println in there

PR #2665, opened 2016-03-08
Adding the X language file for emails and allowing setting the X lang…

amontenegro:
…uage in the DB

### DIFF
--- a/orcid-model/src/main/java/org/orcid/jaxb/model/message/Locale.java
+++ b/orcid-model/src/main/java/org/orcid/jaxb/model/message/Locale.java
@@ -81,7 +81,10 @@ public enum Locale implements Serializable {
     IT("it"),
     
     @XmlEnumValue("ja")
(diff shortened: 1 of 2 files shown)


orcid-model/src/main/java/org/orcid/jaxb/model/message/Locale.java
```diff
@@ -81,7 +81,10 @@
     IT("it"),
     
     @XmlEnumValue("ja")
-    JA("ja");
+    JA("ja"), 
+    
+    @XmlEnumValue("xx")
```
amontenegro: Hey @wjrsimpson @rcpeters please look at this, this is the fastest way to allow a user to have the X language in his locale.

Is this ok?

Thanks

amontenegro:
https://trello.com/c/KDBPt77J/2651-registration-email-sent-only-in-english-italian

PR #2677, opened 2016-03-11
Throw error if member tries to update visibility

amontenegro:


amontenegro:
https://trello.com/c/NLvZKgCy/2665-api-2-0-throw-an-error-if-the-member-tries-to-change-the-visibility-of-an-item

PR #2684, opened 2016-03-16
remove hear about table and associated classes

rcpeters:


orcid-persistence/src/test/java/org/orcid/persistence/dao/AddressDaoTest.java
```diff
@@ -1,124 +0,0 @@
-/**
```
amontenegro: Hey @rcpeters why you deleted this class? AddressDaoTest.java ... AddressDao should not be affected by this pull request, right?

amontenegro:
Ok, this looks good now!

PR #2687, opened 2016-03-17
Fixing other unit test

amontenegro:


amontenegro:
Fixing unit test that fail

PR #2688, opened 2016-03-18
Fix item level privacy in apis

TomDemeranville:
This branch now passes all test.

Section level visibility for Keywords, ResearcherUrls, OtherNames and ExternalIdentifiers are no longer persisted or retrieved from the database. 

1.2 API now populates section level privacy attribute with the MOST restrictive in the section
It then filters out all items unavailable to client due to visibility.
It then removes sections that have no items in them.

When creating or updating items within a section using 1.2, the users default activity visibility is used and the section level on the incoming model is ignored.

TODO: write tests to meet "getting identifiers when are they returned?" section in https://docs.google.com/document/d/15jYA2wJLlzaO2K8M7_AEuZg7q3moq38MOB12uti0cxc/edit#heading=h.8otfkd6cs53t 

TODO: write test to prove activity default is being applied to items when updating via 1.2


amontenegro:
Looking at this

PR #2689, opened 2016-03-18
Card2628

amontenegro:


amontenegro:
https://trello.com/c/fMMmpug3/2628-rc2-identifiers-types-listed-in-an-unexpected-case-don-t-save-correctly

PR #2699, opened 2016-03-23
Checking the request url from the httpRequest not from the UriInfo, w…

amontenegro:
…hich cant know if the request was secure or not


amontenegro:
https://trello.com/c/K05L6GSD/2673-calls-to-deprecated-records-with-the-location-option-l-on-2-0-return-an-error-that-calls-must-be-made-on-https

PR #2706, opened 2016-03-29
Set initial visibility value on empty bio sections

amontenegro:


amontenegro:
https://trello.com/c/uU1kQaJp/2699-in-the-ui-person-fields-initially-have-no-visibility-set

PR #2712, opened 2016-03-30
Security question fix

hexplus:


hexplus:
Security Questions refactored to use ng-options instead of <#list ..., now the blank option is not shown any more.

https://trello.com/c/hmriOQDd/2698-after-saving-a-security-question-the-question-field-goes-blank

PR #2724, opened 2016-03-31
Check for duplicate researcher urls on the UI

amontenegro:


amontenegro:
https://trello.com/c/J8QafUfB/2722-adding-duplicate-websites-does-not-show-error-messages-in-ui

PR #2743, opened 2016-04-06
auth form testing issue

rcpeters:


orcid-integration-test/src/test/java/org/orcid/integration/blackbox/api/PopulateOAuthSignInCodeIntegrationTest.java
```diff
@@ -193,16 +195,21 @@ public void givenAndFamilyNamesPrepopulate() throws JSONException, InterruptedEx
         // test populating form with family name
         url = authorizeScreen + "&family_names=test_family_names";
         webDriver.get(url);
-        element = By.xpath("//input[@name='familyNames']");
+        element = By.xpath("//input[@ng-model='registrationForm.familyNames.value']");
         (new WebDriverWait(webDriver, DEFAULT_TIMEOUT_SECONDS)).until(ExpectedConditions.presenceOfElementLocated(element));                        
         assertTrue(webDriver.findElement(element).getAttribute("value").equals("test_family_names"));
```
amontenegro: Hey @rcpeters, if angular doesn't always populate the value, then we should remove this line (200) and leave only the assert on line 203 right?

rcpeters: No, we can still verify the model is upto date client side.

PR #2744, opened 2016-04-07
Return null as target url from a saved request that have /signout

amontenegro:


amontenegro:
https://trello.com/c/5inbrR9Q/2735-on-https-qa-orcid-org-signout-your-first-attempt-to-log-in-just-reloads-the-page

wjrsimpson:
@amontenegro I think I have a more through fix for this general problem in https://github.com/ORCID/ORCID-Source/pull/2745. See https://github.com/ORCID/ORCID-Source/blob/federatedLogInStuff/orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidUrlManager.java.

amontenegro:
Great @wjrsimpson! Yea, I was going to bring that up to our stand up, that we need a better way to handle these cases, but your pull request totally fix it, so, I will close this pull request and reference the card in yours.

PR #2750, opened 2016-04-08
Improve PublicProfileVisibilityTest

rcpeters:


rcpeters:
Could use more improvements. However this was good stopping point for now. Created new methods 
* angularHasFinishedProcessing
* noSpinners
* ngAwareClick
That should be hand for other test.

PR #2752, opened 2016-04-08
Upgrade selenium

rcpeters:


rcpeters:
@amontenegro Test the branch locally first just to make you we are not hosing the windows folks. Don't forget the maven update in eclipse. LMK if something doesn't work.

amontenegro:
ok, am looking at this

PR #2758, opened 2016-04-12
Add some keys to the list of need translation

amontenegro:


rcpeters:
@amontenegro cs was updated last night. I think it will also need keys.

amontenegro:
@rcpeters done!

PR #2765, opened 2016-04-14
Set max 5GB disk space limit for caches.

wjrsimpson:


orcid-core/src/main/java/org/orcid/core/utils/OrcidEhCacheFactoryBean.java
```diff
@@ -154,6 +165,8 @@ private CacheConfiguration createConfig() {
         CacheConfiguration config = new CacheConfiguration();
         config.setName(this.cacheName);
         config.setMaxEntriesLocalHeap(this.maxElementsInMemory);
+        config.setMaxElementsOnDisk(maxElementsOnDisk);
```
rcpeters: oops good catch!

PR #2776, opened 2016-04-21
Flush before merging the existing profile, so, it generates the pendi…

amontenegro:
…ng ids on biography elements before merge


amontenegro:
https://trello.com/c/Ghx1gzeQ/2770-error-when-adding-multiple-personal-identifiers-to-records

PR #2785, opened 2016-04-26
Remove old tables we dont need anymore for works

amontenegro:


amontenegro:
https://trello.com/c/wgf3WJXO/2760-remove-the-old-work-ext-ids-table-from-all-envs

PR #2787, opened 2016-04-27
Set visibility on names when creating

amontenegro:


amontenegro:
https://trello.com/c/LzqjivJI/2781-activity-items-when-the-user-is-the-source-the-id-is-listed-instead-of-their-name

PR #2791, opened 2016-04-28
Remove delegation from v2

wjrsimpson:
https://trello.com/c/TraHRL5w/2797-remove-delegation-for-v2


orcid-model/src/main/java/org/orcid/jaxb/model/record_rc2/Record.java
```diff
@@ -54,7 +54,7 @@
     protected Person person;
     @XmlElement(namespace = "http://www.orcid.org/ns/activities", name = "activities-summary")
     protected ActivitiesSummary activitiesSummary;
-    @XmlAttribute
+    @XmlTransient
```
amontenegro: Hey @wjrsimpson, I think this should include also a @JsonIgnore

Thanks

wjrsimpson: @amontenegro I was thinking that too, but when I tried, it seemed to work with just @XmlTransient.

amontenegro: Intresting! so, this is good to go!

PR #2792, opened 2016-04-28
Card2795

amontenegro:


amontenegro:
https://trello.com/c/FXZLYInl/2795-can-get-a-2-step-token-to-read-limited-activities

amontenegro:
@rcpeters @wjrsimpson  please take a look at this pull request, if the request is done without any scope or after clearing the scopes we end up without any scope, we will throw an exception of type IllegalArgumentException indicating that the scopes used are not valid for this kind of grant type.

Please let me know your comments.

amontenegro:
Done merge, this could cause some issues

PR #2800, opened 2016-05-02
The method to create a new bio should start with persist so it is cau…

amontenegro:
…ght by the last modified aspect


amontenegro:
https://trello.com/c/WSXg6A14/2807-on-new-records-the-biography-doesn-t-display-on-the-public-record-or-on-the-api-until-you-sign-out-of-the-record

PR #2810, opened 2016-05-05
Fix the orer in which we get the fundings before display them in the UI

amontenegro:


amontenegro:
https://trello.com/c/yV4UOTRz/2806-preferred-source-isn-t-saving-for-funding

PR #2830, opened 2016-05-11
Allowing adding multiple addresses using the 1.2 api

amontenegro:


amontenegro:
https://trello.com/c/nudguiGo/2843-allow-creating-multiple-addresses-using-api-1-2

PR #2861, opened 2016-05-26
remove usused method

rcpeters:


amontenegro:
Since no code errors, it should be good to go

PR #2870, opened 2016-05-27
Fix visibility when null

amontenegro:


amontenegro:
https://trello.com/c/ysagpsXQ/2865-putting-an-other-name-with-2-0-rc2-removes-the-visibility-if-it-is-not-especified

PR #2871, opened 2016-05-30
Fix error on UI that allow you to create other names, keywords or res…

amontenegro:
…earcher urls without visibility


amontenegro:
https://trello.com/c/iSvYmDND/2774-some-other-names-keywords-researcher-urls-and-addresses-are-being-saved-with-null-visibility-on-the-db

PR #2878, opened 2016-05-31
Remove primary indicator from address

amontenegro:


amontenegro:
https://trello.com/c/FUEF6Bu2/2868-remove-primary-country-from-v2-0-rc2-just-assume-first-in-the-order-of-countries-is-primary

PR #2880, opened 2016-06-01
Fixing countries display for v2

hexplus:
https://trello.com/c/bbwvfLXr/2878-v2-public-record-only-shows-primary-country


amontenegro:
https://trello.com/c/DrhNCTIO/2894-remove-primary-country-selector-from-ui

PR #2890, opened 2016-06-03
Fix the recirect url variable

amontenegro:


amontenegro:
https://trello.com/c/QXveG7YE/2724-there-is-no-response-when-clicking-the-deny-button-during-registration-on-oauth

PR #2892, opened 2016-06-06
If visibility is null we have a problem, so, we should return the nul…

amontenegro:
…l one instead of a default value


amontenegro:
https://trello.com/c/645WMewc/2900-rc2-external-identifiers-updates-return-a-visibility-error-when-no-visibility-is-specified

PR #2893, opened 2016-06-07
Allow clients to see private activities they are the source of

amontenegro:


amontenegro:
https://trello.com/c/zDJ0uygW/2897-1-2-api-does-not-return-private-activities-which-the-client-is-the-source-of

orcid-core/src/main/java/org/orcid/core/security/visibility/aop/OrcidApiAuthorizationSecurityAspect.java
```diff
@@ -59,6 +58,7 @@
 /**
  * @author Declan Newman (declan) Date: 16/03/2012
  */
+@Deprecated
```
amontenegro: Lets mark this class as deprecated, we should not use it anymore in 2.0 API

orcid-core/src/main/java/org/orcid/core/security/visibility/aop/OrcidApiAuthorizationSecurityAspect.java
```diff
@@ -366,5 +370,29 @@ private boolean allowAnonymousAccess(Authentication auth, AccessControl accessCo
...
+    /**
+     * Check if a member have a specific scope over a client
+     * 
+     * @param clientId
+     * @param userName
+     * @param scopes
+     * @return true if the member have access to any of the specified scope on the specified user
+     * */
+    @Deprecated
+    private boolean checkIfScopeIsAvailableForMember(String clientId, String userName, List<String> requiredScopes) {
```
amontenegro: This function is tricky, it will return true if the client have any of the requiredScopes available for the specified user.
We should not use it anywhere, thats why it is private.

orcid-core/src/main/java/org/orcid/core/security/visibility/aop/OrcidApiAuthorizationSecurityAspect.java
```diff
@@ -318,13 +321,14 @@ private boolean isUpdateOrReadScope(ScopePathType requiredScope) {
             return false;
         }
     }
-
-    private boolean hasScopeEnabled(String clientId, String userName, String scope, String equivalentScope) {
+    
+    @Deprecated
+    public boolean hasScopeEnabled(String clientId, String userName, String scope, String equivalentScope) {
```
amontenegro: Make it public so we can do a round of unit testing on it, but we should not use it anywhere outside this class.

orcid-core/src/main/java/org/orcid/core/oauth/service/OrcidOauth2TokenDetailServiceImpl.java
```diff
@@ -178,26 +183,7 @@ public void removeConflictsAndCreateNew(OrcidOauth2TokenDetail detail) {
...
-    
-    /**
-     * Check if a member have a specific scope over a client
-     * @param clientId
-     * @param userName
-     * @param scopes
-     * @return true if the member have access to any of the specified scope on the specified user
-     * */
-    @Override
-    public boolean checkIfScopeIsAvailableForMember(String clientId, String userName, List<String> scopes) {
```
amontenegro: Lets remove this function from the OrcidOauth2TokenDetailServiceImpl since we only need it in the OrcidApiAuthorizationSecurityAspect and it is actually a risky function to use.

rcpeters:
@wjrsimpson please still review even if it's already merged.

PR #2896, opened 2016-06-07
Modify the copy manager so when it copies the bio it only keep the el…

amontenegro:
…ements in the request


amontenegro:
https://trello.com/c/JmjFJpvr/2898-1-2-api-bio-update-adds-new-items-instead-of-updating-existing-ones

PR #2901, opened 2016-06-08
Card2909

amontenegro:


amontenegro:
https://trello.com/c/K7XV6yWP/2909-updating-an-address-in-2-0-causes-the-record-not-to-display-in-the-ui

PR #2915, opened 2016-06-13
Allow creating duplicated fundings from the UI

amontenegro:


amontenegro:
https://trello.com/c/FtREdJ7V/2920-users-can-t-add-two-funding-items-with-the-same-identifier

PR #2917, opened 2016-06-13
Card2921

amontenegro:


amontenegro:
https://trello.com/c/hjVqHN9C/2921-when-adding-an-item-to-a-bio-section-via-the-api-increase-existing-items-display-index-by-1

PR #2918, opened 2016-06-14
Fix display index in1.2 works and funding

amontenegro:


amontenegro:
https://trello.com/c/wBK6Hygb/2925-when-adding-works-or-funding-with-the-1-2-api-increase-existing-items-display-index-by-1

PR #2922, opened 2016-06-15
don't call retrieveLastModifiedDate in the DAO

rcpeters:


orcid-core/src/main/java/org/orcid/core/manager/impl/ProfileEntityManagerImpl.java
```diff
@@ -609,17 +609,16 @@ public Date getLastModified(String orcid) {
         if (sra != null)
             lastMod = (Date)sra.getAttribute(ProfileLastModifiedAspect.REQUEST_PROFILE_LAST_MODIFIED, ServletRequestAttributes.SCOPE_REQUEST);
         if (lastMod == null) {
-            return lastMod = updateLastModifed(orcid);
```
rcpeters: that was likely a critical mistake

PR #2932, opened 2016-06-17
Check the user cant hold more than 10000 works

amontenegro:


amontenegro:
https://trello.com/c/HW7tldkz/2932-return-an-error-when-adding-works-to-a-record-with-more-than-10000-works-already

rcpeters:
@caoyler please inform the API users group this is going out this week.

PR #2937, opened 2016-06-19
Second go at caching last modified per a request

rcpeters:


rcpeters:
Not 100% sure this is worth the complexity.

PR #2945, opened 2016-06-23
Refactor to delete activity in group correctly

rcpeters:
https://trello.com/c/WKzqdLuL/2947-funding-in-a-group-is-deleted-the-grouping-needs-to-be-rechecked
https://trello.com/c/tLThAffa/2580-review-refreshing-works-after-deletion

### DIFF
--- a/orcid-web/src/main/webapp/static/javascript/angularOrcid.js
+++ b/orcid-web/src/main/webapp/static/javascript/angularOrcid.js
@@ -69,6 +69,53 @@ PRIVACY.PUBLIC = 'PUBLIC';
 PRIVACY.LIMITED = 'LIMITED';
 PRIVACY.PRIVATE = 'PRIVATE';
 
+var GroupedActivitiesUtil = function() {
+};
+
+GroupedActivitiesUtil.prototype.group = function(activity, type, groupsArray) {
+    var matches = new Array();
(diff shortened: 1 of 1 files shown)


orcid-web/src/main/webapp/static/javascript/angularOrcid.js
```diff
@@ -738,9 +781,7 @@ orcidNgModule.factory("fundingSrvc", ['$rootScope', function ($rootScope) {
                 var rmFunding;
                 for (var idx in fundingSrvc.groups) {
                     if (fundingSrvc.groups[idx].hasPut(putCode)) {
-                        rmFunding = fundingSrvc.groups[idx].rmByPut(putCode);
-                        if (fundingSrvc.groups[idx].activitiesCount == 0)
-                            fundingSrvc.groups.splice(idx,1);
+                        rmFunding = fundingSrvc.groups[idx].getByPut(putCode);
```
hexplus: Is getByPut correct? I think it should be rmByPut.

rcpeters: Look at what happens once we get it 6 lines down.

hexplus: Yeah, that make sense.

PR #2947, opened 2016-06-23
first pass at improving grouping speed

TomDemeranville:
Modifies algorithm to use a lookup rather than recursively iterating lists.  Scales in a more linear fashion and provides massive performance improvements seen with large sets of works.  Small sets will see no improvement. Order of groups is slightly different but contents the same.

https://trello.com/c/xWbVxENG/2949-refactor-the-way-grouping-is-done-to-improve-performance-over-large-groups-of-works#

For a user with 16k works, this change makes the request 90 seconds faster on my dev machine, halving request time with a cold cache and reducing request time to a few seconds with a warm one.
Before cold: 03:14
Before warm: 01:32
After cold: 01:43
After warm: 00:04

Tested against record with:
workA: doi:1
workB: doi:2
workC: no external id
workD: agr:1
workE: doi:1
workF: doi:2, agr:1

And groups correctly into three groups: 
group (doi:1) = A, E
group (no id) = C
group (doi2 + agr:1) = B, D, F

### DIFF
--- a/orcid-core/src/main/java/org/orcid/core/utils/activities/ActivitiesGroupGenerator.java
+++ b/orcid-core/src/main/java/org/orcid/core/utils/activities/ActivitiesGroupGenerator.java
@@ -17,13 +17,18 @@
 package org.orcid.core.utils.activities;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
+import org.orcid.jaxb.model.record_rc2.GroupAble;
 import org.orcid.jaxb.model.record_rc2.GroupableActivity;
 
 public class ActivitiesGroupGenerator {    
 
     private List<ActivitiesGroup> groups = new ArrayList<ActivitiesGroup>();
+    
+    private Map<GroupAble, ActivitiesGroup> lookup = new HashMap<GroupAble, ActivitiesGroup>();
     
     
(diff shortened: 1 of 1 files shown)


orcid-core/src/main/java/org/orcid/core/utils/activities/ActivitiesGroupGenerator.java
```diff
@@ -17,50 +17,66 @@
...
             //If it is not the first activity, check which groups it belongs to
-            List<Integer> belongsTo = new ArrayList<Integer>();
-            for(int i = 0; i < groups.size(); i++) {
-                ActivitiesGroup group = groups.get(i);
-                if(group.belongsToGroup(activity)) {
-                    belongsTo.add(i);
-                }
+            List<ActivitiesGroup> belongsTo = new ArrayList<ActivitiesGroup>();
+            ActivitiesGroup thisGroup = new ActivitiesGroup(activity);
+            for (GroupAble g :thisGroup.getGroupKeys()){
```
rcpeters: So we have a group with 150,000 works with the same key + a unique key what happens?

TomDemeranville: Well, they should all be in the same group.

PR #2951, opened 2016-06-24
Couple of API optimisations 

TomDemeranville:
1: modified createWork to use the cached version of already transformed works if available, similarly modified update works to use actual last updated instead of system.getTime.  Will not help populate the cache as the cache is invalidated afterwards (because the methods modify the record!).  Also improves consistency as they both work the same way now.

2: Moved factories in JSONUtils and DateUtils (ObjectMapper, DatatypeFactory) to static context to avoid recreating every time they're needed.  Saves a lot of processor cycles! (47s request became 27s after change)

Note that ObjectMapper is guaranteed to be thread safe, DatatypeFactory is not.  However, it appears that ORACLE implementations of DatatypeFactory _are_ thread safe (and I expect so are all the others) because otherwise everyone would be complaining due to hefty initialisation time.


orcid-core/src/main/java/org/orcid/core/manager/impl/WorkManagerImpl.java
```diff
@@ -185,15 +185,15 @@ public Work createWork(String orcid, Work work, boolean applyAPIValidations) {
...
+            
             // If it is the user adding the peer review, allow him to add
             // duplicates
             if (!sourceEntity.getSourceId().equals(orcid)) {
-                if (works != null) {
-                    List<Work> workEntities = jpaJaxbWorkAdapter.toMinimizedWork(works);
-                    for (Work existing : workEntities) {
+                Date lastModified = profileEntityManager.getLastModified(orcid);
+                long lastModifiedTime = (lastModified == null) ? 0 : lastModified.getTime();
+                List<Work> existingWorks = this.findWorks(orcid, lastModifiedTime);
```
amontenegro: Hey @TomDemeranville did you tested this? I think the @cachable annotation is bypassed if the call is withing the same class: http://stackoverflow.com/questions/12115996/spring-cache-cacheable-method-ignored-when-called-from-within-the-same-class, thats why I didn't do this before.

Could you please give it a double check and confirm if the cache is used or not when called from the same class?

@rcpeters @wjrsimpson FYI

Thanks

orcid-core/src/main/java/org/orcid/core/manager/impl/WorkManagerImpl.java
```diff
@@ -240,7 +240,11 @@ public Work updateWork(String orcid, Work work, boolean applyAPIValidations) {
         
         if (applyAPIValidations) {
             activityValidator.validateWork(work, sourceEntity, false, applyAPIValidations, workEntity.getVisibility());
-            List<Work> existingWorks = this.findWorks(orcid, System.currentTimeMillis());            
+            
+            Date lastModified = profileEntityManager.getLastModified(orcid);
+            long lastModifiedTime = (lastModified == null) ? 0 : lastModified.getTime();
+            List<Work> existingWorks = this.findWorks(orcid, lastModifiedTime);
```
amontenegro: Same as my first comment; the cache might be ignored from calls withtin the same class

rcpeters: That issue existed before this change. We can take care of that in a different change set (or we can refactor to do entity comparisons instead.....).

TomDemeranville: @amontenegro is right here - it's not something I knew.  What I did was saw that update was using findWorks and create wasn't.  So I changed them to both use it and assumed the cache would work and didn't test to check.

amontenegro: Ok, so, lets go with this change and refactor later for best performance.

@TomDemeranville please fix the merge conflict

PR #2952, opened 2016-06-24
Fixing variable name due some backend changes that were made

hexplus:
**orcidId** variable was changed to **effectiveUserOrcid** in the backend but not in the Public Profile, fixing this.

https://trello.com/c/NlcZFZaG/2928-activities-vs-bio-how-to-display-the-source-when-the-name-is-private


rcpeters:
thanks!

PR #2955, opened 2016-06-27
Preserve works order after streamming and when an element might be re…

amontenegro:
…freshed in the cache


amontenegro:
https://trello.com/c/6uw7SAn0/2954-preferred-sources-aren-t-saving-for-works

PR #2961, opened 2016-06-29
Modify required scope, so it allows /orcid-bio/read-limited to read b…

amontenegro:
…io limited data


amontenegro:
https://trello.com/c/UniMq6jQ/2922-rc2-orcid-bio-read-limited-scope-does-not-return-the-biography-field

PR #2962, opened 2016-06-29
JSON end point for notifications alerts

wjrsimpson:


rcpeters:
should we add by type?

PR #2963, opened 2016-06-29
Move source to the address and then modify the filter to allow privat…

amontenegro:
…e addresses to be displayed to the source


amontenegro:
https://trello.com/c/ih8pP4Ic/2924-private-addresses-not-returned-with-1-2-when-call-by-the-client-who-is-the-source

amontenegro:
Hey @rcpeters @wjrsimpson, could you please take a look at this one?

On 1.2 the Address object used to have the source and, the country inside it had the visibility element, so , the VisibilityFilter couldn't parse this case, and that's why it didn't return the private address event for the client source of it.

With this PR Im moving the source element inside the Country, and, modified the VisibilityFilter to check the Country inside the Address element

PR #2964, opened 2016-06-29
When an url comes null from the UI, create the url element but let th…

amontenegro:
…e value null, so, the orika map works


amontenegro:
https://trello.com/c/odIvHkgk/2961-cant-remove-an-url-from-an-existing-work

PR #2980, opened 2016-07-05
Store non public names as empty strings, so we know we already have t…

amontenegro:
…he name but it is not public


amontenegro:
https://trello.com/c/UYEhMdIu/2968-source-name-cache-cant-deal-with-non-public-names

PR #2986, opened 2016-07-07
Disable token when expired

amontenegro:


orcid-core/src/main/java/org/orcid/core/oauth/service/OrcidTokenStoreServiceImpl.java
```diff
@@ -132,7 +133,7 @@ public OAuth2AccessToken readAccessToken(String tokenValue) {
      *            The token to remove from the database.
      */
     @Override
-    @Transactional
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
```
amontenegro: Hey @wjrsimpson could you please take a look at this one? So, with Propagation.REQUIRES_NEW we create a separate transaction to persist the disable token query.

amontenegro:
https://trello.com/c/x6TFBPYm/2974-tokens-cant-be-marked-as-disabled

orcid-core/src/main/java/org/orcid/core/oauth/service/OrcidTokenStoreServiceImpl.java
wjrsimpson: I think this is OK, if the outer transaction does not write to the database row for the token. If the outer transaction has a write lock on the row, then the new transaction will not be able to write to the row, and then there would be deadlock.

I suspect that the outer transaction does not write to the row, so this is OK.

See also https://github.com/spring-projects/spring-security-oauth/issues/399.

PR #2988, opened 2016-07-08
only auto approve if the matching token is a long life tokens

rcpeters:


amontenegro:
Looks good! @rcpeters do you want this in this release or in the next one?

PR #2996, opened 2016-07-12
Fixed If multiple permission for the same scopes and same client exist, revoking one does not revoke the other

amontenegro:


amontenegro:
https://trello.com/c/rKc8zf1E/2981-if-multiple-permission-for-the-same-scopes-and-same-client-exist-revoking-one-does-not-revoke-the-other

rcpeters:
what is  #2996 :-)

amontenegro:
@rcpeters I just put the card link, this will allow to delete all tokens with the same scopes+client at the same time.

I think @hexplus is working on a separate card to hide repeated tokens form the UI

PR #3022, opened 2016-07-20
Enable refresh token

amontenegro:


amontenegro:
https://trello.com/c/avx1qtWo/2976-implement-oauth-token-grant-type-refresh-token

PR #3026, opened 2016-07-21
Show the right error message on the oauth screen when for when the us…

amontenegro:
…er is deactivated or unclaimed


amontenegro:
https://trello.com/c/JWNMdjE4/2869-oauth-sign-in-errors-when-trying-to-log-in-with-a-unclaimed-or-deactivated-email-address-are-missing

PR #3034, opened 2016-07-26
Use getOrcidIdBase function to fill the full path of the source eleme…

amontenegro:
…nt on every item


amontenegro:
https://trello.com/c/JOH0qr23/3005-uri-and-host-missing-from-source-in-1-2

PR #3056, opened 2016-08-04
Centralize get access token in a single class

amontenegro:


amontenegro:
@wjrsimpson could you please take a look?

PR #3067, opened 2016-08-16
Validate work type is not null

amontenegro:


amontenegro:
https://trello.com/c/aSOagbhO/3028-validate-that-works-created-in-2-0-xml-and-json-have-a-type-if-not-throw-an-exception

PR #3077, opened 2016-08-23
Strip all html when required

amontenegro:


amontenegro:
https://trello.com/c/Nr578EM7/3039-strip-all-html-from-names-but-allow-simple-text-html-for-client-descriptions-and-custom-emails

PR #3083, opened 2016-08-25
Convert unicode escapes utf8

lizkrznarich:
https://trello.com/c/0wDxSj7T/2929-portugese-translations-from-transifex-are-broken


lizkrznarich:
@wjrsimpson @rcpeters Spacing in email templates needs fixing - will be easier to fix if we don't merge this yet, so closing and will submit a new PR.

PR #3093, opened 2016-08-30
Changed work sorting to be reverse chronological in 1.2 API.

wjrsimpson:
https://trello.com/c/BBIDxfpP/3011-update-1-2-api-to-use-work-entity-cache


rcpeters:
Just to verify you are still using display index to sort right?

wjrsimpson:
@rcpeters Yes, but by pub date (reverse chronological) and then title, and then display index.

The display index is not in the comparator, but the works are pulled from the db ordered by display index.

rcpeters:
Display index should be first.

PR #3094, opened 2016-08-30
Decode amp, single quote and double quote from names

amontenegro:


rcpeters:
You need to chat with Will. Doesn't match what we need to do.

amontenegro:
Ok, updated! @wjrsimpson please verify 

@rcpeters FYI

rcpeters:
Doesn't look right. Also we need some test to verify.

PR #3097, opened 2016-09-05
Custom validation for work bulk

amontenegro:


amontenegro:
https://trello.com/c/QF6GTFzv/3044-for-discussion-how-to-handle-errors-with-one-work-when-bulk-adding-works

PR #3101, opened 2016-09-07
Applying some validations that where previously applied in the XSD va…

amontenegro:
…lidation, but now, given the bulk, should be applied on the manager


amontenegro:
https://trello.com/c/JCYt219f/3058-bulk-work-lets-you-add-works-with-bad-metadata

amontenegro:
https://trello.com/c/JncjXPjt/3059-500-error-when-bulk-adding-works-if-one-work-has-a-missing-common-external-id-value-field

PR #3102, opened 2016-09-07
Validate publication dates values

amontenegro:


amontenegro:
https://trello.com/c/NosHlBDo/3061-can-add-invalid-dates-with-bulk-works

PR #3113, opened 2016-09-13
Card3064

amontenegro:


amontenegro:
https://trello.com/c/Rl8KmSUn/3064-rc1-requires-the-relationship-field-on-work-identifiers-rc2-does-not

PR #3116, opened 2016-09-14
Enforce person ext id to have the relationship field

amontenegro:


amontenegro:
https://trello.com/c/HQbfWtqK/3071-personal-identifiers-can-be-added-without-a-relationship-tag

PR #3120, opened 2016-09-15
Fix digest email db problem

wjrsimpson:
https://trello.com/c/2GqvM3kD/3074-notifications-being-sent-more-frequently-than-user-selected-setting


orcid-core/src/main/java/org/orcid/core/manager/impl/EmailMessageSenderImpl.java
```diff
@@ -213,7 +216,11 @@ private void flagAsSent(List<Notification> notifications) {
         for (Notification notification : notifications) {
             notificationIds.add(notification.getPutCode());
         }
-        notificationDao.flagAsSent(notificationIds);
+        List<List<Long>> batches = Lists.partition(notificationIds, 30000);
```
rcpeters: Integer.MAX_VALUE might be clearing as to why.

PR #3121, opened 2016-09-15
Added escaping to bad orgs email.

wjrsimpson:


wjrsimpson:
@rcpeters I already put this change in the patch, but please can it also go into master just in case.

PR #3124, opened 2016-09-16
Fixing error where the relationship field was null when a funding wit…

amontenegro:
…hout ext ids was being edited


amontenegro:
https://trello.com/c/oN8OXAcd/3078-can-t-save-changes-to-a-funding-item-without-an-identifier

PR #3126, opened 2016-09-16
Fix how we get the empty ext id

amontenegro:


amontenegro:
https://trello.com/c/LOKeYqce/3080-grant-number-added-to-one-funding-item-shows-up-on-edit-screen-of-any-funding-item-without-an-existing-grant-number

PR #3134, opened 2016-09-23
added pdb identifier using db only (no modify to enums in v2rc1 or 1.2)

TomDemeranville:
Behaviour:
You can add and get PDB ids on works rc2+.
You cannot add on rc1 or below.
If you query on older apis, you get OTHER-ID (checked on rc1 and v1.2).
Internationalisation is handled as before (for now).

bugfix to org/orcid/jaxb/model/record_rc1/WorkExternalIdentifierType.java


rcpeters:
Have we looked at the broken build?

TomDemeranville:
Apologies, now fixed.

PR #3135, opened 2016-09-23
Set bulk max size

amontenegro:


amontenegro:
https://trello.com/c/d9s9IiFS/3089-set-the-max-number-of-works-in-a-bulk-to-100

PR #3152, opened 2016-10-07
Try two hours for time to idle

rcpeters:


rcpeters:
@wjrsimpson what do you think about trying this out in prod?

PR #3171, opened 2016-10-20
Fixing some blackbox test and then update the scheduler config to ign…

amontenegro:
…ore sending messages to the MQ when messaging is disabled


amontenegro:
- Some xpath changes to fix black box tests
- Change scheduler config so it sends messages directly to SOLR and then send messages to the MQ when messaging is enable, so, we can feed S3
- Disable sending messages to S3 when messaging is disabled

PR #3172, opened 2016-10-20
Disabled institutional connection notification when client is no longer configured for that.

wjrsimpson:


wjrsimpson:
[Inbox will not load when there are notifications from a disabled institutional round-trip](https://trello.com/c/5dU1jKdv/3131-inbox-will-not-load-when-there-are-notifications-from-a-disabled-institutional-round-trip)

PR #3174, opened 2016-10-20
Adding some headers on Orcid20APIClient to fake https call

amontenegro:


orcid-message-listener/src/main/java/org/orcid/listener/clients/Orcid20APIClient.java
```diff
@@ -42,11 +43,14 @@
     protected Client jerseyClient;
     
     protected final URI baseUri;
+    
+    protected final String host;
 
     @Autowired
-    public Orcid20APIClient(@Value("${org.orcid.message-lisener.api20BaseURI}") String baseUri) throws URISyntaxException {
-        LOG.info("Creating Orcid20APIClient with baseUri = " + baseUri);
+    public Orcid20APIClient(@Value("${org.orcid.message-listener.api20BaseURI}") String baseUri, @Value("${org.orcid.message-listener.host}") String host) throws URISyntaxException {
```
rcpeters: `org.orcid.message-listener.host` should be `org.orcid.message-listener.host_header_override`

amontenegro: Fixed

PR #3181, opened 2016-10-24
Handle the deprecated and locked exception errors

amontenegro:


orcid-message-listener/src/main/java/org/orcid/listener/ReIndexListener.java
```diff
@@ -64,23 +69,49 @@
...
+                // Should never happen, since it is already deprecated
+            }
+            return;
+        }
+                
+        //Fetch 2.0 record
+        try {
+            record = orcid20ApiClient.fetchPublicProfile(orcid);
+        } catch(Exception e) {
+            LOG.warn("Unable to fetch record " + orcid + " from 2.0 API");
```
rcpeters: error or warn?

amontenegro: Warn for now, on this phase (phase 1), if we where able to process the 1.2 record, we are fine, so, if something wrong happens fetching 2.0 we just log it as a warn.

On phase 2 the idea is to have some place (a DB) to write an error and a status to indicate the record was not sent to S3, but, for phase 1 is ok to just log a warn message

PR #3183, opened 2016-10-24
Add a listener to configure log4j for the messaging project

amontenegro:


amontenegro:
@wjrsimpson I think you have configured log4j for new projects before ... could you please take a look at this one? the message listener log goes directly to cataline.out file which is wrong, it should create its own file. Is this all we need to configure from the spring/java side?

The server config and the log4j.xml files are already in the servers.

Thanks

PR #3187, opened 2016-10-25
Remove jersey https config from the message listener

amontenegro:


amontenegro:
@wjrsimpson could you please take a look at this? I cant make it work locally yet

amontenegro:
Look at https://github.com/ORCID/ORCID-Source/pull/3188

PR #3188, opened 2016-10-25
Feature flag to enable disable jersey ssl on message listener

amontenegro:


amontenegro:
@rcpeters @wjrsimpson please take a look at this.

The idea is that the jerseyClient will be created depending on the feature flag "org.orcid.message-listener.development_mode" 

So, if that variable is true, we will configure the full SSL context
If it is false or doesn't exists (so, we dont have to bother about adding it to any of the servers), it will create a default client with no SSL context.

Please let me know your comments.

Thanks

PR #3194, opened 2016-10-26
Include 'peer-review service' as a group id type

amontenegro:


amontenegro:
[Add "peer-review service" to the list of group-types for group-id-type](https://trello.com/c/2aUoFc5I/3151-add-peer-review-service-to-the-list-of-group-types-for-group-id-type)

PR #3195, opened 2016-10-26
Feature flag to enable disable jersey ssl on message listener with will changes

wjrsimpson:


wjrsimpson:
@amontenegro This enables local dev with self signed certs.

rcpeters:
👍

PR #3196, opened 2016-10-26
Card3103

amontenegro:


amontenegro:
[Create a path for members to read just employment, education, peer-review, funding, and works instead of the entire activities section](https://trello.com/c/1x5DfyUi/3103-create-a-path-for-members-to-read-just-employment-education-peer-review-funding-and-works-instead-of-the-entire-activities-secti)

PR #3204, opened 2016-10-28
Move email params, so, the orcid id is the first one and the last mod…

amontenegro:
…ified aspect can pick it


amontenegro:
Hi @wjrsimpson,

Please take a look at this small change:

When we create a brand new record, the emailDao.addSourceToEmail params are the email and then the source id.

After this method executes, the ProfileLastModifiedAspect is invoked, but, given that the first param is the email, and, the aspect uses that first param as the orcid id (@AfterReturning(POINTCUT_DEFINITION_BASE + " && args(orcid, ..)")), we try to update the last modified and the indexing status using the email as param, which is wrong.

This PR changes the order of the params, however, we might also consider ignoring the last modified aspect from the emailDao.addSourceToEmail, since the source never changes and it happens only when creating a user, which will update the indexing status and last modified elsewhere.

Thanks

PR #3205, opened 2016-10-28
remove message listener calls from the last modified aspect

amontenegro:


amontenegro:
[Message listener: New records are not being sent to the message listener](https://trello.com/c/yUFPlnuP/3185-message-listener-new-records-are-not-being-sent-to-the-message-listener)

PR #3206, opened 2016-10-28
Manual test102016 update

caoyler:
Updates to make the instructions easier to follow and switched create to use json so an XML is not needed.


caoyler:
[Help Jeff run through the manual test.](https://trello.com/c/MivtcelG/3094-help-jeff-run-through-the-manual-test)

PR #3207, opened 2016-10-31
Fix namespace on educations and employments elements

amontenegro:


amontenegro:
[When reading /employments and /educations the namespace is missing or given as ns2](https://trello.com/c/5NMFn2kh/3186-when-reading-employments-and-educations-the-namespace-is-missing-or-given-as-ns2)

PR #3209, opened 2016-10-31
Modify the message listener so it uses the members API for fetching p…

amontenegro:
…ublic records


amontenegro:
[Modify the message listener project to use the API](https://trello.com/c/7w5RMfTg/3182-modify-the-message-listener-project-to-use-the-api)

PR #3210, opened 2016-10-31
Remove dangerous liquibase script

amontenegro:


amontenegro:
[Remove empty external identifiers from works](https://trello.com/c/JHw1VzvM/3145-remove-empty-external-identifiers-from-works)

PR #3214, opened 2016-11-02
Modify the code so not all elements are in memory at the same time

amontenegro:


amontenegro:
[Modify the code so we dont need that much memory](https://trello.com/c/Umui63N6/3191-modify-the-code-so-we-dont-need-that-much-memory)

orcid-message-listener/src/main/java/org/orcid/listener/ReIndexListener.java
```diff
@@ -98,26 +98,29 @@ public void processMessage(final Map<String, String> map) throws JsonProcessingE
...
-            record = orcid20ApiClient.fetchPublicProfile(orcid);
-        } catch(Exception e) {
-            LOG.warn("Unable to fetch record " + orcid + " from 2.0 API");            
-        }
-        
+             
         if(profile != null) {
             //Update solr
             if(isSolrIndexingEnalbed) {
                 solrIndexUpdater.updateSolrIndex(record,profile.toString());
```
TomDemeranville: This is now going to have horrible null pointer exceptions.  SOLR indexing requires 1.2 and 2.0.

orcid-message-listener/src/main/java/org/orcid/listener/listeners/updated/UpdatedOrcidWorker.java
```diff
@@ -69,23 +69,20 @@
...
                 if(dumpIndexingEnabled) {
-                    updateS3(orcid, profile, record);    
+                    updateS3_1_2_API(orcid);
+                    updateS3_2_0_API(orcid);
                 }
                 
                 // Phase # 2 - update solr
                 if(solrIndexingEnabled) {
-                    updateSolr(orcid, record, profile.toString());    
+                    updateSolr(orcid);
```
TomDemeranville: Doing it this way will mean 4 requests to the API instead of 2.

orcid-message-listener/src/main/java/org/orcid/listener/ReIndexListener.java
amontenegro: Uf thats true! Let me fix the PR, however, we dont have to worry too much, since "isSolrIndexingEnalbed" is false until we decide to use enable it

PR #3215, opened 2016-11-02
Setting concurrency for message listener

amontenegro:


amontenegro:
[Should we enable multi-threading on the message listener?](https://trello.com/c/vnw8AVQO/3181-should-we-enable-multi-threading-on-the-message-listener)

PR #3216, opened 2016-11-02
ci: archive war files + slack notification

jpeerz:
* archive all war modules
* include slack notification at the end of wf


jpeerz:
@TomDemeranville @amontenegro @rcpeters 

Please include on next push, in order to save disk space and get slack notifications :)

PR #3217, opened 2016-11-03
added lens ID

TomDemeranville:
https://trello.com/c/SA7ATSxf/3184-create-work-identifier-lensid


TomDemeranville:
this is failing as a file is missing: Error Reading Migration File: class path resource [db/updates/fix_empty_work_external_ids.xml] cannot be resolved to URL because it does not exist

PR #3218, opened 2016-11-03
Remove fix_empty_work_external_ids.xml because we dont have that scri…

amontenegro:
…pt and we dont want it in prod


amontenegro:
Hey @wjrsimpson @rcpeters FYI, it looks like one PR we merged had a liquibase script we removed here: https://github.com/ORCID/ORCID-Source/pull/3210

rcpeters:
yikes!!!! @TomDemeranville and @wjrsimpson

PR #3220, opened 2016-11-03
Fix marshalling profile when multithreading

amontenegro:


amontenegro:
https://jaxb.java.net/guide/Performance_and_thread_safety.html

PR #3221, opened 2016-11-04
Public record no longer returned when client uses wrong token to acce…

bobcaprice:
…ss it

Previously returning public record of requested user, now returns 401 unauthorized


bobcaprice:
[2.0 Should return an error when reading a record with a token for a different record](https://trello.com/c/FDunTlSp/3157-2-0-should-return-an-error-when-reading-a-record-with-a-token-for-a-different-record)

PR #3224, opened 2016-11-07
Reactivation workflow

wjrsimpson:


wjrsimpson:
[Implement Re-activate workflow for deactivated records](https://trello.com/c/MWm13x69/3134-implement-re-activate-workflow-for-deactivated-records)

wjrsimpson:
See https://github.com/ORCID/ORCID-Source/pull/3233.

PR #3233, opened 2016-11-09
Reactivation workflow part2

wjrsimpson:


wjrsimpson:
[Implement Re-activate workflow for deactivated records](https://trello.com/c/MWm13x69/3134-implement-re-activate-workflow-for-deactivated-records)

PR #3234, opened 2016-11-09
Member list fix filters

lizkrznarich:
https://trello.com/c/cEwUnB3y/3196-member-list-don-t-look-at-description-data-when-filtering


jpeerz:
[Member list: Don't look at description data when filtering](https://trello.com/c/cEwUnB3y/3196-member-list-don-t-look-at-description-data-when-filtering)

PR #3240, opened 2016-11-10
Auto deprecate

amontenegro:


amontenegro:
[Implement new auto deprecate workflow for members unclaimed iDs](https://trello.com/c/ouHyr0mp/3144-implement-new-auto-deprecate-workflow-for-members-unclaimed-ids)

PR #3250, opened 2016-11-21
Return error for orcid profile wrong token

bobcaprice:
Refactored how MemberV2ApiServiceDelegatorImpl carries out security checks, allowing non-client credentials scope read only access to data


bobcaprice:
[Read-public tokens return an error on member API 2.0](https://trello.com/c/tWXd45Lz/3208-read-public-tokens-return-an-error-on-member-api-2-0)

rcpeters:
merge monster

PR #3243, opened 2016-11-11
Return error for orcid profile wrong token

bobcaprice:
Changed so that no information is returned when wrong token presented for:

- view activities
- view researcher urls
- view emails
- view other names
- view external identifiers
- view keywords
- view addresses
- view biography
- view personal details
- view person
- view education
- view educations
- view education summary
- view employment
- view employments
- view employment summary
- view external identifier
- view funding
- view funding summary
- view keyword
- view other names
- view peer review
- view peer review summary
- view researcher url 
- view work
- view work summary

PR #3254, opened 2016-11-22
Revert "Return error for orcid profile wrong token"

bobcaprice:
Reverts ORCID/ORCID-Source#3243


bobcaprice:
[2.0 Should return an error when reading a record with a token for a different record](https://trello.com/c/FDunTlSp/3157-2-0-should-return-an-error-when-reading-a-record-with-a-token-for-a-different-record)

PR #3264, opened 2016-11-25
Get the password before it gets encrypted, so, we can then use it to …

amontenegro:
…authenticate the user after register


amontenegro:
[Returned to Sign In and not ORCID record when registering](https://trello.com/c/U4FmzhWX/3224-returned-to-sign-in-and-not-orcid-record-when-registering)

PR #3265, opened 2016-11-25
Update code so it follows the reenable workflow when needed

amontenegro:


amontenegro:
[When a user tries to register with an email address associated with a deactivated iD they are shown the old message: (email address) already exists in our system. Would you like to Sign in using (email address)](https://trello.com/c/GkbRcK5E/3226-when-a-user-tries-to-register-with-an-email-address-associated-with-a-deactivated-id-they-are-shown-the-old-message-email-addres)

PR #3266, opened 2016-11-25
Fix French translation

wetneb:
"12345 ORCID iDs and counting" cannot be translated in French as "12345 ORCID iDs et comptage". This is obviously a very bad literal translation. However, I cannot think of a satisfactory translation for this idiom. Wiktionary suggests "- et ce n'est pas fini" (https://fr.wiktionary.org/wiki/and_counting) but that's a bit long and not very clear in this context.
So I propose to remove the "et comptage" altogether: this does not express the increase, but it's much safer than the current version, which really hurts the credibility of the website.


alainna:
Thanks for your suggestion. I quite agree. 

However we will not be accepting this item as translations are now being handled in Transifex. 

We have contacted you directly by email regarding this item; please respond there. If you did not receive our message, please contact support@orcid.org. 

Many thanks, and warm regards, 
Alainna

PR #3274, opened 2016-11-29
Institutional sign in heuristics

wjrsimpson:


wjrsimpson:
[Institutional sign in heuristic check](https://trello.com/c/Ktvva7EC/3240-institutional-sign-in-heuristic-check)

PR #3275, opened 2016-11-29
Return 301 when trying to update deprecated account

amontenegro:


amontenegro:
[Auto deprecate- return an error when a member updates a deprecated record](https://trello.com/c/7bWBLWpD/3236-auto-deprecate-return-an-error-when-a-member-updates-a-deprecated-record)

PR #3276, opened 2016-11-29
Remove the deprecated account from the cache after depreciating it

amontenegro:


amontenegro:
[When a record is auto deprecated it still displays in the UI](https://trello.com/c/9NEBVZwp/3235-when-a-record-is-auto-deprecated-it-still-displays-in-the-ui)

PR #3283, opened 2016-11-30
When reactivating an account, set it as claimed as well

amontenegro:


amontenegro:
[Incorrect workflow - When a user tries to register with a deactivated unclaimed id](https://trello.com/c/jQliqnc9/3227-incorrect-workflow-when-a-user-tries-to-register-with-a-deactivated-unclaimed-id)

PR #3285, opened 2016-11-30
Fix for NPE on biography where biography is not public

bobcaprice:


bobcaprice:
[NPE exception when fetching a biography from a record without biography](https://trello.com/c/TYO8eX6s/3212-npe-exception-when-fetching-a-biography-from-a-record-without-biography)

PR #3287, opened 2016-11-30
Verify the email used to do the reactivation workflow

amontenegro:


amontenegro:
[Mark the email address as verified when the user completes the reactivate process.](https://trello.com/c/2XrDmq4G/3234-mark-the-email-address-as-verified-when-the-user-completes-the-reactivate-process)

PR #3289, opened 2016-11-30
Modify the code so it compare emails case insensitive

amontenegro:


amontenegro:
[Reactivate trigger should not be case sensitive](https://trello.com/c/z0J47FgH/3247-reactivate-trigger-should-not-be-case-sensitive)

PR #3296, opened 2016-12-02
Card3221 don t allow an empty citation

mjcalvo21:
https://trello.com/c/5IG5amzN/3221-don-t-allow-an-empty-citation-field-when-citation-format-is-unspecified


mjcalvo21:
[Don't allow an empty citation field when citation format is "unspecified"](https://trello.com/c/5IG5amzN/3221-don-t-allow-an-empty-citation-field-when-citation-format-is-unspecified)

PR #3299, opened 2016-12-05
Showing the subject on the digest for custom notifications

amontenegro:


amontenegro:
[The auto deprecate email notification sent to users just contains the users name as the body instead of the full notification](https://trello.com/c/yOkwM1NA/3252-the-auto-deprecate-email-notification-sent-to-users-just-contains-the-users-name-as-the-body-instead-of-the-full-notification)

PR #3303, opened 2016-12-06
Member list use public display name

lizkrznarich:
https://trello.com/c/wEfvtZOO/3250-member-list-use-public-display-name-instead-of-name


lizkrznarich:
[Activities should sort by date then title](https://trello.com/c/qMcYS1MT/3267-activities-should-sort-by-date-then-title)

lizkrznarich:
[Member list: Use Public Display Name instead of Name](https://trello.com/c/wEfvtZOO/3250-member-list-use-public-display-name-instead-of-name)

PR #3306, opened 2016-12-06
Card3141 update works labels on manual entry

mjcalvo21:
https://trello.com/c/720RoM7d/3141-update-works-labels-on-manual-entry-to-be-dependent-on-work-type


mjcalvo21:
[Update Works labels on manual entry to be dependent on Work type](https://trello.com/c/720RoM7d/3141-update-works-labels-on-manual-entry-to-be-dependent-on-work-type)

lizkrznarich:
@mjcalvo21 There's one title type missing from the both the properties and the mapping: Newspaper title. In https://github.com/ORCID/ORCID-Source/pull/3306/files#diff-d144651ab88d31fdf180486112bd823bR1031, newspaper-article title type should be Newspaper title instead of Newsletter title. Also, to make the code easier to read, I think the var names should match the values - ex: instead of labelTitleBook, use labelTitlePublisher and instead of labelTitleDissertation, use labelTitleInstitution. As we add new work types, it'll be a little easier to manage the mapping.

mjcalvo21:
@lizkrznarich Feedback has benn aplied. Tks!

PR #3308, opened 2016-12-06
Remove default citation value for works on UI

amontenegro:


amontenegro:
[When manually adding a work citation type needs to default to blank not "unspecified"](https://trello.com/c/XrmWCntC/3273-when-manually-adding-a-work-citation-type-needs-to-default-to-blank-not-unspecified)

PR #3310, opened 2016-12-06
Display 'part-of' label for fundings on public page

amontenegro:


amontenegro:
["part-of" does not display next to part-of funding identifiers](https://trello.com/c/Q5AuOrya/3230-part-of-does-not-display-next-to-part-of-funding-identifiers)

PR #3314, opened 2016-12-07
Some more usefull tests that verifies everything is public when it sh…

amontenegro:
…ould


amontenegro:
[Read-public tokens return an error on member API 2.0](https://trello.com/c/tWXd45Lz/3208-read-public-tokens-return-an-error-on-member-api-2-0)

PR #3315, opened 2016-12-07
Add reactivation to sign in

wjrsimpson:


wjrsimpson:
[User signing in with an email address that is associated with a deactivated record](https://trello.com/c/PZvIi2Nr/3225-user-signing-in-with-an-email-address-that-is-associated-with-a-deactivated-record)

PR #3320, opened 2016-12-08
made page-size default to 100 and page default to 1

TomDemeranville:
test https://localhost:8443/orcid-api-web/v2.0_rc3/group-id-record works with zero params, with name only and that if provided page size and page override default values.

https://trello.com/c/dJAYkWxZ/3265-don-t-require-page-and-page-size-on-peer-review-api-search


TomDemeranville:
[Don't require page and page size on peer-review API search](https://trello.com/c/dJAYkWxZ/3265-don-t-require-page-and-page-size-on-peer-review-api-search)

PR #3322, opened 2016-12-09
Changed so that we check deactivation before claimed status

wjrsimpson:
...so that we can use the correct workflow for sign in.


wjrsimpson:
[Incorrect workflow - When a user tries to register with a deactivated unclaimed id](https://trello.com/c/jQliqnc9/3227-incorrect-workflow-when-a-user-tries-to-register-with-a-deactivated-unclaimed-id)

PR #3323, opened 2016-12-09
[3283] Updating copy

mjcalvo21:
https://trello.com/c/uvF6IhaZ/3283-journal-title-field-label-needs-to-default-to-journal-title-when-no-category-is-selected


mjcalvo21:
[Journal-title field label needs to default to "Journal title" when no category is selected.](https://trello.com/c/uvF6IhaZ/3283-journal-title-field-label-needs-to-default-to-journal-title-when-no-category-is-selected)

PR #3330, opened 2016-12-12
Added primary and verified attributes to emails for rc4

bobcaprice:
Added to rc4


bobcaprice:
[2.0 does not include primary or verified status with email addresses, should it?](https://trello.com/c/QCdG0Txb/3177-2-0-does-not-include-primary-or-verified-status-with-email-addresses-should-it)

PR #3333, opened 2016-12-12
Api20 last modified consistency

amontenegro:


amontenegro:
[Return the last modified dates on private sections with read-public](https://trello.com/c/GnKfVChe/3152-return-the-last-modified-dates-on-private-sections-with-read-public)

PR #3337, opened 2016-12-14
Remove fonts dot com

lizkrznarich:


lizkrznarich:
[Remove fonts.com headers and css](https://trello.com/c/xf3iovZ2/3287-remove-fonts-com-headers-and-css)

PR #3341, opened 2016-12-15
Fixed case of peer review group type.

wjrsimpson:


wjrsimpson:
[Don't have the peer-review id type in all caps](https://trello.com/c/ygoJyCuv/3291-don-t-have-the-peer-review-id-type-in-all-caps)

PR #3345, opened 2016-12-16
Country visibility

bobcaprice:


bobcaprice:
[Removed unused visibility tags from JSON returns](https://trello.com/c/yv11WdHa/3155-removed-unused-visibility-tags-from-json-returns)

PR #3346, opened 2016-12-16
[2282] Updating function code for bulk delete

mjcalvo21:
https://trello.com/c/j6MxNY1u/2282-bulk-delete-doesn-t-work-in-safari


mjcalvo21:
[Bulk delete doesn't work in Safari](https://trello.com/c/j6MxNY1u/2282-bulk-delete-doesn-t-work-in-safari)

PR #3348, opened 2016-12-16
Card3215 sort item

mjcalvo21:
https://trello.com/c/p1xTU8LQ/3215-sort-item-identifies-alphabetically-and-list-part-of-identifiers-last


mjcalvo21:
[Sort item identifies alphabetically and list part-of identifiers last](https://trello.com/c/p1xTU8LQ/3215-sort-item-identifies-alphabetically-and-list-part-of-identifiers-last)

lizkrznarich:
This will work for the moment; may need to revisit if we add other relationship values in the future.

PR #3349, opened 2016-12-16
Refactor address test

amontenegro:


amontenegro:
Hi @wjrsimpson, @rcpeters,

This is an example of how the BB tests would be refactored. In this case, we merge the AddressTest of every release candidate in a single one, which make it faster (It goes from ~350 secs run all tests to ~75 secs to run just this single test), however, it make it a bit harder to maintains, which I think is ok given the performance boost we get.

Thanks

PR #3352, opened 2016-12-20
Truncated citations fix

bobcaprice:
Corrected loop in write method, shorter number of iterations was truncating some longer citations


bobcaprice:
[Truncated citations in XML](https://trello.com/c/DWWmmsM1/3307-truncated-citations-in-xml)

PR #3353, opened 2016-12-20
Include setupClientsandUsers as conditional step

jpeerz:
* Include setupClientsandUsers as conditional step
* Include firefox binary, correct path


amontenegro:
Nice!

PR #3369, opened 2016-12-22
removed default-value from Citation in swagger model serialisation

TomDemeranville:


TomDemeranville:
[Remove default citation style](https://trello.com/c/er7uS7nn/3334-remove-default-citation-style)

PR #3371, opened 2016-12-22
Corrected email schema and samples

bobcaprice:


bobcaprice:
[Email addresses returned with verified and primary attributes are not valid against the schema](https://trello.com/c/c8eOrjhz/3324-email-addresses-returned-with-verified-and-primary-attributes-are-not-valid-against-the-schema)

PR #3374, opened 2016-12-22
Rollback changes for card 3310

amontenegro:


amontenegro:
[Token endpoint error response is wrong](https://trello.com/c/4JBaZYOR/3338-token-endpoint-error-response-is-wrong)

PR #3384, opened 2016-12-29
Remove the annotations and format the code

amontenegro:


amontenegro:
[Remove unused security annotations from public API](https://trello.com/c/Wzt76xRI/3348-remove-unused-security-annotations-from-public-api)

PR #3387, opened 2016-12-30
Return empty list instead of null element

amontenegro:


amontenegro:
[Records without an email field are invalid, but records with only private emails don't return an email field](https://trello.com/c/voqIa45r/3323-records-without-an-email-field-are-invalid-but-records-with-only-private-emails-don-t-return-an-email-field)

PR #3392, opened 2017-01-03
Check that public client can't access the members API

amontenegro:


amontenegro:
[OrcidError for using a public API read-public token on the member API](https://trello.com/c/StyIpt6b/3318-orciderror-for-using-a-public-api-read-public-token-on-the-member-api)

PR #3397, opened 2017-01-04
Rewrote bibtex to remove external libraries and generate it ourselves…

TomDemeranville:
… - this now escapes properly

Added ISSN and ISBN types
Added check to ensure bad DOIs are skipped


TomDemeranville:
[BibTeX Export: Files generated from ORCID metadata only take DOI identifiers](https://trello.com/c/GaWNPsiT/3343-bibtex-export-files-generated-from-orcid-metadata-only-take-doi-identifiers)

TomDemeranville:
[Escape exported bibtex](https://trello.com/c/xWwLTGgd/3354-escape-exported-bibtex)

PR #3399, opened 2017-01-04
Fixed format of json when accept header is application/json.

wjrsimpson:


orcid-api-common/src/main/java/org/orcid/api/common/jaxb/OrcidJacksonJaxbJsonProvider.java
```diff
@@ -31,8 +31,8 @@
  * @author Will Simpson
  */
 @Provider
-@Consumes({ VND_ORCID_JSON, ORCID_JSON, "text/orcid+json" })
-@Produces({ VND_ORCID_JSON, ORCID_JSON, "text/orcid+json" })
+@Consumes({ MediaType.APPLICATION_JSON })
+@Produces({ MediaType.APPLICATION_JSON })
```
rcpeters: should it be more then 1?

wjrsimpson:
[reading a 1.2 record with an Accept header of application/json returns the biography value field as "$"](https://trello.com/c/07WGy57E/3360-reading-a-1-2-record-with-an-accept-header-of-application-json-returns-the-biography-value-field-as)

orcid-integration-test/src/test/java/org/orcid/integration/blackbox/api/v2/rc4/MemberV2ApiClientImpl.java
```diff
@@ -310,7 +311,12 @@ public ClientResponse viewBiography(String orcid, String accessToken) {
...
-    
+
+    public ClientResponse viewBiographyJson(String orcid, String accessToken) {
+        URI uri = UriBuilder.fromPath(BIOGRAPHY).build(orcid);
+        return orcidClientHelper.getClientResponseWithToken(uri, MediaType.APPLICATION_JSON, accessToken);
+    }
+
     public ClientResponse createKeyword(String orcid, Keyword keyword, String accessToken) {
         URI uri = UriBuilder.fromPath(KEYWORDS).build(orcid);
         return orcidClientHelper.postClientResponseWithToken(uri, VND_ORCID_XML, keyword, accessToken);
```
amontenegro: Hey @wjrsimpson, why you have "viewPersonJson" in the public api but not in the members API?

amontenegro:
Hey @wjrsimpson, why you modify only the RC4 endpoints? what happen with the other RC endpoints?

amontenegro:
@wjrsimpson one thing more, why you only add json functions for biography and Person? viewBiographyJson and viewPersonJson, the error doesn't happen to the other elements?

orcid-api-common/src/main/java/org/orcid/api/common/jaxb/OrcidJacksonJaxbJsonProvider.java
wjrsimpson: @rcpeters See also OrcidJacksonJaxbJsonProviderPretty. The other media types are pretty, whereas application/json is not. It looks like application/json was working by 'luck' previously, and that's why it changed when i removed the default Jersey providers.

wjrsimpson:
@amontenegro I only added tests for the rc4 and for some example calls. The actual changes are global, so affect all versions and all calls. Our black box test coverage for json is not good. I want to discuss that at our meeting tomorrow. It is a bigger job to add tests for all json calls.

PR #3404, opened 2017-01-05
Dont show the required scope in the error message

amontenegro:


amontenegro:
[Don't list old scopes with a 9038 error message](https://trello.com/c/oBk9jI0i/3362-don-t-list-old-scopes-with-a-9038-error-message)

PR #3405, opened 2017-01-05
Changed error response from /oauth/token to be oauth compliant

bobcaprice:


bobcaprice:
[OrcidError for locked client id in 2 steps process](https://trello.com/c/pvcNDYRz/3315-orciderror-for-locked-client-id-in-2-steps-process)

PR #3407, opened 2017-01-05
Card3309 sign in with verifyedit flag still shows old verify email text

mjcalvo21:
https://trello.com/c/Ajn2vWBa/3309-sign-in-with-verifyedit-flag-still-shows-old-verify-email-text


mjcalvo21:
[Sign-in with verifyEdit flag still shows old verify email text](https://trello.com/c/Ajn2vWBa/3309-sign-in-with-verifyedit-flag-still-shows-old-verify-email-text)

lizkrznarich:
@mjcalvo21 To avoid messing up existing translations, keys for existing properties need to stay the same - use new names the new properties. Ex: workspace.ensure_future_access_old needs to be workspace.ensure_future_access, and the new version should get a new key. Otherwise, the translations for those existing strings disappear until someone goes into Transifex and clicks a lot of buttons to match the new key name up with its existing translation.

PR #3411, opened 2017-01-06
Fixed work sorting in 2.0 API.

wjrsimpson:


wjrsimpson:
[Return works in the 2.0 API in the same order as on the website](https://trello.com/c/ERFBR1qB/3049-return-works-in-the-2-0-api-in-the-same-order-as-on-the-website)

PR #3413, opened 2017-01-06
Remove old orcid client group xsds

amontenegro:


amontenegro:
[Do we still need the orcid-client-group xsd's?](https://trello.com/c/GcSyBb3v/3183-do-we-still-need-the-orcid-client-group-xsd-s)

PR #3415, opened 2017-01-06
remove alert used for debugging and verifyEdit feature flag

lizkrznarich:


lizkrznarich:
[Verify to Edit is turned on without the feature flag](https://trello.com/c/UdKauJJf/3380-verify-to-edit-is-turned-on-without-the-feature-flag)

PR #3421, opened 2017-01-09
sf member list for parent org field, check that consortia opportunity…

lizkrznarich:
… stage is closed won


lizkrznarich:
[Parent Organisation still showing on member list after the information has been removed in SalesForce](https://trello.com/c/ZAuTF6fu/3383-parent-organisation-still-showing-on-member-list-after-the-information-has-been-removed-in-salesforce)

PR #3422, opened 2017-01-09
Contributors visibility

bobcaprice:


bobcaprice:
[Bring Funding Contributors in line with Work Contributors, where visibility is inherited from parent.](https://trello.com/c/97SXJINb/3356-bring-funding-contributors-in-line-with-work-contributors-where-visibility-is-inherited-from-parent)

orcid-web/src/main/java/org/orcid/frontend/web/controllers/WorksController.java
```diff
@@ -327,19 +327,6 @@ WorkForm getWorkInfo(@RequestParam(value = "workId") Long workId) {
                             ProfileEntity profileEntity = profileEntityCacheManager.retrieve(contributorOrcid);
                             String publicContributorCreditName = cacheManager.getPublicCreditName(profileEntity);
                             contributor.setCreditName(Text.valueOf(publicContributorCreditName));
-                            if(contributorOrcid.equals(getEffectiveUserOrcid())) {
```
rcpeters: @bobcaprice @wjrsimpson This is going to be an issue. Researcher changes name to private but the record will still reflect the researchers name for every work (on their own record). Please research and discuss before tomorrow meeting.

rcpeters: One solution possible solution....

1. Never return email address
2. If it has an valid ORCID id return the same name as you would source.
3. If there is no ORCID just return the name provided.

...just thinking

PR #3424, opened 2017-01-10
fix admin-actions controller names to match names used in ftl templates

lizkrznarich:


lizkrznarich:
[Broken display in admin actions](https://trello.com/c/x2ukronX/3389-broken-display-in-admin-actions)

PR #3429, opened 2017-01-11
change feature flag name for stateless public menu

lizkrznarich:


lizkrznarich:
[Change feature flag name for stateless public record menu layout](https://trello.com/c/CQvzqZ3r/3398-change-feature-flag-name-for-stateless-public-record-menu-layout)

PR #3434, opened 2017-01-11
New contributors credit name will not have visibility, so, we need to…

amontenegro:
… ignore that property for the old ones

### DIFF
--- a/orcid-model/src/main/java/org/orcid/jaxb/model/common_rc4/CreditName.java
+++ b/orcid-model/src/main/java/org/orcid/jaxb/model/common_rc4/CreditName.java
@@ -31,6 +31,8 @@ import javax.xml.bind.annotation.XmlRootElement;
 import javax.xml.bind.annotation.XmlType;
 import javax.xml.bind.annotation.XmlValue;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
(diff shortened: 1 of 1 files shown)


amontenegro:
[Existing works with contributors won't have the details section display and return an error when called via the API](https://trello.com/c/DayMQAcg/3400-existing-works-with-contributors-won-t-have-the-details-section-display-and-return-an-error-when-called-via-the-api)

orcid-model/src/main/java/org/orcid/jaxb/model/common_rc4/CreditName.java
```diff
@@ -53,6 +55,7 @@
 @XmlAccessorType(XmlAccessType.FIELD)
 @XmlType( propOrder = { "content" })
 @XmlRootElement(name = "credit-name")
+@JsonIgnoreProperties(ignoreUnknown = true)
```
rcpeters: Let's be explicit. https://fasterxml.github.io/jackson-annotations/javadoc/2.6/com/fasterxml/jackson/annotation/JsonIgnoreProperties.html

amontenegro: Done!

PR #3437, opened 2017-01-12
Check if the given element have contributors to avoid NPE

amontenegro:


amontenegro:
[Can't post funding with the 2.0 API](https://trello.com/c/1q2obpY6/3403-can-t-post-funding-with-the-2-0-api)

PR #3439, opened 2017-01-12
make publicRecordMenu flag work on locked and deprecated records

lizkrznarich:


lizkrznarich:
[publicRecordMenu: new menu doesn't display on locked and deprecated accounts](https://trello.com/c/YJ2KolBo/3402-publicrecordmenu-new-menu-doesn-t-display-on-locked-and-deprecated-accounts)

PR #3442, opened 2017-01-12
Handle undefined group id on peer reviews

amontenegro:


amontenegro:
[publicRecordMenu: Going to the http public record page with that flag logs you out.](https://trello.com/c/OlBtr86M/3401-publicrecordmenu-going-to-the-http-public-record-page-with-that-flag-logs-you-out)

PR #3445, opened 2017-01-12
Group ids with peer reviews should not be deleted

amontenegro:


amontenegro:
[Prevent deleting group ids that have one or more peer reviews associated to it](https://trello.com/c/dsTLGwrZ/3409-prevent-deleting-group-ids-that-have-one-or-more-peer-reviews-associated-to-it)

PR #3447, opened 2017-01-13
Updated test to match changes in https://trello.com/c/pvcNDYRz/3315-orciderror-for-locked-client-id-in-2-steps-process

wjrsimpson:


wjrsimpson:
[OrcidError for locked client id in 2 steps process](https://trello.com/c/pvcNDYRz/3315-orciderror-for-locked-client-id-in-2-steps-process)

PR #3448, opened 2017-01-13
remove membership start date = this year param from queries, since me…

lizkrznarich:
…mberships no longer all begin jan 1


lizkrznarich:
[Member list: Members with start/end dates in different calendar years not displayed correctly.](https://trello.com/c/SMw6BtSi/3412-member-list-members-with-start-end-dates-in-different-calendar-years-not-displayed-correctly)

PR #3452, opened 2017-01-13
Increase max number of default results for search

amontenegro:


amontenegro:

PR #3453, opened 2017-01-16
Fix2 0 client id in schema

wjrsimpson:


wjrsimpson:
[Determine how (if) we include links to clients in source info](https://trello.com/c/UtetAmWA/3153-determine-how-if-we-include-links-to-clients-in-source-info)

PR #3454, opened 2017-01-16
Rollback changes to the SOLR config, so, we dont affect the 1.2 API

amontenegro:


amontenegro:

PR #3459, opened 2017-01-17
Include path checking in some unit tests

amontenegro:


amontenegro:
[Reading the employments and educations section with the public api the section path is missing](https://trello.com/c/bghcAcVn/3421-reading-the-employments-and-educations-section-with-the-public-api-the-section-path-is-missing)

PR #3462, opened 2017-01-18
Get the number of results found from solr, not from the size of the list

amontenegro:


amontenegro:
[2.0 Search return total number of results](https://trello.com/c/3vPFfcu7/3367-2-0-search-return-total-number-of-results)

amontenegro:

PR #3464, opened 2017-01-18
Fix the validation error for 16 digits clients (old style clients) an…

amontenegro:
…d add some sample files


amontenegro:
[2.0 stable: 16 digit format client ids as source have a validation error.](https://trello.com/c/tU8jqj2f/3430-2-0-stable-16-digit-format-client-ids-as-source-have-a-validation-error)

PR #3466, opened 2017-01-18
Fix validation of element lists

amontenegro:


amontenegro:
[When reading a subsection of activities- the parent element returned is not valid.](https://trello.com/c/9fnngc56/3431-when-reading-a-subsection-of-activities-the-parent-element-returned-is-not-valid)

PR #3472, opened 2017-01-19
When bio is empty or not visible, dont return it

amontenegro:


amontenegro:
[personal-details- inconsistently providing path to empty sections.](https://trello.com/c/MBwkQIUp/3422-personal-details-inconsistently-providing-path-to-empty-sections)

PR #3476, opened 2017-01-23
Corrected black box test to expect 400 from /oauth/token endpoint due…

bobcaprice:
… to bad scope


bobcaprice:
[InternalAPITest black box test failures](https://trello.com/c/xWKmIBWg/3442-internalapitest-black-box-test-failures)

PR #3477, opened 2017-01-23
Store the legacy client ids in memory, for fast look up.

wjrsimpson:


wjrsimpson:
[Optimize 2.0 client id mapping](https://trello.com/c/pvWRRTci/3444-optimize-2-0-client-id-mapping)

PR #3478, opened 2017-01-23
Don't check country source if client ID is not passed to filter.

wjrsimpson:


wjrsimpson:
[500 error when posting external identifiers to some records](https://trello.com/c/u3znV4zj/3378-500-error-when-posting-external-identifiers-to-some-records)

PR #3480, opened 2017-01-24
Add read biography section to README.md

alainna:
Added sample cURL statement to read biography section of record


caoyler:
Good catch that this was missing. But it should be added in the Read record items list rather than the Read Sections as it's a single field and behaves like you're reading one item not a list.

alainna:
Will add along with other sugg 👍

PR #3485, opened 2017-01-26
Add option to round trip notification alert

wjrsimpson:


wjrsimpson:
[The yellow box that connects the accounts for the "roundtrip" is not working as it should](https://trello.com/c/Q0k3uKmJ/3424-the-yellow-box-that-connects-the-accounts-for-the-roundtrip-is-not-working-as-it-should)

PR #3487, opened 2017-01-26
fixed identifier API to work on the v2.0 endpoint only.

TomDemeranville:
Also added default xml content type so list is accessable from the browser
Example: https://localhost:8443/orcid-pub-web/v2.0/identifiers?locale=fr


TomDemeranville:
[Move identifier type definitions to the database.](https://trello.com/c/FNF4Kufu/3269-move-identifier-type-definitions-to-the-database)

PR #3488, opened 2017-01-26
Archive notification alerts for the IdP when a user connection is removed

wjrsimpson:


wjrsimpson:
[Multiple yellow boxed for the Inst. roundtrip on production](https://trello.com/c/N5RHvIRY/3425-multiple-yellow-boxed-for-the-inst-roundtrip-on-production)

PR #3493, opened 2017-01-27
make sure the colorbox closes before moving on

rcpeters:


orcid-integration-test/src/test/java/org/orcid/integration/blackbox/api/BlackBoxBase.java
```diff
@@ -841,6 +841,7 @@ public static void createEmployment(String institutionName) {
         
         BBBUtil.extremeWaitFor(BBBUtil.angularHasFinishedProcessing(), webDriver);
         BBBUtil.ngAwareClick(webDriver.findElement(By.xpath("//button[@id='save-affiliation']")), webDriver);
+        BBBUtil.noCboxOverlay(webDriver);
```
amontenegro: I think we should add the same here: https://github.com/ORCID/ORCID-Source/pull/3493/files#diff-957f5944b79eb67fb597edddb8c79bcaL787 in the function that creates Educations

PR #3495, opened 2017-01-27
added https to the DOI definition in the DB.

TomDemeranville:
updated js in UI


TomDemeranville:
[Change default URL for DOIs](https://trello.com/c/OfYenJ21/3137-change-default-url-for-dois)

PR #3502, opened 2017-01-30
fixed character encoding issue.

TomDemeranville:


TomDemeranville:
[New identifier names can't handle special characters](https://trello.com/c/a2CS6qq6/3458-new-identifier-names-can-t-handle-special-characters)

PR #3503, opened 2017-01-30
update registration form text

lizkrznarich:


lizkrznarich:
[Registration form: Update notifications text to match practice](https://trello.com/c/USvsvRcF/3455-registration-form-update-notifications-text-to-match-practice)

PR #3504, opened 2017-01-30
removed cv-id

TomDemeranville:


TomDemeranville:
[2 new identifiers added for FCT](https://trello.com/c/u5VYDZ67/3339-2-new-identifiers-added-for-fct)

PR #3505, opened 2017-01-30
Archive obsolote notification alerts

wjrsimpson:


wjrsimpson:
[Archive IdP round trip alerts when connection made](https://trello.com/c/7CqMhzv5/3451-archive-idp-round-trip-alerts-when-connection-made)

PR #3512, opened 2017-01-31
Use name if public display name is not present for the member.

wjrsimpson:


wjrsimpson:
[Member list: Names are blank for orgs with no public display name](https://trello.com/c/B7sMLpGi/3457-member-list-names-are-blank-for-orgs-with-no-public-display-name)

PR #3513, opened 2017-02-01
Added reason field for locked accounts

bobcaprice:


bobcaprice:
[Add a field for why a record is locked](https://trello.com/c/LHINYoqK/3465-add-a-field-for-why-a-record-is-locked)

PR #3518, opened 2017-02-01
Include the org disambiguated id for funding and affiliations

amontenegro:


amontenegro:
[Affiliations- checking remove after adding an org doesn't completely remove it.](https://trello.com/c/vvvs3A1N/3471-affiliations-checking-remove-after-adding-an-org-doesn-t-completely-remove-it)

PR #3521, opened 2017-02-02
User driven deprecation now shows all primary emails when prompting f…

bobcaprice:
…or confirmation of deprecation


bobcaprice:
[User Deprecation: additional emails on primary account are missing](https://trello.com/c/lEcXkoBI/3469-user-deprecation-additional-emails-on-primary-account-are-missing)

PR #3524, opened 2017-02-03
removed dx from doi .org when autogenerating urls in the ui

TomDemeranville:


TomDemeranville:
[Change default URL for DOIs](https://trello.com/c/OfYenJ21/3137-change-default-url-for-dois)

PR #3525, opened 2017-02-03
modified support URLs to use https in messages_en

TomDemeranville:


TomDemeranville:
[BibTex export: Please change text](https://trello.com/c/QCbxNMXP/3467-bibtex-export-please-change-text)

PR #3527, opened 2017-02-03
added case sensitivity to identifier types.  Updated API.

TomDemeranville:


TomDemeranville:
[Add a column to record if an external identifier is case sensitive](https://trello.com/c/FwPVzMyZ/3485-add-a-column-to-record-if-an-external-identifier-is-case-sensitive)

PR #3530, opened 2017-02-07
Widget name visibility

bobcaprice:


bobcaprice:
[Fix widgets to follow our name rules](https://trello.com/c/x5V8i2zL/3486-fix-widgets-to-follow-our-name-rules)

PR #3545, opened 2017-02-09
Unsupported scope account settings

bobcaprice:


bobcaprice:
[Records which have an active token for an unsupported scope show no permission under trusted organizations](https://trello.com/c/ORAaqa8R/3492-records-which-have-an-active-token-for-an-unsupported-scope-show-no-permission-under-trusted-organizations)

PR #3547, opened 2017-02-09
make sure long member names wrap

lizkrznarich:


lizkrznarich:
[Member list: long names don't wrap](https://trello.com/c/Bc7DL9NH/3505-member-list-long-names-don-t-wrap)

PR #3550, opened 2017-02-11
Common passwords now prevented when resetting password using forgot p…

bobcaprice:
…assword link or account settings


bobcaprice:
[Common password can be set on Account Settings or Reset password](https://trello.com/c/QlIaXSox/3507-common-password-can-be-set-on-account-settings-or-reset-password)

PR #3552, opened 2017-02-13
Corrected bulk namespace for xml docs

bobcaprice:


bobcaprice:
[bulk works returns message with namespace ns22](https://trello.com/c/7EfgdD3P/3514-bulk-works-returns-message-with-namespace-ns22)

PR #3562, opened 2017-02-15
Use the new OrcidType in the switchUserFilter and in the Notifications

amontenegro:


amontenegro:
[Using the admin function to switch to another record takes you to signin instead of the other record.](https://trello.com/c/Vf1u8grt/3535-using-the-admin-function-to-switch-to-another-record-takes-you-to-signin-instead-of-the-other-record)

PR #3567, opened 2017-02-16
Update properties script

lizkrznarich:


lizkrznarich:
[Jython script to synchronize test languages](https://trello.com/c/ofaKr5G0/3541-jython-script-to-synchronize-test-languages)

PR #3568, opened 2017-02-17
Ignore bulk schema validation and update xsd to better describe his content

amontenegro:


amontenegro:
@rcpeters please take a look.

Now the bulk XSD describes what a bulk have, then, in the validation phase we just ignore bulk schema validation and delegate it to manager.

rcpeters:
❤️

PR #3569, opened 2017-02-18
Fix research community

wjrsimpson:


wjrsimpson:
[Consortia self service: Additional fields](https://trello.com/c/ZgDbdqjW/3512-consortia-self-service-additional-fields)

PR #3570, opened 2017-02-20
Clear the info element when the colorbox close

amontenegro:


amontenegro:
[User Deprecation: Success message no longer displays the iDs](https://trello.com/c/2cYq23pv/3543-user-deprecation-success-message-no-longer-displays-the-ids)

PR #3576, opened 2017-02-22
Resend claim form now uses AJAX

bobcaprice:


bobcaprice:
[Make resend claim functionality use AJAX, bring in line with reset password request changes](https://trello.com/c/JNPwGaFm/3539-make-resend-claim-functionality-use-ajax-bring-in-line-with-reset-password-request-changes)

PR #3577, opened 2017-02-22
Modifying buildAdvancedSearchUrl so it returns an empty search query …

amontenegro:
…instead of null in case query is empty


amontenegro:
[Fix empty search url result https://orcid.org/orcid-search/quick-search?searchQuery=](https://trello.com/c/Q8ZtH9Rl/3553-fix-empty-search-url-result-https-orcid-org-orcid-search-quick-search-searchquery)

PR #3578, opened 2017-02-23
Sub member stuff

wjrsimpson:


wjrsimpson:
[Consortia self-service: UI to add and remove consortium members](https://trello.com/c/WbFfFiY6/3557-consortia-self-service-ui-to-add-and-remove-consortium-members)

PR #3581, opened 2017-02-23
Made page for resetting password ajax and added validation of passwor…

bobcaprice:
…d against common password list


bobcaprice:
[Common password can be set on Account Settings or Reset password](https://trello.com/c/QlIaXSox/3507-common-password-can-be-set-on-account-settings-or-reset-password)

bobcaprice:
[Make page for resetting password use AJAX](https://trello.com/c/WZTL1ndN/3529-make-page-for-resetting-password-use-ajax)

PR #3590, opened 2017-02-28
External id typeahead

TomDemeranville:
Modified UI to use a typeahead for external IDs instead of a dropdown:

* matching based on starts of words.  So 'hello id' and 'yes, I said hello id' both match for 'he', but 'the id' does not.
* validation same as live service
* URL creation is now possible using data from the database, but is not turned on due to questions about URLS as identifiers... (the logic in the UI does things like look for domain names)
* Added JS dependency to angular-ui.  Compiled custom version of angular-ui 2.5.0 with just the typeahead component
* 11 most popular identifier types appear on first click - derived from 2016 datadump.


TomDemeranville:
[Update external identifier UI to use a typeahead with "most popular" suggestions instead of a pick list.](https://trello.com/c/fwXASrqB/3558-update-external-identifier-ui-to-use-a-typeahead-with-most-popular-suggestions-instead-of-a-pick-list)

PR #3593, opened 2017-03-01
Made search case insensitive and put starts-with matches at top

TomDemeranville:
Added client side request caching


TomDemeranville:
[work id type ahead shouldn't be case sensitive](https://trello.com/c/S9qpJJnX/3568-work-id-type-ahead-shouldn-t-be-case-sensitive)

TomDemeranville:
[work id type ahead: sort by most popular or best match not alphabetical](https://trello.com/c/kDG91WCS/3570-work-id-type-ahead-sort-by-most-popular-or-best-match-not-alphabetical)

PR #3596, opened 2017-03-01
Reduce unnecessary updates of activities during 1.2 API PUTs.

wjrsimpson:


wjrsimpson:
[Don't update last modified data of activities when not changed](https://trello.com/c/CuHRHRG2/3574-don-t-update-last-modified-data-of-activities-when-not-changed)

PR #3598, opened 2017-03-02
added permissions for ro database on grant-id

TomDemeranville:


TomDemeranville:
[Peer-review group lookup only works on 2.0_rc3](https://trello.com/c/pfuSqvHv/3565-peer-review-group-lookup-only-works-on-2-0-rc3)

PR #3600, opened 2017-03-02
Get multiple works

bobcaprice:


bobcaprice:
[Implement GET /[ORCID iD]/works?put-codes=11111,2222,3333](https://trello.com/c/2fJDAHDo/3545-implement-get-orcid-id-works-put-codes-11111-2222-3333)

PR #3601, opened 2017-03-02
made default list contain all identifier types

TomDemeranville:


TomDemeranville:
[Make external ID typeahead default list contain all ids](https://trello.com/c/YSXtJL2d/3577-make-external-id-typeahead-default-list-contain-all-ids)

PR #3602, opened 2017-03-02
Fixed notification archiving when put code is 1 million or above.

wjrsimpson:


wjrsimpson:
[Can't archive notifications](https://trello.com/c/oCXPJfzg/3566-can-t-archive-notifications)

PR #3605, opened 2017-03-02
Verify edit change primary popup

lizkrznarich:


lizkrznarich:
[Verify to edit: Add popup when unverified email is set to primary](https://trello.com/c/5Y0Wv8O6/3534-verify-to-edit-add-popup-when-unverified-email-is-set-to-primary)

PR #3612, opened 2017-03-06
On groupable activities, display an empty <common:external-ids/> elem…

amontenegro:
…ent instead of none when the activity have no groupable identifiers


amontenegro:
[Return empty external ids instead of null element](https://trello.com/c/Pe7FBApU/3584-return-empty-external-ids-instead-of-null-element)

PR #3614, opened 2017-03-07
fixed work id validation

TomDemeranville:


TomDemeranville:
[Manually adding work without external id fails](https://trello.com/c/iOAvI3Oy/3583-manually-adding-work-without-external-id-fails)

PR #3615, opened 2017-03-07
update Czech translation

helix84:


lizkrznarich:
Hi there! Thanks so much for updating Czech translations! Rather than editing properties directly, however, we're now using [Transifex](https://www.transifex.com/) to manage our translation workflow, and we're pushing completed translations automatically to Github from there. Properties changed manually in Github aren't picked up by Transifex, so I'm going to close this PR and load the changes from this branch into Transifex. For future, we can set you up with access to Transifex - someone from our support team (suppport@orcid.org) will be in touch with you soon.

helix84:
@lizkrznarich I see, thanks. My handle at Transifex is helix84.

PR #3625, opened 2017-03-09
Reset password prefill email from oauth url

lizkrznarich:


lizkrznarich:
[Reset Password doesn't pull email addresses if they are included as part of the OAuth URL](https://trello.com/c/gyP9dxXi/3578-reset-password-doesn-t-pull-email-addresses-if-they-are-included-as-part-of-the-oauth-url)

PR #3627, opened 2017-03-10
Forced load of external id descriptions from server on display of work.

TomDemeranville:
Seems to be a known bug that typeahead input formatting does not work with promises.  So populating from an existing model fails.  ffs.


TomDemeranville:
[When editing a work with an existing identifier, the identifier type and url do not display.](https://trello.com/c/oBTMnCJH/3593-when-editing-a-work-with-an-existing-identifier-the-identifier-type-and-url-do-not-display)

PR #3631, opened 2017-03-13
Filter only the INSTITUTIONAL_CONNECTION if the client already knows …

amontenegro:
…the user


amontenegro:
[Issues with the inbox when following the link in notification emails](https://trello.com/c/D4fdncDi/3595-issues-with-the-inbox-when-following-the-link-in-notification-emails)

PR #3632, opened 2017-03-13
Card3232 bibtex import display works

mjcalvo21:
https://trello.com/c/aGFHQgPL/3232-bibtex-import-display-works-to-be-imported-as-they-will-be-saved-to-the-record


mjcalvo21:
[Bibtex import: display works to be imported as they will be saved to the record](https://trello.com/c/aGFHQgPL/3232-bibtex-import-display-works-to-be-imported-as-they-will-be-saved-to-the-record)

PR #3636, opened 2017-03-14
Reset password fix prefill var

lizkrznarich:


lizkrznarich:
[Console error with ?ResetPassword](https://trello.com/c/qiEg39ut/3597-console-error-with-resetpassword)

PR #3638, opened 2017-03-15
Dao refactor

TomDemeranville:
Removed references to DAO objects outside of core classes.  Pushed some existing DAO methods into managers, and created basic GivenPermissionToManager.

Still to do in another card is to refactor out entities from non-core classes, but this is less important.

Integration tests in same state pre and post refactor (three fails)


TomDemeranville:
[Update codebase so that all requests to DAOs come from managers only (required for message driven webhooks)](https://trello.com/c/LTbIlzhI/3556-update-codebase-so-that-all-requests-to-daos-come-from-managers-only-required-for-message-driven-webhooks)

PR #3639, opened 2017-03-16
Reset password ui improvements

lizkrznarich:


lizkrznarich:
[Inline password reset initiation improvements](https://trello.com/c/YmhL7phH/3598-inline-password-reset-initiation-improvements)

PR #3640, opened 2017-03-16
[3232] Bibtex - Removing unwanted html when there is no value for the…

mjcalvo21:
https://trello.com/c/aGFHQgPL/3232-bibtex-import-display-works-to-be-imported-as-they-will-be-saved-to-the-record


mjcalvo21:
[Bibtex import: display works to be imported as they will be saved to the record](https://trello.com/c/aGFHQgPL/3232-bibtex-import-display-works-to-be-imported-as-they-will-be-saved-to-the-record)

PR #3641, opened 2017-03-16
fix style on admin dropdown to make all options visible

lizkrznarich:


lizkrznarich:
[Add 'Europe' as a new region for the Search and Link Wizard filter](https://trello.com/c/Ks0JcN4D/3511-add-europe-as-a-new-region-for-the-search-and-link-wizard-filter)

PR #3642, opened 2017-03-16
add property and explicit true false for show_login

rcpeters:


rcpeters:
[Explicit setting of show_login and properties setting](https://trello.com/c/Ek8g0Qr2/3599-explicit-setting-of-show-login-and-properties-setting)

PR #3646, opened 2017-03-21
Languages missing Javascript properties now use English language (US …

bobcaprice:
…locale) counterparts to avoid blanks


bobcaprice:
[javascript_??.properties fall back to English](https://trello.com/c/lxrqnsGK/3542-javascript-properties-fall-back-to-english)

PR #3650, opened 2017-03-22
Separate analytics filters for member and public API, event labels pr…

bobcaprice:
…efixed with client type, API version prefixed with public or member


bobcaprice:
[API analytics Add a public/member API filter](https://trello.com/c/so5Lx799/3612-api-analytics-add-a-public-member-api-filter)

PR #3653, opened 2017-03-22
remove ResestPassword request param conditional

lizkrznarich:
remove ResestPassword request param conditional to unhide inline password reset initiation


lizkrznarich:
[inline password reset initiation /signin?ResetPassword](https://trello.com/c/xH5dHflL/3591-inline-password-reset-initiation-signin-resetpassword)

PR #3655, opened 2017-03-23
Reset password remove asterisk

lizkrznarich:


lizkrznarich:
[inline password reset initiation /signin?ResetPassword](https://trello.com/c/xH5dHflL/3591-inline-password-reset-initiation-signin-resetpassword)

PR #3657, opened 2017-03-23
Analytics data no longer sent if no tracking code; removed tracking c…

bobcaprice:
…ode from runtime properties


bobcaprice:
[Don't send API analytics data if no tracking code](https://trello.com/c/DIoQHdg5/3618-don-t-send-api-analytics-data-if-no-tracking-code)

PR #3658, opened 2017-03-23
Use read only db on scheduler

amontenegro:


amontenegro:
[Update scheduler configuration to read from the replication DB](https://trello.com/c/6CSxEU0z/3508-update-scheduler-configuration-to-read-from-the-replication-db)

PR #3662, opened 2017-03-27
Excluded unclaimed accounts

bobcaprice:


bobcaprice:
[Exclude unclaimed accounts from verified required 2017 announcement](https://trello.com/c/FTwRHZTT/3624-exclude-unclaimed-accounts-from-verified-required-2017-announcement)

PR #3664, opened 2017-03-28
Changed so that the user is always allowed to archive the notification.

wjrsimpson:


wjrsimpson:
[Archive link on institutional round trip doesn't archive message](https://trello.com/c/E50FoHgg/3619-archive-link-on-institutional-round-trip-doesn-t-archive-message)

PR #3670, opened 2017-03-31
Filter for token owner

amontenegro:


amontenegro:
[Filter check url ORCID iD against token user ORCID iD.](https://trello.com/c/zqbTQkUH/2800-filter-check-url-orcid-id-against-token-user-orcid-id)

PR #3675, opened 2017-04-07
Add token expired message

bobcaprice:


bobcaprice:
[No error message returned when clicking on expired password reset link](https://trello.com/c/biMkTAoO/3627-no-error-message-returned-when-clicking-on-expired-password-reset-link)

PR #3676, opened 2017-04-10
Added URL to analytics data and changed null API version to empty string

bobcaprice:


bobcaprice:
[API analytics: Public API user posting?](https://trello.com/c/wUAQZyyf/3613-api-analytics-public-api-user-posting)

PR #3682, opened 2017-04-13
Limit public api search offset and page size

bobcaprice:


bobcaprice:
[Disallow 'start' parameter for searching from exceeding 10,000 where request has no client credentials](https://trello.com/c/zmLBbtle/3646-disallow-start-parameter-for-searching-from-exceeding-10-000-where-request-has-no-client-credentials)

PR #3683, opened 2017-04-17
Print record new tab

lizkrznarich:


lizkrznarich:
[Print Record: Peer-review section oddness](https://trello.com/c/rjQpjXuZ/3325-print-record-peer-review-section-oddness)

PR #3687, opened 2017-04-20
update new footer text per most recent copyight statement edits

lizkrznarich:


lizkrznarich:
[Update footer layout](https://trello.com/c/D9V4sFZh/3464-update-footer-layout)

PR #3699, opened 2017-05-02
Add validation error message for unmarshall exceptions

amontenegro:


amontenegro:
[Better error message when adding an item with a field that exceeds the max character limit.](https://trello.com/c/4PREcoSO/3688-better-error-message-when-adding-an-item-with-a-field-that-exceeds-the-max-character-limit)

PR #3706, opened 2017-05-03
Remove the scope-path-type from the common.xsd since thats not info w…

amontenegro:
…e need there


amontenegro:
[Remove references to old scopes from the XSD](https://trello.com/c/12hrDTrL/3699-remove-references-to-old-scopes-from-the-xsd)

PR #3709, opened 2017-05-04
Fixed error message about no persistent id when there are no display attributes

wjrsimpson:


wjrsimpson:
[Institutional sign in error message wrong when IdP provides no attributes](https://trello.com/c/0cS6WCok/3701-institutional-sign-in-error-message-wrong-when-idp-provides-no-attributes)

PR #3711, opened 2017-05-04
Dedupe header check

wjrsimpson:


wjrsimpson:
[Institutional sign in check should not fail if the same value appears twice](https://trello.com/c/DMSJOET6/3702-institutional-sign-in-check-should-not-fail-if-the-same-value-appears-twice)

amontenegro:
Hey @wjrsimpson, merge monster!

PR #3712, opened 2017-05-05
anonymize ip addresses collected by google analytics trackers in registry UI

lizkrznarich:


lizkrznarich:
[Enable IP Anonymization for all GA scripts on all ORCID websites](https://trello.com/c/TUyzU1yV/3676-enable-ip-anonymization-for-all-ga-scripts-on-all-orcid-websites)

PR #3717, opened 2017-05-05
Adding header to the validation error message

amontenegro:


amontenegro:
[Add heading before the full error text in the error messages](https://trello.com/c/2NCJC9XL/3706-add-heading-before-the-full-error-text-in-the-error-messages)

PR #3723, opened 2017-05-09
Varchar to text

amontenegro:


amontenegro:
[Elements that use the common url type don't return validation errors for length](https://trello.com/c/P4bNfUut/3689-elements-that-use-the-common-url-type-don-t-return-validation-errors-for-length)

PR #3724, opened 2017-05-09
default oauth screen to signin 

lizkrznarich:
and remove support for #show_login fragment


lizkrznarich:
[Change OAuth screen default from signin to register](https://trello.com/c/driZpjfZ/3721-change-oauth-screen-default-from-signin-to-register)

PR #3726, opened 2017-05-10
Don't use transaction around digest email sending

wjrsimpson:


wjrsimpson:
[Send digest email without using transactions](https://trello.com/c/1WpmcQpY/3728-send-digest-email-without-using-transactions)

PR #3733, opened 2017-05-11
Add agreement signatory to the contact role list

amontenegro:


amontenegro:
[Consortia self-service: Add signatory role](https://trello.com/c/aevt30Wf/3656-consortia-self-service-add-signatory-role)

PR #3734, opened 2017-05-11
use ISO 8601 date on stats page

lizkrznarich:


lizkrznarich:
[Registry/ORCID.org: Use ISO 8601 dates](https://trello.com/c/rRK3Nx5N/3587-registry-orcid-org-use-iso-8601-dates)

PR #3735, opened 2017-05-11
member list only retrieve active integrations from salesforce

lizkrznarich:


lizkrznarich:
[Members List: Integration should not appear if flagged as inactive in SalesForce](https://trello.com/c/mBX9s1yV/3658-members-list-integration-should-not-appear-if-flagged-as-inactive-in-salesforce)

PR #3736, opened 2017-05-11
Member list check opp start date

lizkrznarich:


lizkrznarich:
[Member List: Members appearing on multiple times due to future  opportunities](https://trello.com/c/pNO1N7tC/3726-member-list-members-appearing-on-multiple-times-due-to-future-opportunities)

PR #3737, opened 2017-05-12
Contact role validation

wjrsimpson:


wjrsimpson:
[Consortia self-service: Allow only 1 main contact](https://trello.com/c/apHKU1OH/3654-consortia-self-service-allow-only-1-main-contact)

PR #3739, opened 2017-05-15
[3720] Updating markup to follow the same pattern for error display

mjcalvo21:
https://trello.com/c/l9K1EiOh/3720-error-messages-for-the-translated-title-appear-above-the-field-where-on-all-other-fields-the-error-is-below


mjcalvo21:
[Error messages for the translated title appear above the field, where on all other fields the error is below](https://trello.com/c/l9K1EiOh/3720-error-messages-for-the-translated-title-appear-above-the-field-where-on-all-other-fields-the-error-is-below)

PR #3741, opened 2017-05-16
Fix naming error that prevents from picking the org disambiguated id …

amontenegro:
…on fundings


amontenegro:
[Funding organization identifiers do not save when added manually](https://trello.com/c/hlaKKybn/3739-funding-organization-identifiers-do-not-save-when-added-manually)

PR #3745, opened 2017-05-17
Catch exceptions thrown from the works cache and send an alert so the…

amontenegro:
… dev team can know before other people does


amontenegro:
[OptionalDataException not captured on api_log4j.log](https://trello.com/c/yvORbDO3/3737-optionaldataexception-not-captured-on-api-log4j-log)

PR #3748, opened 2017-05-18
added rrid identifier

TomDemeranville:


TomDemeranville:
[Add RRID](https://trello.com/c/Q3sv3dLc/3708-add-rrid)

PR #3751, opened 2017-05-19
Fix contact role editing

wjrsimpson:


wjrsimpson:
[Consortia self-service: Add checkbox for voting](https://trello.com/c/9yjKHj16/3655-consortia-self-service-add-checkbox-for-voting)

wjrsimpson:
[Consortia self-service: Allow only 1 main contact](https://trello.com/c/apHKU1OH/3654-consortia-self-service-allow-only-1-main-contact)

PR #3752, opened 2017-05-19
Print view tweaks

lizkrznarich:


lizkrznarich:
[Print View: ?print](https://trello.com/c/T20kuy50/3711-print-view-print)

PR #3756, opened 2017-05-23
Clear incoming source

amontenegro:


amontenegro:
[Clear source from incoming elements](https://trello.com/c/MXsnfKfz/3757-clear-source-from-incoming-elements)

amontenegro:
@wjrsimpson could you please take a look a this?

Thanks

orcid-core/src/main/java/org/orcid/core/adapter/impl/MapperFacadeFactory.java
```diff
@@ -261,23 +261,6 @@ public void registerSourceConverters(MapperFactory mapperFactory, ClassMapBuilde
     private class SourceMapper<T, U> extends CustomMapper<SourceAware, SourceAwareEntity<?>> {
 
         @Override
```
amontenegro: Hey @wjrsimpson,

Please take a look at this, I'm thinking that we should map the source from DB to JAXB only, not the other way, and, in the managers, before storing the DB element we manually set the source.

Thanks

wjrsimpson: Yes, sounds good to me.

PR #3762, opened 2017-05-25
Update examples so all sources are HTTPS, fix orcid id to display https

amontenegro:


amontenegro:
[2.1 identifier returned at the start of a record when reading an entire record uses http](https://trello.com/c/wWutg6CS/3769-2-1-identifier-returned-at-the-start-of-a-record-when-reading-an-entire-record-uses-http)

amontenegro:
[2.1 client sources use http](https://trello.com/c/q1ja0IE6/3770-2-1-client-sources-use-http)

PR #3763, opened 2017-05-26
Fix some lgtm alerts

jhelie:
Hi!
Just wanted to quickly fix these alerts flagged up on lgtm.com - I hope it helps.

Some the enums comparison appeared to be wrong and using the '==' and '!=' operators on boxed integers only tests for reference equality, rather than for value equality.

For instance the below triggers if `index` and other `otherIndex` are both null (since `null == null`) or have the same reference but it does not test for equality of the values:
`if (index == otherIndex) return o2.compareTo(o1); `

A total of 252 alerts have been flagged up by lgtm.com:
https://lgtm.com/projects/g/ORCID/ORCID-Source/alerts/

You can enable pull request integration for fully automated PR reviews that will flag these in the future so they don't get past code review :)

### DIFF
--- a/orcid-core/src/main/java/org/orcid/core/oauth/OrcidOauth2UserAuthentication.java
+++ b/orcid-core/src/main/java/org/orcid/core/oauth/OrcidOauth2UserAuthentication.java
@@ -169,7 +169,7 @@ public class OrcidOauth2UserAuthentication implements Authentication {
             return StringUtils.EMPTY;
         }
         
-        if(Visibility.PUBLIC.value().equals(recordName.getVisibility())) {
+        if(recordName.getVisibility() == Visibility.PUBLIC) {
             if(!PojoUtil.isEmpty(recordName.getCreditName())) {
                 return recordName.getCreditName();
             } else {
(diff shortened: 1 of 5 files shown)


orcid-core/src/main/java/org/orcid/core/oauth/OrcidOauth2UserAuthentication.java
```diff
@@ -169,7 +169,7 @@ public String getName() {
             return StringUtils.EMPTY;
         }
         
-        if(Visibility.PUBLIC.value().equals(recordName.getVisibility())) {
```
rcpeters: This can't be correct as the change means they will never be equal?

jhelie: apologies if I got it wrong but if `.getVisibility()` returns a `Visibility` enum why couldn't it be `Visibility.PUBLIC`? 
A comparison on the enum value would indeed require `recordName.getVisibility().value().equals(Visibility.PUBLIC.values())` but that it isn't necessary to assess that the enum is `PUBLIC`?

jhelie: I'm not sure why `Visibility` has a `value()` method which is the same than its `toString()` one so maybe I'm missing something here

PR #3766, opened 2017-05-29
Set source on group record after conversion to entity (to fix null source).

wjrsimpson:


wjrsimpson:
[Source is null on group id record](https://trello.com/c/FyWVUabx/3775-source-is-null-on-group-id-record)

PR #3778, opened 2017-06-05
Adding a custom request matcher for CSRF validation, so, we can bypas…

amontenegro:
…s it on the /userStatus.json endpoint


rcpeters:
why are we doing this?

amontenegro:
The org.springframework.web.HttpRequestMethodNotSupportedException exception we want to fix is due to invalid CSRF tokens that are sent to the server after the session expired.

So, currently we check the user status with a POST to the /userStatus.json endpoint.

This filter will allow us to bypass CSRF validation for calls to /userStatus.json

PR #3780, opened 2017-06-05
Stop calling user status after error happens

amontenegro:


amontenegro:
[Stop calling userStatus.json when the CSRF token become invalid](https://trello.com/c/KkRvAFWO/3789-stop-calling-userstatusjson-when-the-csrf-token-become-invalid)

PR #3781, opened 2017-06-06
modified so that jwks is an array of keys

TomDemeranville:


TomDemeranville:
[Update jwks representation to be an array of keys instead of a single key](https://trello.com/c/doxqruMp/3788-update-jwks-representation-to-be-an-array-of-keys-instead-of-a-single-key)

PR #3784, opened 2017-06-06
Oauth2 screens on register redirect to authorize page

amontenegro:


amontenegro:
[OAuth 2 screens when registering permission is automatically granted](https://trello.com/c/vdZVbO5g/3793-oauth-2-screens-when-registering-permission-is-automatically-granted)

PR #3786, opened 2017-06-07
updated swagger test to use https

TomDemeranville:


TomDemeranville:
[Fix swagger test to use https](https://trello.com/c/SHOUZrUQ/3804-fix-swagger-test-to-use-https)

wjrsimpson:
It looks like this wasn't the problem after all. However, https is good so thanks for the update @TomDemeranville!

PR #3789, opened 2017-06-07
[3790] Fixing error for page not refreshing when session expired, fix…

mjcalvo21:
3790-in-some-cases-the-oauth-authentication-page-doesnt-reload-after-the-session-expire-or-the-csrf-token-become-invalid

### DIFF
--- a/orcid-web/src/main/webapp/static/javascript/script.js
+++ b/orcid-web/src/main/webapp/static/javascript/script.js
@@ -462,7 +462,7 @@ function addShibbolethGa(oauthGaString){
         }
     );
 }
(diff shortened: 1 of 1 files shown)


mjcalvo21:
[In some cases, the oauth authentication page doesn't reload after the session expire or the CSRF token become invalid](https://trello.com/c/cIGAbxb3/3790-in-some-cases-the-oauth-authentication-page-doesnt-reload-after-the-session-expire-or-the-csrf-token-become-invalid)

amontenegro:
Code looks ok, the functionality works as we expect, however, there are a pending item that we should discuss.

PR #3791, opened 2017-06-07
Log the error generated on BBBUtil.angularHasFinishedProcessing

amontenegro:


wjrsimpson:
@amontenegro Thanks, but I was already looking at that in https://github.com/ORCID/ORCID-Source/pull/3787.

PR #3792, opened 2017-06-08
Fixed config for token target filter.

wjrsimpson:


amontenegro:
[Token filter not configured](https://trello.com/c/p3jEHFM0/3806-token-filter-not-configured)

PR #3795, opened 2017-06-08
Openid discovery

TomDemeranville:
This works in manual testing, but I cannot get the integration tests to work at the moment.


TomDemeranville:
[Investigate openid discovery and see if we can support it with minimal effort. Use this to configure third party test tool.](https://trello.com/c/h6iStwhZ/3791-investigate-openid-discovery-and-see-if-we-can-support-it-with-minimal-effort-use-this-to-configure-third-party-test-tool)

wjrsimpson:
@TomDemeranville There are merge conflicts in your pull request. I am guessing you didn't want to commit changes to localhost.webpack.config.js?

TomDemeranville:
@wjrsimpson there is a problem with that file being in git as it contains local config information (if you want it to deploy to tomcat).  I will update so that it modifies the local source tree and re-commit.  Not sure what else to do.

wjrsimpson:
@TomDemeranville Sorry Tom, more conflicts!

PR #3798, opened 2017-06-09
Profiles created via API now have hashedOrcid column populated

bobcaprice:


bobcaprice:

TomDemeranville:
[Fix citeproc so that is can handle weird unicode characters.](https://trello.com/c/eDlVLvjd/3745-fix-citeproc-so-that-is-can-handle-weird-unicode-characters)

PR #3800, opened 2017-06-09
Add ng web driver

wjrsimpson:


orcid-integration-test/src/test/java/org/orcid/integration/blackbox/api/BBBUtil.java
```diff
@@ -243,8 +245,12 @@ public Boolean apply(WebDriver driver) {
         return new ExpectedCondition<Boolean>() {
             @Override
             public Boolean apply(WebDriver driver) {
-                ((JavascriptExecutor) driver).executeScript(angularWaitScript);
-                return Boolean.valueOf(((JavascriptExecutor) driver).executeScript("" + "return window._selenium_angular_done;").toString());
+                ((JavascriptExecutor) driver).executeScript(jQueryWaitScript);
```
rcpeters: I was thinking about race conditions.
-- Ng is running
-- Jquery check runs
-- Ng fires off jquery http, post jquery check
-- Ng finishes 
-- Ng check runs

PR #3805, opened 2017-06-13
throw a 400 invalid_grant errot for invalid auth code

TomDemeranville:


TomDemeranville:
[OAuth token endpoint returns 500 error if code invalid. It should return a more useful code.](https://trello.com/c/l9jOJPjb/3799-oauth-token-endpoint-returns-500-error-if-code-invalid-it-should-return-a-more-useful-code)

PR #3806, opened 2017-06-13
citeproc now replaces unicode with normal chars & strips control chars

TomDemeranville:


TomDemeranville:
[Fix citeproc so that is can handle weird unicode characters.](https://trello.com/c/eDlVLvjd/3745-fix-citeproc-so-that-is-can-handle-weird-unicode-characters)

PR #3807, opened 2017-06-13
Fixed max_age in openID to use seconds.

TomDemeranville:


TomDemeranville:
[OpenID max_age should be in seconds not milliseconds](https://trello.com/c/UeIVnlui/3802-openid-maxage-should-be-in-seconds-not-milliseconds)

PR #3810, opened 2017-06-13
Fix bb test access token security checks test

amontenegro:


amontenegro:
[Token filter not configured](https://trello.com/c/p3jEHFM0/3806-token-filter-not-configured)

PR #3811, opened 2017-06-14
don't show protocol in widget

lizkrznarich:


lizkrznarich:
[Widget: remove protocol from widget display](https://trello.com/c/ozQzbhib/3826-widget-remove-protocol-from-widget-display)

PR #3813, opened 2017-06-14
Modify the OrcidSwitchUserFilter to it works with the new 2FA changes

amontenegro:


amontenegro:
[Switching to another account returns a server error on a page that keeps reloading](https://trello.com/c/XtFRpyjE/3827-switching-to-another-account-returns-a-server-error-on-a-page-that-keeps-reloading)

bobcaprice:
@wjrsimpson fyi

PR #3814, opened 2017-06-15
Update existing accounts with scopes

bobcaprice:


bobcaprice:
[SetupClientsAndUsers does not add missing scopes](https://trello.com/c/cxQ7sqhW/3807-setupclientsandusers-does-not-add-missing-scopes)

PR #3815, opened 2017-06-15
Revoke token if code reused - mark as disabled

TomDemeranville:
example response (400):
{"error":"invalid_grant","error_description":"Reused authorization code: Diko4F"}


TomDemeranville:
[Trying to use authorization code twice should result in revoking previously issued access tokens](https://trello.com/c/8Rzc8HGO/3800-trying-to-use-authorization-code-twice-should-result-in-revoking-previously-issued-access-tokens)

PR #3817, opened 2017-06-16
Renamed dao method for updating hashed_orcids so no longer picked up …

bobcaprice:
…by aop last_modified stuff


bobcaprice:

PR #3820, opened 2017-06-20
moved togglz from web to core

TomDemeranville:


TomDemeranville:
[Move togglz from orcid-web to orcid-core](https://trello.com/c/wytVwZNk/3833-move-togglz-from-orcid-web-to-orcid-core)

PR #3823, opened 2017-06-21
Added toggles to black box tests

TomDemeranville:


TomDemeranville:
[Trying to use authorization code twice should result in revoking previously issued access tokens](https://trello.com/c/8Rzc8HGO/3800-trying-to-use-authorization-code-twice-should-result-in-revoking-previously-issued-access-tokens)

PR #3824, opened 2017-06-21
Catch all exceptions thrown from the Cache and alert slack

amontenegro:


amontenegro:
[Catch OptionalDataException on WorkEntityCacheManager](https://trello.com/c/rnBOrHmL/3839-catch-optionaldataexception-on-workentitycachemanager)

PR #3825, opened 2017-06-21
[3741] Upgrading pending external files to be used with orcid cdn. Up…

mjcalvo21:
…grading current paths to use orcid staticCdn instead of staticLocal

https://trello.com/c/OXsJoQAh/3741-self-host-noto-font-on-our-cdn


mjcalvo21:
[Self host Noto font on our CDN.](https://trello.com/c/OXsJoQAh/3741-self-host-noto-font-on-our-cdn)

PR #3830, opened 2017-06-26
liquibase script for creating index on notification.sent_date

bobcaprice:


bobcaprice:
[Create liquibase script to add index to notification.sent_date](https://trello.com/c/xJvSjZXC/3851-create-liquibase-script-to-add-index-to-notificationsentdate)

PR #3831, opened 2017-06-26
Members should not be able to post to deactivated records

amontenegro:


amontenegro:
[Members can post to deactivated accounts.](https://trello.com/c/2v4ENKeg/3835-members-can-post-to-deactivated-accounts)

PR #3836, opened 2017-06-28
Return 404 error for invalid uris in 1.2 and 2.0 APIs

amontenegro:


amontenegro:
[1.2 return better error for invalid endpoints](https://trello.com/c/7Ri1c65R/3822-12-return-better-error-for-invalid-endpoints)

orcid-api-common/src/main/java/org/orcid/api/common/jaxb/OrcidExceptionMapper.java
```diff
@@ -247,7 +247,9 @@ private OrcidMessage getLegacyOrcidEntity(String prefix, Throwable e) {
     }
 
     private Response newStyleErrorResponse(Throwable t, String version) {
-        if (WebApplicationException.class.isAssignableFrom(t.getClass())) {
+        if(NotFoundException.class.isAssignableFrom(t.getClass())) {
+            return getOrcidErrorResponse(t, version);
+        } else if (WebApplicationException.class.isAssignableFrom(t.getClass())) {
```
amontenegro: @wjrsimpson, any reason why the NotFoundException (which is a 404) should be returned along with all other WebApplicationException (400, 405, 415)? 

This was causing that the developer message for this error had nothing to do with the real error: 

apiError.9001.developerMessage=400 Bad Request: There is an issue with your data or the API endpoint. 405 Method Not Allowed: Endpoint and method mismatch. 415 Unsupported Media Type: data must be in XML or JSON format.

So, after this change we will return this developer message:

apiError.9016.developerMessage=404 Not Found: The resource was not found.

PR #3843, opened 2017-07-03
Changed 'Public API User' client name to 'Unknown', which is more acc…

bobcaprice:
…urate


bobcaprice:
[Public API user showing as client using member API](https://trello.com/c/W880YcAt/3861-public-api-user-showing-as-client-using-member-api)

PR #3844, opened 2017-07-03
Password expiry ui and email updates

amontenegro:


amontenegro:
[Password expiry: Make duration of password reset link explicit in UI and email](https://trello.com/c/JEtCqs0q/3856-password-expiry-make-duration-of-password-reset-link-explicit-in-ui-and-email)

PR #3845, opened 2017-07-05
Api analytics process now uses accept header and takes account of Acc…

bobcaprice:
…eptFilter, where content type will be used if accept header isn't


bobcaprice:
[API analytics bug: null being sent for content type](https://trello.com/c/deS74RV5/3860-api-analytics-bug-null-being-sent-for-content-type)

PR #3846, opened 2017-07-05
User orcid and/or email to find existing contacts before creating a n…

amontenegro:
…ew one


orcid-core/src/main/java/org/orcid/core/salesforce/adapter/SalesForceMapperFacadeFactory.java
```diff
@@ -136,9 +136,9 @@ private void registerContactMap(MapperFactory mapperFactory) {
         classMap.field("id", "Id");
         classMap.field("orcid", "ORCID_iD_Path__c");
         classMap.field("role.votingContact", "Voting_Contact__c");
+        classMap.field("email", "Email");
         classMap.fieldAToB("firstName", "FirstName");
-        classMap.fieldAToB("lastName", "LastName");
-        classMap.fieldAToB("email", "Email");
```
amontenegro: @wjrsimpson, was there any reason why mapping the email only from Contact to JSONObject and not the other way? With that, when the contact comes from SalesForce, we ignore mapping the email, so, we don't have a way to identify the contact if there are no ORCID id.

amontenegro:
[Consortium self service: Duplicated contact in SalesForce](https://trello.com/c/DQVqwvwS/3836-consortium-self-service-duplicated-contact-in-salesforce)

amontenegro:
Fixed

PR #3849, opened 2017-07-06
[3855] Updating font files to test fix on IE11

mjcalvo21:
https://trello.com/c/23E3t2NW/3855-noto-font-doesnt-appear-in-ie


mjcalvo21:
[Noto font doesn't appear in IE](https://trello.com/c/23E3t2NW/3855-noto-font-doesnt-appear-in-ie)

PR #3858, opened 2017-07-11
now check for private name and return nulls

TomDemeranville:
e.g. {"sub":"0000-0002-2601-8132","name":null,"family_name":null,"given_name":null}


TomDemeranville:
[OpenID: userinfo endpoint returns an error if the user's name is not public.](https://trello.com/c/aRiGh1k4/3885-openid-userinfo-endpoint-returns-an-error-if-the-users-name-is-not-public)

PR #3864, opened 2017-07-11
Deactivated accounts should throw exception on updates only

amontenegro:


amontenegro:
[Deactivated accounts: return an empty record with the <history:deactivation-date> element](https://trello.com/c/z24kFq3D/3895-deactivated-accounts-on-the-api-do-we-return-an-error-or-an-empty-record-with-the-historydeactivation-date-element)

PR #3866, opened 2017-07-12
Fix how redirect uris are added for public clients

amontenegro:


amontenegro:
[Can't type in redirect field when adding a redirect uri to an existing public client.](https://trello.com/c/yZpd3W7g/3898-cant-type-in-redirect-field-when-adding-a-redirect-uri-to-an-existing-public-client)

PR #3867, opened 2017-07-12
An invalid ORCID iD should return a 404

rcpeters:


rcpeters:
[Invalid ORCID iD should return 404](https://trello.com/c/UXTK4QUt/3900-invalid-orcid-id-should-return-404)

PR #3868, opened 2017-07-12
Modify client so it doesnt map config elements

amontenegro:


amontenegro:
[Edit clients: Long lived token option turns itself off when you edit redirect](https://trello.com/c/xeOtLEIh/3896-edit-clients-long-lived-token-option-turns-itself-off-when-you-edit-redirect)

PR #3870, opened 2017-07-12
Update the last modified on the client after removing old secret keys

amontenegro:


amontenegro:
[Revoked client secrets still work](https://trello.com/c/kEfmTBFV/3899-revoked-client-secrets-still-work)

PR #3875, opened 2017-07-13
fix incorrect var in family name query

lizkrznarich:


lizkrznarich:
[UI Affiliation search for Fundref ids results in endless spinny arrows](https://trello.com/c/2ZaizfJA/3784-ui-affiliation-search-for-fundref-ids-results-in-endless-spinny-arrows)

PR #3879, opened 2017-07-18
Fixed SF website handling.

wjrsimpson:


wjrsimpson:
[NPE error when trying to add new consortium from Salesforce](https://trello.com/c/BFiOauNH/3914-npe-error-when-trying-to-add-new-consortium-from-salesforce)

PR #3881, opened 2017-07-18
Ga event tracking to features

amontenegro:


amontenegro:
Hi @lizkrznarich, could you please take a look?

I moved the OrcidGA code to its own js file, and, import it in the html-head.ftl file so it is accessible from the scripts generated by the freemarker templates 

Thanks

amontenegro:
[Add GA event tracking to features](https://trello.com/c/t3K97XFc/3912-add-ga-event-tracking-to-features)

rcpeters:
@amontenegro why did you have to move this to it's own file?

amontenegro:
@rcpeters because the macro to show/hide features checkFeatureStatus could be used anywhere in the code, even before the script.js is loaded, so, in order to be able to use the OrcidGA function, we need to load that function as soon as possible in the code, hence, I move it to its own script file and load that file in the header, so, it is ready for any call to the checkFeatureStatus macro

rcpeters:
@amontenegro @lizkrznarich we don't need yet another javascript browser request. You can put in a freemarker file and do a freemarker include.

amontenegro:
great idea, I just moved the script to a freemarker file @rcpeters

lizkrznarich:
@amontenegro @rcpeters I can buy that.

PR #3882, opened 2017-07-19
fixed two leg auth for the member swagger API - affects group ID and …

TomDemeranville:
…search

note, known issue - you can't request more than one two-leg scope at once.
Test by using search and group-id read on member swagger interface.


TomDemeranville:
[Swagger not using /read-public token](https://trello.com/c/oB5EioHu/3915-swagger-not-using-read-public-token)

PR #3884, opened 2017-07-19
migrate other ehcaches

rcpeters:


orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidProfileCacheManagerImpl.java
```diff
@@ -155,11 +171,10 @@ public void put(OrcidProfile orcidProfile) {
 
     public void put(String orcid, OrcidProfile orcidProfile) {
         try {
-            synchronized (profileLockers.obtainLock(orcid)) {
-                profileCache.put(new Element(new OrcidCacheKey(orcid, releaseName), orcidProfile));
-            }
+            profileCache.acquireWriteLockOnKey(orcid);
```
amontenegro: Hey @rcpeters, i think here you should create the key _new OrcidCacheKey(orcid, releaseName)_ and acquire the lock on that key, not in the orcid id right?

orcid-core/src/main/java/org/orcid/core/manager/impl/ProfileEntityCacheManagerImpl.java
```diff
@@ -81,11 +82,10 @@ public void put(ProfileEntity profileEntity) {
     }
     public void put(String orcid, ProfileEntity profile) {
         try {
-            synchronized (lockers.obtainLock(orcid)) {
-                profileCache.put(new Element(new OrcidCacheKey(orcid, releaseName), profile));
-            }
+            profileCache.acquireWriteLockOnKey(orcid);
```
amontenegro: @rcpeters same here, create the key and acquire the lock on that key

orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidProfileCacheManagerImpl.java
rcpeters: hum.....if we are replacing `profileLockers.obtainLock(orcid)) ` then no. If we are saying the original code is incorrect then yes.

orcid-core/src/main/java/org/orcid/core/manager/impl/ProfileEntityCacheManagerImpl.java
rcpeters: hum.....if we are replacing `lockers.obtainLock(orcid)) ` then no. If we are saying the original code is incorrect then yes.

PR #3886, opened 2017-07-19
Change param boolean to Boolean so orika mapper works

amontenegro:


amontenegro:
[2.1 unverified email addresses returns null instead of false](https://trello.com/c/mIXLdriO/3918-21-unverified-email-addresses-returns-null-instead-of-false)

PR #3887, opened 2017-07-19
Fix details view on the member developer tools

amontenegro:


amontenegro:
[Details on the members developer tools is broken](https://trello.com/c/mG5n464V/3923-details-on-the-members-developer-tools-is-broken)

PR #3893, opened 2017-07-20
fix acquire lock

rcpeters:


orcid-core/src/main/java/org/orcid/core/manager/impl/ClientDetailsEntityCacheManagerImpl.java
```diff
@@ -76,23 +79,23 @@ public ClientDetailsEntity retrieveByIdP(String idp) throws IllegalArgumentExcep
...
                 if (needsFresh(dbDate, clientDetails)) {
                     clientDetails = clientDetailsManager.findByIdP(idp);
                     if (clientDetails == null)
                         throw new IllegalArgumentException("Invalid idp " + idp);
-                    clientDetailsCache.put(new Element(key, clientDetails));
+                    clientDetailsIdPCache.put(new Element(key, clientDetails));
                 }
             } finally {
-                clientDetailsCache.acquireWriteLockOnKey(key);
+                clientDetailsIdPCache.acquireWriteLockOnKey(key);
```
wjrsimpson: @rcpeters Should be releasing the lock there?

rcpeters: yes!

PR #3899, opened 2017-07-24
specify height and width for oauth screen orcid logo

lizkrznarich:


lizkrznarich:
[2Screen OAuth: logo is very large on IE10](https://trello.com/c/xHXww2ss/3926-2screen-oauth-logo-is-very-large-on-ie10)

PR #3901, opened 2017-07-26
add id count back to signin page

lizkrznarich:


lizkrznarich:
[Count of records missing from sign in page](https://trello.com/c/v6X4TMrz/3945-count-of-records-missing-from-sign-in-page)

PR #3902, opened 2017-07-26
Transform the object into a byte array instead

amontenegro:


amontenegro:
[OutOfMemoryError in message listener](https://trello.com/c/8FF2EldZ/3946-outofmemoryerror-in-message-listener)

PR #3918, opened 2017-08-01
Ng2 migration.v.0.4

mjcalvo21:


lizkrznarich:
[There is no response when clicking Display my iD on other sites](https://trello.com/c/yt0jYno1/3957-there-is-no-response-when-clicking-display-my-id-on-other-sites)

PR #3921, opened 2017-08-03
Fix test redirect uris on dev tools and member dev tools

amontenegro:


amontenegro:
[Buttons to add common redirect uris don't work](https://trello.com/c/n2DaQnO1/3968-buttons-to-add-common-redirect-uris-dont-work)

PR #3922, opened 2017-08-04
Remove dates from affiliations

bobcaprice:


bobcaprice:
[Can't remove dates on affiliations](https://trello.com/c/iwhNAcIZ/3958-cant-remove-dates-on-affiliations)

PR #3924, opened 2017-08-07
When loading the list of delegates we are loading the full OrcidProfi…

amontenegro:
…le, so, for now lets change this so it loads only the bio


amontenegro:
[When loading delegates, dont load the full OrcidProfile, load only the Bio](https://trello.com/c/eQ1quhCr/3974-when-loading-delegates-dont-load-the-full-orcidprofile-load-only-the-bio)

PR #3925, opened 2017-08-07
When sending the reactivate email, we load the full orcid profile, mo…

amontenegro:
…dify the code so it only loads the BIO


amontenegro:
[500 error when sending a reactivate email when signing it with an iD](https://trello.com/c/EaYfvSW2/3976-500-error-when-sending-a-reactivate-email-when-signing-it-with-an-id)

PR #3926, opened 2017-08-07
Reduce the max elements on disk by 20% for our top 3 caches: full-wor…

amontenegro:
…k-entity, profile and pub-min-works-maps


amontenegro:
@wjrsimpson @rcpeters, 

This reduces 20% the maxElementsOnDisk for full-work-entity, profile and pub-min-works-maps which are the cache files that causes troubles. 

Along with this one https://github.com/ORCID/ORCID-Source/pull/3924 we will greatly reduce the disk space we use for caches.

Thanks

rcpeters:
@wjrsimpson @amontenegro should also be adding timeToIdleSeconds?

PR #3927, opened 2017-08-08
Fix apos and hard coded url

wjrsimpson:


wjrsimpson:
[Change warning when entering an email already in the system to point to auto deprecate](https://trello.com/c/oumFnhiB/3719-change-warning-when-entering-an-email-already-in-the-system-to-point-to-auto-deprecate)

PR #3928, opened 2017-08-08
Contact role editing rules

wjrsimpson:


wjrsimpson:
[Consortia self-service: Don't allow editing main contact](https://trello.com/c/5tQgLHRR/3516-consortia-self-service-dont-allow-editing-main-contact)

PR #3931, opened 2017-08-08
Load bio only on notifications page

amontenegro:


amontenegro:
[When loading notifications, dont load the full OrcidProfile, load only the bio and internal](https://trello.com/c/SfLWXDX1/3983-when-loading-notifications-dont-load-the-full-orcidprofile-load-only-the-bio-and-internal)

PR #3932, opened 2017-08-09
Correct v3 read me

bobcaprice:


bobcaprice:
[Record_3.0_dev1 REAME is not correct.](https://trello.com/c/GhMpFhGh/3984-https-githubcom-orcid-orcid-source-tree-master-orcid-model-src-main-resources-record30dev1)

bobcaprice:
[Record_3.0_dev1 REAME is not correct.](https://trello.com/c/GhMpFhGh/3984-https-githubcom-orcid-orcid-source-tree-master-orcid-model-src-main-resources-record30dev1)

PR #3935, opened 2017-08-10
Refresh token fixes

wjrsimpson:


wjrsimpson:
[Access token required when using a refresh token](https://trello.com/c/7yA3Zwwb/3997-access-token-required-when-using-a-refresh-token)

PR #3937, opened 2017-08-11
Fall back to actual account name, if there is no public display name

wjrsimpson:


wjrsimpson:
[Consortia self service: Orgs with no public display name have no name](https://trello.com/c/lzc6uryT/3875-consortia-self-service-orgs-with-no-public-display-name-have-no-name)

PR #3944, opened 2017-08-14
Cache eviction utilities

amontenegro:


amontenegro:
[Create a cron that remove evicted cached elements from the every cache.](https://trello.com/c/TNRIvmha/3978-create-a-cron-that-remove-evicted-cached-elements-from-the-every-cache)

PR #3946, opened 2017-08-16
Changed v3 URLs to use https as no conversion process to correct this…

bobcaprice:
… (as there is in 2.1)


bobcaprice:
[URIs using http protocol in v3.0_dev1](https://trello.com/c/saFDoF3t/4009-uris-using-http-protocol-in-v30dev1)

PR #3951, opened 2017-08-17
Corrected version of Title imported

bobcaprice:


bobcaprice:
[3.0 XML header has reference to xmlns:ns23="http://www.orcid.org/ns/orcid"](https://trello.com/c/AsAhIGio/4001-30-xml-header-has-reference-to-xmlnsns23http-wwworcidorg-ns-orcid)

PR #3954, opened 2017-08-17
Register multiple emails

lizkrznarich:
Starting with the existing form at /register . Will migrate to OAuth register and 2-screen register after a round of QA.


lizkrznarich:
[Request multiple emails at registration](https://trello.com/c/TcFfkOSA/3960-request-multiple-emails-at-registration)

PR #3956, opened 2017-08-17
4029 menu and content do not load on ie 10 11

mjcalvo21:


mjcalvo21:
[Menu and content do not load on IE 10 & 11](https://trello.com/c/cfga0OFF/4029-menu-and-content-do-not-load-on-ie-10-11)

PR #3958, opened 2017-08-18
Corrected external identifiers caching - omitted to remove caching fr…

bobcaprice:
…om v2 manager


bobcaprice:
[3.0: Error reading some records with the public API](https://trello.com/c/I0pmWc7P/4005-30-error-reading-some-records-with-the-public-api)

PR #3966, opened 2017-08-29
added workaround to deal with spring security userinfo requests

TomDemeranville:
Spring boot clients don't work because they send 'bearer' instead of 'Bearer' when requesting /oauth/userinfo.  Already an issue at spring, fixed in spring security 2.1.2, but not yet used by spring boot.

https://github.com/spring-projects/spring-security-oauth/issues/457


TomDemeranville:
[Workaround spring security 'bearer' instead of 'Bearer' bug for /oauth/userinfo](https://trello.com/c/zc9SMppU/4044-workaround-spring-security-bearer-instead-of-bearer-bug-for-oauth-userinfo)

rcpeters:
merging, but thinking one case-insensitive statement would be better

PR #3969, opened 2017-08-31
another try at fixing inconsistent whitespace in idp roundtrip alert

lizkrznarich:


lizkrznarich:
[Translations: All languages, Round trip notification is missing spaces before the provider and client name](https://trello.com/c/Ueo5yfVJ/4018-translations-all-languages-round-trip-notification-is-missing-spaces-before-the-provider-and-client-name)

PR #3970, opened 2017-08-31
Decouple profile entity from work entity

amontenegro:


amontenegro:
[Detach ProfileEntity from Work entity](https://trello.com/c/KjQJHv5h/4058-detach-profileentity-from-work-entity)

rcpeters:
😍 😍 😍 😍 😍 😍 😍 😍

amontenegro:
Hi @wjrsimpson could you please take a look at this? 

As a summary, 1.2 code will use LegacyWorkEntity and 2.0 will use WorkEntity, both of this classes will extend from WorkFullEntity which have all fields from work table but the orcid; then, LegacyWorkEntity will have the ProfileEntity referenced by orcid and WorkEntity will have just the orcid id.

Important to note in WorkEntity, **orcid** field will be insertable but not updatable, so, once a work have its orcid it will not change: 

_@Column(name = "orcid", updatable = false, insertable = true)_

orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidProfileManagerImpl.java
```diff
@@ -382,50 +382,63 @@ public OrcidProfile createOrcidProfileAndNotify(OrcidProfile orcidProfile) {
...
-        addSourceToEmails(orcidProfile, existingProfileEntity, amenderOrcid);
-
-        Boolean claimed = orcidProfile.getOrcidHistory() != null ? orcidProfile.getOrcidHistory().isClaimed() : existingProfileEntity.getClaimed();
-
-        if (orcidProfile.getOrcidInternal() != null && orcidProfile.getOrcidInternal().getPreferences() != null
-                && orcidProfile.getOrcidInternal().getPreferences().getActivitiesVisibilityDefault() != null) {
-            defaultVisibility = orcidProfile.getOrcidInternal().getPreferences().getActivitiesVisibilityDefault().getValue();
-        }
-        addDefaultVisibilityToBioItems(orcidProfile, defaultVisibility, claimed);
+        transactionTemplate.execute(new TransactionCallbackWithoutResult() {
```
amontenegro: Hi @wjrsimpson,

So here is where magic happens, instead of having a _@transactional_ on each method that invoke _updateOrcidProfile(OrcidProfile orcidProfile, UpdateOptions updateOptions)_, I removed all transactional annotations from them and execute this method into a single transaction.

Thanks

orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidProfileManagerImpl.java
```diff
@@ -382,50 +382,63 @@ public OrcidProfile createOrcidProfileAndNotify(OrcidProfile orcidProfile) {
...
+                        && orcidProfile.getOrcidInternal().getPreferences().getActivitiesVisibilityDefault() != null) {
+                    defaultVisibility = orcidProfile.getOrcidInternal().getPreferences().getActivitiesVisibilityDefault().getValue();
+                }
+                addDefaultVisibilityToBioItems(orcidProfile, defaultVisibility, claimed);
+
+                ProfileEntity profileEntity = adapter.toProfileEntity(orcidProfile, existingProfileEntity, updateOptions);
+                profileEntity.setLastModified(new Date());
+                profileEntity.setIndexingStatus(IndexingStatus.PENDING);
+                profileDao.flush();
+                ProfileEntity updatedProfileEntity = profileDao.merge(profileEntity);
```
amontenegro: @wjrsimpson: Then we update the PofileEntity and all its dependent elements (bio and activities but works)

orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidProfileManagerImpl.java
```diff
@@ -382,50 +382,63 @@ public OrcidProfile createOrcidProfileAndNotify(OrcidProfile orcidProfile) {
...
+                profileDao.refresh(updatedProfileEntity);
+                
+                // Then persist the works
+                if(orcidProfile.getOrcidActivities() != null && orcidProfile.getOrcidActivities().getOrcidWorks() != null) {
```
amontenegro: @wjrsimpson: And then we call adapter.setWorks which will iterate over the list of works and create/update/delete as needed.

After this, the transaction is closed, so, everything get stored in the DB at the same time

orcid-core/src/main/java/org/orcid/core/manager/impl/OrcidProfileManagerImpl.java
```diff
@@ -382,50 +382,63 @@ public OrcidProfile createOrcidProfileAndNotify(OrcidProfile orcidProfile) {
...
+                    adapter.setWorks(profileEntity, orcidProfile.getOrcidActivities().getOrcidWorks());
+                } else {
+                    adapter.setWorks(profileEntity, null);
+                }
+            }
+        });
+        
+        OrcidProfile profile = transactionTemplate.execute(new TransactionCallback<OrcidProfile>() {
```
amontenegro: And then we open another transaction to fetch the latest OrcidProfile with the latest changes

amontenegro:
Hi @wjrsimpson,

Please take a look at these changes, i couldn't find any bug, but I want you to take a look and hear your opinion about how I'm doing it.

Thanks

PR #3971, opened 2017-09-01
changed swagger return code for bulk works to 200

TomDemeranville:


TomDemeranville:
[Bulk work create swagger should handle 200 response.](https://trello.com/c/mtstNPgv/4056-bulk-work-create-swagger-should-handle-200-response)

PR #3973, opened 2017-09-01
Use replace all regex

amontenegro:


orcid-web/src/main/java/org/orcid/frontend/web/controllers/OpenIDController.java
```diff
@@ -75,9 +75,8 @@
         String authHeader = request.getHeader("Authorization"); //note we do not support form post per https://tools.ietf.org/html/rfc6750 because it's a MAY and pointless
         if (authHeader != null) {
             //lookup token, check it's valid, check scope.
-            String tokenValue = authHeader.replace("Bearer", "").trim();
             //deal with incorrect bearer case in request (I'm looking at you spring security!)
-            tokenValue = authHeader.replace("bearer", "").trim();
```
amontenegro: Hey @TomDemeranville this was wrong, it was replacing again in the authHeader, so, if the authHeader contains "Bearer xxxxxx", your replace will not work.

TomDemeranville: Looks good.  I fixed this myself earlier as well 👍

PR #3974, opened 2017-09-04
Hide roundtrip notification where no active token or it doesnt contai…

amontenegro:
…n some specific scopes


amontenegro:
[Always show roundtrip IdP notification unless there is an active token that allows adding affiliations](https://trello.com/c/wHlczZBd/4055-always-show-roundtrip-idp-notification-unless-there-is-an-active-token-that-allows-adding-affiliations)

PR #3976, opened 2017-09-05
Corrected category field for v3 record calls where no category specif…

bobcaprice:
…ied (ie just orcid id), now behaves as v2


bobcaprice:
[API Analytics: Add 3.0](https://trello.com/c/Vy2nr7s4/4042-api-analytics-add-30)

PR #3980, opened 2017-09-06
Added thread pooling for api analytics processing'

bobcaprice:


bobcaprice:
[API Analytics: Use thread pool](https://trello.com/c/p6M7esti/4041-api-analytics-use-thread-pool)

PR #3984, opened 2017-09-07
Ampersands in client names changed to + to avoid breaking display in GA

bobcaprice:


bobcaprice:
[API Analytics: Ampersands in client names break the client name display](https://trello.com/c/uvJ1qYky/4073-api-analytics-ampersands-in-client-names-break-the-client-name-display)

PR #3986, opened 2017-09-08
Disable access tokens on deactivated or deprecated records

amontenegro:


amontenegro:
[Revoke permissions when an account is deactivated](https://trello.com/c/oB7WBXYr/3733-revoke-permissions-when-an-account-is-deactivated)

PR #3990, opened 2017-09-08
Remove java web form classes that we dont use anymore

amontenegro:


amontenegro:
[Remove java web forms that we dont use anymore](https://trello.com/c/zizSqoDw/4087-remove-java-web-forms-that-we-dont-use-anymore)

PR #3992, opened 2017-09-11
Add sub member duplicate check msgs

lizkrznarich:


lizkrznarich:
[Consortia self-service: If a member org already exists in SF, inform the editor that existing info will be retrieved](https://trello.com/c/WQXFr5PK/4079-consortia-self-service-if-a-member-org-already-exists-in-sf-inform-the-editor-that-existing-info-will-be-retrieved)

lizkrznarich:
[Consortia self-service: Don't allow adding same org to Consortium members list multiple times](https://trello.com/c/Pjbh5EDG/4085-consortia-self-service-dont-allow-adding-same-org-to-consortium-members-list-multiple-times)

PR #3993, opened 2017-09-11
Add url to education and employment

bobcaprice:


bobcaprice:
[3.0_dev1 add url to education and employment](https://trello.com/c/ogm178B5/3993-30dev1-add-url-to-education-and-employment)

PR #3995, opened 2017-09-12
removed name and orcid from implicit response

TomDemeranville:


TomDemeranville:
[remove orcid and username from implicit response.](https://trello.com/c/xNNwhY6z/4098-remove-orcid-and-username-from-implicit-response)

PR #3996, opened 2017-09-12
Better implicit scope error

TomDemeranville:


TomDemeranville:
[Better error when using the wrong scopes with Implicit OAuth](https://trello.com/c/LovTaFGl/4097-better-error-when-using-the-wrong-scopes-with-implicit-oauth)

PR #4003, opened 2017-09-13
only check for submembers with the current member as parent account id

lizkrznarich:


lizkrznarich:
[Consortia self-service: If a member org already exists in SF, inform the editor that existing info will be retrieved](https://trello.com/c/WQXFr5PK/4079-consortia-self-service-if-a-member-org-already-exists-in-sf-inform-the-editor-that-existing-info-will-be-retrieved)

PR #4009, opened 2017-09-15
replace ftl string builtin used for converting numbers and booleans

lizkrznarich:


lizkrznarich:
[Replace ftl string builtins used with numbers and booleans](https://trello.com/c/iJNrg9Pv/4107-replace-ftl-string-builtins-used-with-numbers-and-booleans)

PR #4012, opened 2017-09-19
Use current flag for sf contacts

wjrsimpson:


wjrsimpson:
[Consortia self-service: Only show contacts with Current = true in membership contact role](https://trello.com/c/KUTd4iBP/4092-consortia-self-service-only-show-contacts-with-current-true-in-membership-contact-role)

wjrsimpson:
@lizkrznarich Can you take a quick look and merge this one?

PR #4020, opened 2017-09-21
disable signin button after click and move form#loginForm attr reset

lizkrznarich:


lizkrznarich:
[Clicking the sign in button multiple times messes up OAuth](https://trello.com/c/Kpq5piJ5/4122-clicking-the-sign-in-button-multiple-times-messes-up-oauth)

orcid-web/src/main/webapp/static/javascript/script.js
```diff
@@ -495,10 +495,16 @@ $(function() {
     $('form#loginForm')
             .submit(
                     function() {
+                        $('#form-sign-in-button').prop('disabled', true);
```
rcpeters: Login should never work if disable is before disabled check.

rcpeters: Ah I see. Let's move all the disabled here.

PR #4023, opened 2017-09-22
Changes for adding submembers

wjrsimpson:


orcid-core/src/main/java/org/orcid/core/manager/impl/SalesForceManagerImpl.java
```diff
@@ -436,6 +471,32 @@ public void createContact(Contact contact) {
...
 
+    private void createOpportunityContact(Contact contact, String opportunityId) {
+        String accountId = contact.getAccountId();
+        String contactOrcid = contact.getOrcid();
+        if (StringUtils.isBlank(contact.getEmail())) {
+            Email primaryEmail = emailManager.getEmails(contactOrcid).getEmails().stream().filter(e -> e.isPrimary()).findFirst().get();
+            contact.setEmail(primaryEmail.getEmail());
+        }
+
+        List<Contact> existingContacts = salesForceDao.retrieveAllContactsByAccountId(accountId);
```
lizkrznarich: @wjrsimpson I suspect we might be asked to check all SF contacts for matches, but we can roll with this for now.

wjrsimpson:
[Consortia self-service: Change add consortium member process](https://trello.com/c/DnR1BCTS/3999-consortia-self-service-change-add-consortium-member-process)

orcid-core/src/main/java/org/orcid/core/manager/impl/SalesForceManagerImpl.java
wjrsimpson: Yep, hoping we can discuss further at some point.

PR #4024, opened 2017-09-22
First attempt at sub-member removal approval process.

wjrsimpson:


wjrsimpson:
[Consortia self-service: Change remove consortium member process](https://trello.com/c/sStuK37G/4000-consortia-self-service-change-remove-consortium-member-process)

PR #4031, opened 2017-09-26
Use the source manager to set the activities source id in 1.2

amontenegro:


amontenegro:
[Check client_source_id is populated](https://trello.com/c/M197qRP8/2697-check-clientsourceid-is-populated)

PR #4032, opened 2017-09-27
added namespaced auth code generation,

TomDemeranville:
with a cache to prevent re-issuing the same code in a ten minute window

requires new config in persistence.properties: (defaults to 1,1)
org.orcid.core.node = 1
org.orcid.core.numberOfNodes = 1


TomDemeranville:
[Cache auth codes for 10 mins to ensure they're not re-issued](https://trello.com/c/ArxR5mYh/3849-cache-auth-codes-for-10-mins-to-ensure-theyre-not-re-issued)

PR #4036, opened 2017-09-28
Self service error msgs

lizkrznarich:


lizkrznarich:
[Self service: fix error messages when adding new orgs](https://trello.com/c/ExAhocoV/4121-self-service-fix-error-messages-when-adding-new-orgs)

PR #4037, opened 2017-09-28
Include obsolete and deprecated orgs even though they dont exists + r…

amontenegro:
…emove obsolete orgs from solr index


amontenegro:
[How do we manage obsolete institutes?](https://trello.com/c/INliBAGF/4114-how-do-we-manage-obsolete-institutes)

PR #4043, opened 2017-10-03
sanitize member slug and move init function call from template to con…

lizkrznarich:
…troller


lizkrznarich:
[XSS vulnerability detected](https://trello.com/c/Ll28ImrN/4163-xss-vulnerability-detected)

PR #4044, opened 2017-10-04
Fix where we set the content type and update to current RFC recommend…

rcpeters:
…ations


rcpeters:
[JSONP doesn't always return the correct content type.](https://trello.com/c/MnSDq7wq/4165-jsonp-doesnt-always-return-the-correct-content-type)

PR #4045, opened 2017-10-04
Improve disambiguated org validation

bobcaprice:


bobcaprice:
[3.0_dev1: Better error message when adding an affiliation without a org id](https://trello.com/c/QGv0ukao/4146-30dev1-better-error-message-when-adding-an-affiliation-without-a-org-id)

PR #4048, opened 2017-10-06
Disambiguated org public view

lizkrznarich:


lizkrznarich:
[Affiliation org ids: Move role/dept line per mockup](https://trello.com/c/FtCwgdVD/4144-affiliation-org-ids-move-role-dept-line-per-mockup)

lizkrznarich:
[Org identifier urls not linked correctly](https://trello.com/c/rtfMRlPZ/4149-org-identifier-urls-not-linked-correctly)

orcid-core/src/main/java/org/orcid/pojo/ajaxForm/AffiliationForm.java
```diff
@@ -30,8 +33,12 @@
 import org.orcid.jaxb.model.v3.dev1.record.AffiliationType;
 import org.orcid.jaxb.model.v3.dev1.record.Education;
 import org.orcid.jaxb.model.v3.dev1.record.Employment;
+import org.orcid.pojo.OrgDisambiguated;
 
 public class AffiliationForm implements ErrorsInterface, Serializable {
+    
+    @Resource
+    private OrgDisambiguatedManager orgDisambiguatedManager;
```
amontenegro: Hey @lizkrznarich, you don't need the OrgDisambiguatedManager here, it looks you don't use it anywhere and, as long as it is possible, we want to keep form elements as clean as possible and with no references to the managers, managers should be called from controllers only.

orcid-web/src/main/java/org/orcid/frontend/web/controllers/AffiliationsController.java
```diff
@@ -298,6 +298,14 @@ private void editAffiliation(AffiliationForm affiliationForm) throws Exception {
             for (Affiliation affiliation : affiliationsList) {
                 AffiliationForm form = AffiliationForm.valueOf(affiliation);
                 form.setCountryForDisplay(getMessage(buildInternationalizationKey(CountryIsoEntity.class, form.getCountry().getValue())));
+                if(form.getDisambiguatedAffiliationSourceId() != null){
```
amontenegro: Hey @lizkrznarich, just one comment here, are we sure that if **form.getDisambiguatedAffiliationSourceId()** is not null, then **form.getOrgDisambiguatedId()** will not be null either? why not changing the if condition to check for the presence of **form.getOrgDisambiguatedId()** instead?

orcid-web/src/main/java/org/orcid/frontend/web/controllers/PublicProfileController.java
```diff
@@ -559,6 +564,14 @@ private boolean isProfileValidForIndex(ProfileEntity profile) {
             validateVisibility(aff.getVisibility());
             AffiliationForm form = AffiliationForm.valueOf(aff);
             form.setCountryForDisplay(getMessage(buildInternationalizationKey(CountryIsoEntity.class, aff.getOrganization().getAddress().getCountry().name())));
+            if(form.getDisambiguatedAffiliationSourceId() != null){
```
amontenegro: Same as affiliationsController if condition.

PR #4051, opened 2017-10-06
Api12 fix error with old client ids going to source

amontenegro:


amontenegro:
[Check client_source_id is populated](https://trello.com/c/M197qRP8/2697-check-clientsourceid-is-populated)

PR #4056, opened 2017-10-10
Remove locked, deprecated or deactivated records from the solr indexing

amontenegro:


amontenegro:
[Locked and deprecated records in search results](https://trello.com/c/xamnOwt4/4160-locked-and-deprecated-records-in-search-results)

orcid-message-listener/src/main/java/org/orcid/listener/solr/SolrMessageProcessor.java
```diff
@@ -83,6 +83,12 @@ private void updateSolrIndex(String orcid) {
         }
         try{
             org.orcid.jaxb.model.record_v2.Record record = orcid20ApiClient.fetchPublicRecord(orcid); 
+            
+            // Remove deactivated records from SOLR index
+            if (record.getHistory() != null && record.getHistory().getDeactivationDate() != null && record.getHistory().getDeactivationDate().getValue() != null) {
+                solrUpdater.deleteById(orcid);
```
TomDemeranville: You check for a deactivated exception below.  Are both checks necessary?

orcid-persistence/src/main/resources/db/updates/reindex_locked_deprecated_and_deactivated_records.xml
```diff
@@ -0,0 +1,26 @@
...
+
+    =============================================================================
+
+-->
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+
+	<changeSet id="REINDEX-LOCKED-DEPRECATED-AND-DEACTIVATED-RECORDS" author="Angel Montenegro" dbms="postgresql">
+		<sql>UPDATE profile SET indexing_status='REINDEX' WHERE orcid IN (SELECT orcid FROM profile WHERE record_locked = true OR primary_record IS NOT NULL OR profile_deactivation_date IS NOT NULL);</sql>
```
TomDemeranville: I'm a bit confused what happens if this is deployed on more than one node?

orcid-message-listener/src/main/java/org/orcid/listener/solr/SolrMessageProcessor.java
amontenegro: Na @TomDemeranville, deactivated records don't throw exceptions, they just return an empty Record with the deactivation date, that's why we should look for it manually.

The exceptions below catch locked or deprecated records.

orcid-persistence/src/main/resources/db/updates/reindex_locked_deprecated_and_deactivated_records.xml
amontenegro: Hey @TomDemeranville, every liquibase scripts runs in one node and lock the database, when it finish, it mark the script as ran and release the lock, so, the next node will wait until the first one finishes, and then it will notice that the script already ran and will just ignore it.

orcid-message-listener/src/main/java/org/orcid/listener/solr/SolrMessageProcessor.java
TomDemeranville: My mistake!

orcid-persistence/src/main/resources/db/updates/reindex_locked_deprecated_and_deactivated_records.xml
TomDemeranville: Awesome.

rcpeters:
merge monster

amontenegro:
Fixed @rcpeters

PR #4057, opened 2017-10-10
Fix error where it mark as duplicate a PART-OF work because there was…

amontenegro:
… a SELF work with the same id


amontenegro:
[bulk works: Error when adding works where an identifier is self on one and part of on the other](https://trello.com/c/oCo0sx17/4135-bulk-works-error-when-adding-works-where-an-identifier-is-self-on-one-and-part-of-on-the-other)

PR #4058, opened 2017-10-10
fix html error in login form templates

lizkrznarich:


lizkrznarich:
[Register form doesn't load on /signin or oauth](https://trello.com/c/SME4o39k/4185-register-form-doesnt-load-on-signin-or-oauth)

PR #4060, opened 2017-10-11
Member list show badges

lizkrznarich:


lizkrznarich:
[Member list: Add collect & connect badges](https://trello.com/c/AvvLV3R7/4167-member-list-add-collect-connect-badges)

PR #4061, opened 2017-10-11
Added check for null country in disambiguated org.

wjrsimpson:


wjrsimpson:
[Broken affiliations](https://trello.com/c/hWbBgtb1/4187-broken-affiliations)

PR #4062, opened 2017-10-11
Pretty id list

TomDemeranville:
Updated properties to use utf-8 umlaut
Added xsl sort


TomDemeranville:
[Easy to read identifier list page](https://trello.com/c/yFNRaig0/3965-easy-to-read-identifier-list-page)

PR #4063, opened 2017-10-11
Return empty OrcidProfile for locked, deactivated or deprecated records

amontenegro:


amontenegro:
[Locked and deprecated records in search results](https://trello.com/c/xamnOwt4/4160-locked-and-deprecated-records-in-search-results)

PR #4064, opened 2017-10-11
fix problem showing org id popover at small screen sizes

lizkrznarich:


lizkrznarich:
[Fix org ID popover at smaller screen sizes](https://trello.com/c/ra4K3BbF/4133-fix-org-id-popover-at-smaller-screen-sizes)

PR #4066, opened 2017-10-12
Fixed so that email/id is passed to reactivation email function.

wjrsimpson:


wjrsimpson:
[Error when attempting to send reactivate email message](https://trello.com/c/QuxX3hxW/4181-error-when-attempting-to-send-reactivate-email-message)

PR #4067, opened 2017-10-12
add identifiers properties to transifex config

lizkrznarich:


lizkrznarich:
[Easy to read identifier list page](https://trello.com/c/yFNRaig0/3965-easy-to-read-identifier-list-page)

PR #4074, opened 2017-10-13
http://test.orcid.org is now valid, since cloudfare forward it to orc…

amontenegro:
…id.org


rcpeters:
That's not a cloudflare thing?

amontenegro:
Hi @rcpeters ,

http://test.orcid.org used to reply 404, however, since yesterday it now replies 301, which I assumed was related with Cloudflare, isn't it? 

Thanks

rcpeters:
nope.

PR #4077, opened 2017-10-13
Display grid external ids on ui

lizkrznarich:


lizkrznarich:
[Display GRID external identifiers in the UI](https://trello.com/c/9jz0TWjj/4193-display-grid-external-identifiers-in-the-ui)

PR #4085, opened 2017-10-17
Cache message listener api requests

TomDemeranville:


TomDemeranville:
[Cache v2 API requests on message listener to reduce traffic](https://trello.com/c/m0lOGH5Y/4205-cache-v2-api-requests-on-message-listener-to-reduce-traffic)

PR #4090, opened 2017-10-17
fix bio privacy popover alignment

lizkrznarich:


lizkrznarich:
[Bio mouse over is not aligned.](https://trello.com/c/yussOwHs/4215-bio-mouse-over-is-not-aligned)

PR #4094, opened 2017-10-18
Fundref importer now marks deprecated orgs instead of setting them as…

amontenegro:
… external identifiers


amontenegro:
[Fix external identifiers for FundRef organization](https://trello.com/c/uL6K3z22/4194-fix-external-identifiers-for-fundref-organization)

amontenegro:
@wjrsimpson fixed!

PR #4104, opened 2017-10-20
Modify CLI so they can include openid scope and implicit grant type

amontenegro:


amontenegro:
[Turn on implict grant type and openid scope for existing members](https://trello.com/c/GtKoC27e/4199-turn-on-implict-grant-type-and-openid-scope-for-existing-members)

amontenegro:
Hi @TomDemeranville you were right, I just updated the code so it includes the public client as well.

Thanks

PR #4105, opened 2017-10-20
Add implicit grant type and open id scope by default to new clients

amontenegro:


amontenegro:
[Ensure implicit grant type and openid scope are enabled by default for new clients.](https://trello.com/c/C8swCrZT/4200-ensure-implicit-grant-type-and-openid-scope-are-enabled-by-default-for-new-clients)

PR #4108, opened 2017-10-24
Mongo message listener

TomDemeranville:
This change adds Mongo to the list of message listeners, if enabled.
Unit testing done with Fongo (fake mongo) and also tested against a real mongo instance.

Listens to LastUpdatedQueue and adds/updates as appropriate.

If enabled, will connect to mongo. If not enabled, server starts fine and mongo is ignored.

Config in properties file:
org.orcid.message-listener.mongo.enabled=false

Once mongo instance deployed, add:

org.orcid.message-listener.mongo.uri=mongodb://localhost (or whatever)
org.orcid.message-listener.mongo.database=messagelistener
org.orcid.message-listener.mongo.collection=dump

Mongo URI can include security info if we want a secure DB.


TomDemeranville:
[Add mongo to list of message listeners](https://trello.com/c/aszIrqs9/4242-add-mongo-to-list-of-message-listeners)

orcid-message-listener/src/main/java/org/orcid/listener/orcid/Orcid20APIClient.java
```diff
@@ -60,11 +65,17 @@ public Orcid20APIClient(@Value("${org.orcid.message-listener.api20BaseURI}") Str
...
      * 
+     * Caches based on message.
+     * 
      * @param orcid
      * @return Record
      */
-    public Record fetchPublicRecord(String orcid) throws LockedRecordException, DeprecatedRecordException {
-        WebResource webResource = jerseyClient.resource(baseUri).path(orcid + "/record");
+    public Record fetchPublicRecord(BaseMessage message) throws LockedRecordException, DeprecatedRecordException {
+        Record cached = v2PerMessageQueue.getIfPresent(message);
```
amontenegro: Hey @TomDemeranville, 

Shouldn't we implement hashcode and equals on the BaseMessage? if not, how the cache ensures that the given message is exactly the same as the one cached?

Thanks

orcid-message-listener/src/main/resources/log4j.xml
```diff
@@ -46,7 +46,7 @@
 	<!-- Root Logger -->
 	<root>
 		<level value="INFO" />
-		<!-- <appender-ref ref="CONSOLE" /> -->
+		<appender-ref ref="CONSOLE" />
```
amontenegro: Do we really want to send all login to the console as well?

orcid-message-listener/src/main/java/org/orcid/listener/orcid/Orcid20APIClient.java
TomDemeranville: Ah, I think it's abstract, so it would be the implementations that get called, right?  I will add them to RetryMessage and LastModifiedMessage

orcid-message-listener/src/main/resources/log4j.xml
TomDemeranville: Nope, that was from my debugging.  I'll turn it off.

orcid-message-listener/src/main/java/org/orcid/listener/orcid/Orcid20APIClient.java
TomDemeranville: Actually, you're right.  Both subclasses have all their state in the BaseMessage map.  I'll put the equals there.

PR #4114, opened 2017-10-25
Add normalisation engine doi case

TomDemeranville:
added identifier normalisation chain of responsibility engine
added DOI normaliser
added case sensitive normaliser


TomDemeranville:
[Normalisation framework and DOI normaliser. Add case sensitive normaliser and use both for grouping](https://trello.com/c/sYu5eDxI/4245-normalisation-framework-and-doi-normaliser-add-case-sensitive-normaliser-and-use-both-for-grouping)

wjrsimpson:
@TomDemeranville merge monster!

PR #4115, opened 2017-10-25
Return https on 2.1 search results and works bulk

amontenegro:


amontenegro:
[2.1 Search Results use http](https://trello.com/c/Grvq3Vmb/4249-21-search-results-use-http)

PR #4121, opened 2017-10-26
Register OrcidError map from 2.0 to 2.1

amontenegro:


amontenegro:
[2.1 reading bulk works if an invalid put-code is included a 500 error is returned instead of a 400](https://trello.com/c/aOopL6IO/4273-21-reading-bulk-works-if-an-invalid-put-code-is-included-a-500-error-is-returned-instead-of-a-400)

PR #4127, opened 2017-10-27
fix typo in account deprecation confirmation modal

lizkrznarich:


lizkrznarich:
[Duplicate removed message ends in a comma instead of a period](https://trello.com/c/s843l6TA/4281-duplicate-removed-message-ends-in-a-comma-instead-of-a-period)

PR #4129, opened 2017-10-30
Add v3 managers to all controllers

bobcaprice:


bobcaprice:
[Make controllers in orcid-web use V3 managers](https://trello.com/c/zhc50fL7/4258-make-controllers-in-orcid-web-use-v3-managers)

PR #4132, opened 2017-10-31
Removed unused putCode field from affiliation json object

bobcaprice:


bobcaprice:
[affiliations external ids UI: Don't send empty put-codes](https://trello.com/c/f0LsJpGv/4226-affiliations-external-ids-ui-dont-send-empty-put-codes)

PR #4137, opened 2017-11-01
Fixed v3/v2 conversion in switch user.

wjrsimpson:


wjrsimpson:
[Switch user not working](https://trello.com/c/JcWl25vN/4289-switch-user-not-working)

PR #4138, opened 2017-11-01
Change work bulk json validation

bobcaprice:


bobcaprice:
[Validate each work separately for json bulk create](https://trello.com/c/LFIZmMzM/4288-validate-each-work-separately-for-json-bulk-create)

PR #4139, opened 2017-11-01
updated openID id_token to use https://orcid.org/0000-0000-0000-0000 …

TomDemeranville:
…as subject

put in new id_token property "id_path" containing 0000-0000-0000-0000


TomDemeranville:
[OpenID id_token subject is just an ORCID without a protocol. Should it include domain and http or https?](https://trello.com/c/f04f1j82/3845-openid-idtoken-subject-is-just-an-orcid-without-a-protocol-should-it-include-domain-and-http-or-https)

PR #4142, opened 2017-11-01
udpate affiliation org id other ext ids text

lizkrznarich:


lizkrznarich:
[Affiliation org ids: Other ids text change](https://trello.com/c/RAhFSWyk/4286-affiliation-org-ids-other-ids-text-change)

PR #4143, opened 2017-11-01
fix peer review section header width so text isn't cut off in some la…

lizkrznarich:
…nguages


lizkrznarich:
[Count of peer-review items on the public view is missing in many languages](https://trello.com/c/5ZaBSFtQ/4295-count-of-peer-review-items-on-the-public-view-is-missing-in-many-languages)

PR #4144, opened 2017-11-01
fix links in trusted indiv search results

lizkrznarich:


lizkrznarich:
[Bad links to records on trusted individual search](https://trello.com/c/SBUW7ICN/4296-bad-links-to-records-on-trusted-individual-search)

PR #4146, opened 2017-11-02
Allow consortium admin to be member admin too

wjrsimpson:


wjrsimpson:
[500 error updating consortium from /manage-members](https://trello.com/c/WeLQJMFU/4179-500-error-updating-consortium-from-manage-members)

PR #4148, opened 2017-11-02
Fixed v2/v3 comparison for switching user as admin.

wjrsimpson:


wjrsimpson:
[Admin switch user functionality does not work](https://trello.com/c/wV4MPE63/4299-admin-switch-user-functionality-does-not-work)

PR #4150, opened 2017-11-02
remove extra code left in affiliation srv

lizkrznarich:


lizkrznarich:
[Works section on some records doesn't load on public record](https://trello.com/c/b6lFRqVZ/4291-works-section-on-some-records-doesnt-load-on-public-record)

PR #4151, opened 2017-11-03
made swagger avaiable on the 3.0 endpoint and added it to the list of…

TomDemeranville:
… things Swagger documents.


TomDemeranville:
[Make swagger work for v3 API](https://trello.com/c/PpmvSbYI/4257-make-swagger-work-for-v3-api)

PR #4154, opened 2017-11-03
Updated SPEL to use v3 manager.

wjrsimpson:


wjrsimpson:
[Member developer tools not working](https://trello.com/c/3RUEvHKk/4305-member-developer-tools-not-working)

PR #4163, opened 2017-11-07
Blank non normal

TomDemeranville:


TomDemeranville:
[Make normalizers return empty strings for unmatched id values](https://trello.com/c/OhkS9AnL/4318-make-normalizers-return-empty-strings-for-unmatched-id-values)

PR #4167, opened 2017-11-07
fix missing org info in add funding modal

lizkrznarich:


lizkrznarich:
[Null info for funding orgs](https://trello.com/c/NJdlSGiv/4325-null-info-for-funding-orgs)

PR #4169, opened 2017-11-08
Fix minor typo in API readme

davidjb:


rcpeters:
Not sure about this one. https://www.merriam-webster.com/words-at-play/when-to-use-it's-vs-its The item has a visibility.

alainna:
Per modern orthographic practices, the word should be "its" here (i support the change).

davidjb:
`it's` is a contraction of `it is`.  So if we change the sentence to read:

- *last-modified-date* when the item was changed, such as being edited or having **it is** visibility changed. 

you can see that isn't correct.  In this context, `its` is the possessive, referring the *visibility* of the *item*.

From the linked site:

> The rule is actually pretty simple: use the apostrophe after it only when part of a word has been removed: it's raining means it is raining; it's been warm means it has been warm. It's is a contraction, in the style of can't for cannot and she's for she is.

PR #4171, opened 2017-11-08
Fix dupe contact roles

wjrsimpson:


wjrsimpson:
[Consortia self-service: Check membership contact roles for existing non-current record before adding new record](https://trello.com/c/UQsT97TC/4180-consortia-self-service-check-membership-contact-roles-for-existing-non-current-record-before-adding-new-record)

PR #4175, opened 2017-11-09
Load ringgold data v2

amontenegro:


amontenegro:
[Update Ringgold import include provided ISNI mappings.](https://trello.com/c/IWEPJ3pc/4267-update-ringgold-import-include-provided-isni-mappings)

PR #4176, opened 2017-11-10
Lets set the mongo procesor disabled by default

amontenegro:


amontenegro:
Hey @TomDemeranville! FYI, I set org.orcid.message-listener.mongo.enabled to false by default, so, if it is not in the config file it is just ignored.

PR #4177, opened 2017-11-10
fix revoke delegate modal missing when https ids feature disabled

lizkrznarich:


lizkrznarich:
[Can't revoke trusted individual with HTTPS_IDS disabled](https://trello.com/c/rP9job6Y/4334-cant-revoke-trusted-individual-with-httpsids-disabled)

PR #4178, opened 2017-11-10
Remove OpenEntityInView filter from our web project

amontenegro:


amontenegro:
[Lets get rid of OpenEntityManagerInViewFilter from our orcid-web project](https://trello.com/c/ZGivyl37/4324-lets-get-rid-of-openentitymanagerinviewfilter-from-our-orcid-web-project)

orcid-web/src/main/java/org/orcid/frontend/spring/OrcidAuthenticationProvider.java
```diff
@@ -91,24 +97,26 @@ public Authentication authenticate(Authentication auth) throws AuthenticationExc
...
         
-        for (EmailEntity email : emails) {
-            if (email.getPrimary()) {
-                OrcidType orcidType = OrcidType.valueOf(profileEntity.getOrcidType().name());
-                return new OrcidProfileUserDetails(orcid, email.getId(), profileEntity.getPassword(), orcidType);
-            }
-        }
-        return null;
+        try {
+            EmailEntity primaryEmail = emailDaoReadOnly.findPrimaryEmail(orcid);
```
amontenegro: Hey @wjrsimpson @rcpeters, 

Please notice this change, before it, we used to get the primary email from the ProfileEntity object, however, after removing the OpenEntityInView filter that wouldn't work anymore; so, I did the following changes: 

1. Instead of reading the profile entity from the ProfileDao, now we will use the ProfileEntityCacheManager (Which reads from the replication DB)
2. We will fetch the primary email using the EmailDaoReadOnly which reads from the replication DB as well
3. Now, if someone tries to login with a record without primary email, we will send a notification so we know about it.

Thanks

PR #4182, opened 2017-11-14
show https in id banner

lizkrznarich:
show https in id banner on my-orcid, public record and print view


lizkrznarich:
[Show full iD URL in left column](https://trello.com/c/giZCRVmz/4338-show-full-id-url-in-left-column)

PR #4184, opened 2017-11-14
reg multi email duplicate check use lower case

lizkrznarich:


lizkrznarich:
[Multi-email registration: A difference in capitalization between the primary and additional emails throws a contact support error](https://trello.com/c/aG8rJsOZ/4339-multi-email-registration-a-difference-in-capitalization-between-the-primary-and-additional-emails-throws-a-contact-support-error)

PR #4185, opened 2017-11-14
don't show commas in disambiguated org details before empty fields

lizkrznarich:
don't show commas in edit UI disambiguated org details when fields are empty


lizkrznarich:
[How to display Fundref orgs with no city given](https://trello.com/c/xIcC9Pc1/4336-how-to-display-fundref-orgs-with-no-city-given)

PR #4186, opened 2017-11-14
Revoke endpoint

amontenegro:


amontenegro:
[/oauth/revoke](https://trello.com/c/mog1N8NP/4266-oauth-revoke)

PR #4187, opened 2017-11-14
move cursor to 2FA input automatically

lizkrznarich:


lizkrznarich:
[2FA: Automatically put the cursor in the Authenticate Code box](https://trello.com/c/7gXKui51/4335-2fa-automatically-put-the-cursor-in-the-authenticate-code-box)

PR #4188, opened 2017-11-15
Add revocation_date to the tokens table and populate it when a token …

amontenegro:
…is revocated


amontenegro:
[/oauth/revoke](https://trello.com/c/mog1N8NP/4266-oauth-revoke)

PR #4195, opened 2017-11-17
don't make agreement signatory required

lizkrznarich:


lizkrznarich:
[Self-service: Don't make Agreement signatory contact required](https://trello.com/c/PPSxc3h3/4349-self-service-dont-make-agreement-signatory-contact-required)

PR #4196, opened 2017-11-17
fix npe when canceling removal request

lizkrznarich:


lizkrznarich:
[Self-service: 500 error when canceling removal request](https://trello.com/c/1dorpDxM/4352-self-service-500-error-when-canceling-removal-request)

PR #4199, opened 2017-11-20
Api version 2.X and 3.X should allow HTTPS calls only

amontenegro:


amontenegro:
[2.1 API allows HTTP](https://trello.com/c/BZnhr09P/4355-21-api-allows-http)

PR #4200, opened 2017-11-20
Allow 1.x update scopes to be used to add the corresponding activity type in 2.x.

wjrsimpson:


wjrsimpson:
[Allow legacy update scopes for adding activities in 2.x](https://trello.com/c/26qobBsX/4356-allow-legacy-update-scopes-for-adding-activities-in-2x)

PR #4208, opened 2017-11-23
Fix swagger private month

TomDemeranville:


TomDemeranville:
[Fix swagger for private and fix example day value](https://trello.com/c/8jSBD43D/4370-fix-swagger-for-private-and-fix-example-day-value)

PR #4209, opened 2017-11-23
Only list consortium if the consortium opportunity is won, closed, and current.

wjrsimpson:


wjrsimpson:
[Member transitioning to consortium shows on consortia list](https://trello.com/c/DOBcOEvv/4365-member-transitioning-to-consortium-shows-on-consortia-list)

PR #4211, opened 2017-11-27
fixed group id missing exception for API

TomDemeranville:


TomDemeranville:
[Wrong error message for non-existing group id](https://trello.com/c/OKAIGOCb/4367-wrong-error-message-for-non-existing-group-id)

PR #4213, opened 2017-11-28
Ringgold importer filter identifiers

amontenegro:


amontenegro:
[Refactor ringgold importer to allow filtering identifiers by type](https://trello.com/c/gYvgHMzv/4381-refactor-ringgold-importer-to-allow-filtering-identifiers-by-type)

PR #4214, opened 2017-11-28
Fix reference

amontenegro:


amontenegro:
[Switching to a trusted account from the more delegates page does not work](https://trello.com/c/u0OtrpiJ/4375-switching-to-a-trusted-account-from-the-more-delegates-page-does-not-work)

PR #4221, opened 2017-11-30
Urls to2000 chars

amontenegro:


amontenegro:
[Work url length limit is too small](https://trello.com/c/WrN50Lre/4220-work-url-length-limit-is-too-small)

PR #4222, opened 2017-11-30
The index name in the precondition is wrong

amontenegro:


amontenegro:
@wjrsimpson @rcpeters FYI, if we want to generate the unique constraint before the release, we should include this change.

I already removed the entry from the database change log so liquibase doesn't fail on startup.

Thanks

PR #4223, opened 2017-12-01
Modified exception mapper to contain full reason as thrown by the code:

TomDemeranville:
- Gets a message for the error code from the properties. If no message, uses the full class name.
- Appends a localised version of "Full validation error:" followed by the message embedded in the code (in English)


TomDemeranville:
[Wrong error message for bad ID type](https://trello.com/c/K8NcwEJc/4366-wrong-error-message-for-bad-id-type)

PR #4224, opened 2017-12-01
Modify cron task so it reindex any org with indexing status different…

amontenegro:
… than DONE


amontenegro:
[Cron job that reindex disambiguated orgs only picks orgs with PENDING status](https://trello.com/c/NldHSmze/4387-cron-job-that-reindex-disambiguated-orgs-only-picks-orgs-with-pending-status)

PR #4226, opened 2017-12-04
Sort by date now defaults to descending order, sort icon updated to r…

bobcaprice:
…eflect this


bobcaprice:
[Works default sort is reversed](https://trello.com/c/4aYXDE7B/4393-works-default-sort-is-reversed)

PR #4227, opened 2017-12-04
Confirm delete dialog now shown when single work delete button pressed

bobcaprice:


bobcaprice:
[Works when there is only one work the delete button needs to go to confirm delete](https://trello.com/c/3g7Cbdm4/4396-works-when-there-is-only-one-work-the-delete-button-needs-to-go-to-confirm-delete)

bobcaprice:
[Keep existing preferred source display](https://trello.com/c/eYAysUlP/4391-keep-existing-preferred-source-display)

PR #4228, opened 2017-12-04
Sorting by title now case insensitive

bobcaprice:


bobcaprice:
[Works sort by title should not be case sensitive](https://trello.com/c/qYgxW6dH/4394-works-sort-by-title-should-not-be-case-sensitive)

PR #4229, opened 2017-12-04
Fixed refresh works for cases where all works were loaded on page

bobcaprice:


bobcaprice:
[Works making any changes makes all works disappear](https://trello.com/c/78aQ4PYJ/4389-works-making-any-changes-makes-all-works-disappear)

PR #4230, opened 2017-12-04
Fixed closing sources bug where work group disappears

bobcaprice:


bobcaprice:
[After viewing sources on a work there's an extra bar on the screen](https://trello.com/c/sTZtqmtB/4390-after-viewing-sources-on-a-work-theres-an-extra-bar-on-the-screen)

PR #4232, opened 2017-12-05
Fixed sources visibility display

bobcaprice:


bobcaprice:
Oops I've missed a bit, please don't merge just yet

bobcaprice:
Ok committed that extra bit now, please merge away

bobcaprice:
[Works when sources are open the visibility icon does not show the current visibility and can't be used to change the visibility](https://trello.com/c/kQGPWli4/4397-works-when-sources-are-open-the-visibility-icon-does-not-show-the-current-visibility-and-cant-be-used-to-change-the-visibility)

PR #4233, opened 2017-12-05
Fix issue where all locales default to English

amontenegro:


amontenegro:
[Records always default to English on sign in and all emails are sent in English](https://trello.com/c/RKYcfEAm/4400-records-always-default-to-english-on-sign-in-and-all-emails-are-sent-in-english)

PR #4234, opened 2017-12-05
Tweaked works label to fix bug where help icon goes to next line

bobcaprice:


bobcaprice:
[Works if the number of works is 4 digits the help icon is on it's own line](https://trello.com/c/WshDzEhW/4395-works-if-the-number-of-works-is-4-digits-the-help-icon-is-on-its-own-line)

PR #4235, opened 2017-12-05
Works copy and edit button

bobcaprice:


bobcaprice:
[Works copy and edit option is always grayed out](https://trello.com/c/IOT2vB6H/4392-works-copy-and-edit-option-is-always-grayed-out)

PR #4237, opened 2017-12-05
Single query to set the primary email

amontenegro:


amontenegro:
[Add backend validation to prevent a user to end up with no primary email](https://trello.com/c/w4mdtIm4/4300-add-backend-validation-to-prevent-a-user-to-end-up-with-no-primary-email)

amontenegro:
Hi @wjrsimpson, could you please take a look at this PR?

Thanks

orcid-persistence/src/main/java/org/orcid/persistence/dao/impl/EmailDaoImpl.java
```diff
@@ -78,13 +78,10 @@ public void updateEmail(String orcid, String email, boolean isCurrent, Visibilit
...
     @Transactional
     public void updatePrimary(String orcid, String primaryEmail) {
-        Query updatePrimaryQuery = entityManager.createQuery("update EmailEntity set primary = 'true' where orcid = :orcid and email = :primaryEmail");
-        Query updateNonPrimaryQuery = entityManager.createQuery("update EmailEntity set primary = 'false' where orcid = :orcid and email != :primaryEmail");
-        for (Query query : new Query[] { updatePrimaryQuery, updateNonPrimaryQuery }) {
-            query.setParameter("orcid", orcid);
-            query.setParameter("primaryEmail", primaryEmail);
-            query.executeUpdate();
-        }
+        Query query = entityManager.createNativeQuery("UPDATE email SET is_primary= CASE email WHEN :primaryEmail THEN true ELSE false END WHERE orcid = :orcid");
```
rcpeters: Good query, but what if I use `INFO@rcpeters.com`?

amontenegro: Hey @rcpeters, this code is only used by the account settings page "Set Primary", so, it assumes the email it gets is exactly as the one in the database, we can add some more validations to it, but, it has been working like this for a while and there is no other place where we use it.

email_set_primary

PR #4240, opened 2017-12-06
Preferred source not marked

bobcaprice:


bobcaprice:
[When looking at multiple sources the preferred version is not marked](https://trello.com/c/HUmaTFlR/4401-when-looking-at-multiple-sources-the-preferred-version-is-not-marked)

PR #4241, opened 2017-12-06
Works with null publication dates will be sorted by title

bobcaprice:


bobcaprice:
[Works without dates on date sort should be in alphabetical order](https://trello.com/c/t61ly46T/4406-works-without-dates-on-date-sort-should-be-in-alphabetical-order)

PR #4242, opened 2017-12-06
Works refreshed after bibtext import

bobcaprice:


bobcaprice:
[BibTeX import saved works don't display until page is refreshed](https://trello.com/c/ZVjgPN6x/4408-bibtex-import-saved-works-dont-display-until-page-is-refreshed)

PR #4243, opened 2017-12-06
Works no longer reloaded after visibility updated

bobcaprice:


bobcaprice:
don't merge for a moment please, about to push some tidy up

bobcaprice:
ok go for it

bobcaprice:
[Don't reload all works when the visibility of one work is changed](https://trello.com/c/bCSQ02sx/4404-dont-reload-all-works-when-the-visibility-of-one-work-is-changed)

PR #4244, opened 2017-12-06
Works no longer reloaded after changing preferred

bobcaprice:


bobcaprice:
[Switching preferred work version reloads all works and scrolls browser up to the top of the works section](https://trello.com/c/OgsFuM0p/4402-switching-preferred-work-version-reloads-all-works-and-scrolls-browser-up-to-the-top-of-the-works-section)

PR #4245, opened 2017-12-06
Clicking bulk edit now loads all works

bobcaprice:


bobcaprice:
[Bulk edit- add option to select all works, not just works visible on the page](https://trello.com/c/gEQX39Nr/4410-bulk-edit-add-option-to-select-all-works-not-just-works-visible-on-the-page)

PR #4246, opened 2017-12-06
Fixed public page works, which was still calling deleted js function

bobcaprice:


bobcaprice:
[fix public work page](https://trello.com/c/G9vbrwMm/4412-fix-public-work-page)

PR #4247, opened 2017-12-07
Work details sections closed when deleting works or importing bibtex

bobcaprice:


bobcaprice:
[Work details won't load if the details section was open when making other changes](https://trello.com/c/omlgZn3a/4413-work-details-wont-load-if-the-details-section-was-open-when-making-other-changes)

PR #4248, opened 2017-12-07
Visibility indicator now updated when sources open

bobcaprice:


bobcaprice:
[Can't edit visibility when sources is open](https://trello.com/c/psRIqf8f/4414-cant-edit-visibility-when-sources-is-open)

PR #4251, opened 2017-12-07
Reversed secondary title sort for date sort where pub dates are null

bobcaprice:


bobcaprice:
[Works without dates on date sort should be in alphabetical order](https://trello.com/c/t61ly46T/4406-works-without-dates-on-date-sort-should-be-in-alphabetical-order)

PR #4254, opened 2017-12-08
Affiliations sort updates

amontenegro:


amontenegro:
Hey @lizkrznarich, could you please take a look at the frontend changes?

I changed the logic for the date sort string, so, it works like this: 

1) Z-YYYY-MM-DD: Where the date part (YYYY-MM-DD) is the creation date of the element (We have a couple of cases like this)
2) Y-YYYY-MM-DD: Where the date part (YYYY-MM-DD) is the start date of the affiliation
3) X-YYYY-MM-DD-YYYY-MM-DD: Where the first date belongs to the end date of the affiliation and the second date belogs to the start date of the affiliation

With that it is easy to sort them by end date and then by start date, and, we leave the ones with none of them to start with _Z-_ so we can easily make them float to the top

amontenegro:
[Affiliations, funding: On UI, sort by end date by default, add new sort options "By start date" and "By end date"](https://trello.com/c/kSsKTR81/4378-affiliations-funding-on-ui-sort-by-end-date-by-default-add-new-sort-options-by-start-date-and-by-end-date)

orcid-core/src/main/java/org/orcid/pojo/ajaxForm/PojoUtil.java
```diff
@@ -141,7 +146,99 @@ public static String createDateSortString(org.orcid.jaxb.model.v3.dev1.common.Fu
...
+                dateSortString = "Y-";
+                dateSortString += startDate.getYear() == null ? "NaN" : startDate.getYear().getValue();
+                if (!PojoUtil.isEmpty(startDate.getMonth()))
+                    dateSortString += "-" + startDate.getMonth().getValue();
+                if (!PojoUtil.isEmpty(startDate.getDay()))
+                    dateSortString += "-" + startDate.getDay().getValue();
+            } else {
+                dateSortString = "X-";
+                dateSortString += endDate.getYear() == null ? "NaN" : endDate.getYear().getValue();
+                if (!PojoUtil.isEmpty(endDate.getMonth()))
```
wjrsimpson: We should use 00 is there is no value for month or day, so that the sort string is always the same length.

amontenegro: @wjrsimpson Fixed

PR #4255, opened 2017-12-11
added : and () to the DOI regex

TomDemeranville:


TomDemeranville:
[Fix DOI normalizer to work with DOIs containing colons and brackets : ( )](https://trello.com/c/0iu3FWNx/4419-fix-doi-normalizer-to-work-with-dois-containing-colons-and-brackets)

PR #4256, opened 2017-12-11
Null work titles no longer break paginated works in UI

bobcaprice:


bobcaprice:
[Allow for null work titles](https://trello.com/c/mznOieav/4420-allow-for-null-work-titles)

PR #4266, opened 2017-12-15
fix id icon alignment in r to l layout

lizkrznarich:


lizkrznarich:
[right to left: id icon is not aligned with the ORCID iD](https://trello.com/c/F9r6mhcv/4427-right-to-left-id-icon-is-not-aligned-with-the-orcid-id)

PR #4269, opened 2017-12-18
Synchronization on12 add works

amontenegro:


amontenegro:
Hi @rcpeters @wjrsimpson,

This is a simple sync routine to allow just one thread at a time to add works to the same record.

Thanks

PR #4272, opened 2017-12-19
Bulk works allow null titles

bobcaprice:


bobcaprice:
[Bulk works end point lets you add a work without a title](https://trello.com/c/3uRiGyfx/4423-bulk-works-end-point-lets-you-add-a-work-without-a-title)

PR #4273, opened 2017-12-19
Migrate self service to ng2

wjrsimpson:


wjrsimpson:
[Migrate consortia self-service to Ng2](https://trello.com/c/k7SG83xM/4443-migrate-consortia-self-service-to-ng2)

PR #4274, opened 2017-12-20
Set version in assets path

amontenegro:


amontenegro:
[Move version param to path for cache busting](https://trello.com/c/4LmZFajG/4421-move-version-param-to-path-for-cache-busting)

amontenegro:
Ok, let me try removing the staticContentFilter, I will let you know if it works without it.

Thanks

wjrsimpson:
Discussed with Angel, who investigated and found that the filter returns a better error message in the response body, so we are keeping it.

PR #4275, opened 2017-12-20
fix broken prefill registration form from oauth params

lizkrznarich:


lizkrznarich:
[Registration doesn't prefill with name and email params on OAuth URL](https://trello.com/c/hM8Ej2qC/4448-registration-doesnt-prefill-with-name-and-email-params-on-oauth-url)

PR #4276, opened 2017-12-21
show login logic: existing account takes precedence over show_login p…

lizkrznarich:
…aram


lizkrznarich:
[Registration doesn't prefill with name and email params on OAuth URL](https://trello.com/c/hM8Ej2qC/4448-registration-doesnt-prefill-with-name-and-email-params-on-oauth-url)

PR #4278, opened 2017-12-22
Remove markup from properties

bobcaprice:


bobcaprice:
[Remove markup from i18n properties files](https://trello.com/c/QDO17j7e/4449-remove-markup-from-i18n-properties-files)

PR #4279, opened 2017-12-22
Copied prompt=none logic from OauthLoginContoller, so that works with 2 screens OAuth.

wjrsimpson:


wjrsimpson:
@TomDemeranville Please can you check this change for me. OpenId was not working correctly for prompt=none. I noticed because OpenIdConnectTest.testPromptNone() was failing (once I'd worked out it was a real failure!).

wjrsimpson:
[OpenId not working correctly with prompt=none](https://trello.com/c/IdL9qlxa/4457-openid-not-working-correctly-with-promptnone)

PR #4286, opened 2018-01-02
Fixed broken links

bobcaprice:


bobcaprice:
[Link to OAuth2 on Dev Tools set up does not work](https://trello.com/c/tZCevs8Q/4464-link-to-oauth2-on-dev-tools-set-up-does-not-work)

bobcaprice:
[Both links to Public API Terms of service presented when setting up Dev Tools do not work](https://trello.com/c/or7J1V4p/4465-both-links-to-public-api-terms-of-service-presented-when-setting-up-dev-tools-do-not-work)

PR #4287, opened 2018-01-02
Corrected error message logic following i18n property markup updates

bobcaprice:


bobcaprice:
[No warning displays when trying to add and email to an account that already exists on another record.](https://trello.com/c/kbcX1IfS/4462-no-warning-displays-when-trying-to-add-and-email-to-an-account-that-already-exists-on-another-record)

PR #4291, opened 2018-01-04
Changed so that org name is retrieved when searching by website.

wjrsimpson:


wjrsimpson:
[Consortia self-service: Existing org modal is skipped sometimes](https://trello.com/c/NbsBdUHv/4450-consortia-self-service-existing-org-modal-is-skipped-sometimes)

PR #4292, opened 2018-01-04
ensured OpenID Connect prompt=none results in error=interaction_requi…

TomDemeranville:
…red if the user is logged in but has not granted permissions.


TomDemeranville:
[OpenId not working correctly with prompt=none](https://trello.com/c/IdL9qlxa/4457-openid-not-working-correctly-with-promptnone)

PR #4295, opened 2018-01-04
Corrected visibility of work when switching between two of different vis

bobcaprice:


bobcaprice:
[Works when sources are open have the privacy selector display the privacy of the source you are looking at](https://trello.com/c/a092RTkg/4442-works-when-sources-are-open-have-the-privacy-selector-display-the-privacy-of-the-source-you-are-looking-at)

PR #4296, opened 2018-01-04
Zero works header no longer shown on public profile page

bobcaprice:


bobcaprice:
[Don't show the works section on a public record when there are no works](https://trello.com/c/BJz3xCoB/4467-dont-show-the-works-section-on-a-public-record-when-there-are-no-works)

PR #4298, opened 2018-01-05
fix so that implicit tokens work with the persistence flag & prompt none

TomDemeranville:
Fixes another bug whereby we were checking expiry of tokens instead of persistence flag to see if user had valid permissions for a given client.  THIS IS A CHANGE IN BEHAVIOUR.

Add if clause in OrcidTokenEnhancerisPersistentTokenEnabled to deal with case where request_type contains "token" and grantPersistentToken = true.
Modify OrcidRandomValueTokenServicesImpl.longLifeTokenExist to not care if token is expired. Made it use persistent flag instead.
Move check for prompt=none to after we check to see if a token already exists in OauthAuthorizeController.loginGetHandler

### DIFF
--- a/orcid-core/src/main/java/org/orcid/core/oauth/service/OrcidRandomValueTokenServicesImpl.java
+++ b/orcid-core/src/main/java/org/orcid/core/oauth/service/OrcidRandomValueTokenServicesImpl.java
@@ -280,7 +280,7 @@ public class OrcidRandomValueTokenServicesImpl extends DefaultTokenServices impl
         }
         
         for(OAuth2AccessToken token : existingTokens) {
-            if(!token.isExpired() && token.getExpiration().after(java.sql.Timestamp.valueOf(LocalDateTime.now().plusHours(1)))) {
+            if (token.getAdditionalInformation().get("persistent") != null && Boolean.valueOf((token.getAdditionalInformation().get("persistent").toString()))){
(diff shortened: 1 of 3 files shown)


TomDemeranville:
[OpenID blackbox failure](https://trello.com/c/egzbIWnu/4477-openid-blackbox-failure)

orcid-core/src/main/java/org/orcid/core/oauth/service/OrcidRandomValueTokenServicesImpl.java
```diff
@@ -280,7 +280,7 @@ public boolean longLifeTokenExist(String clientId, String userId, Collection<Str
         }
         
         for(OAuth2AccessToken token : existingTokens) {
-            if(!token.isExpired() && token.getExpiration().after(java.sql.Timestamp.valueOf(LocalDateTime.now().plusHours(1)))) {
+            if (token.getAdditionalInformation().get("persistent") != null && Boolean.valueOf((token.getAdditionalInformation().get("persistent").toString()))){
```
amontenegro: Hey @TomDemeranville, could you please use the constant OrcidOauth2Constants.PERSISTENT instead of "persistent"? So, it will be easier to maintain. 

Thanks

TomDemeranville:
done!

Tom Demeranville
Technology Advocate
ORCID Inc
https://orcid.org/0000-0003-0902-4386

PR #4300, opened 2018-01-08
Fix link that loads assets for institutional sign in

amontenegro:


amontenegro:
[Institutional Signin does not load](https://trello.com/c/uGR7oYGZ/4478-institutional-signin-does-not-load)

PR #4304, opened 2018-01-11
Moved schema validation for bulk works back into delegators from mana…

bobcaprice:
…gers


bobcaprice:
[Add peer review v3 blackbox failure](https://trello.com/c/f3sEr17t/4482-add-peer-review-v3-blackbox-failure)

PR #4305, opened 2018-01-11
allow x-csrf-token header on requests from cors allowed domains

lizkrznarich:


lizkrznarich:
[Migrate SearchCtrl to Ng2](https://trello.com/c/vbaB5mIk/4441-migrate-searchctrl-to-ng2)

PR #4306, opened 2018-01-12
Fixed swagger so that it uses release name in path to static resources.

wjrsimpson:
Changed release name utils, so that release name is the same across all web apps.


wjrsimpson:
[Swagger UI seems to be broken.  https://support.orcid.org/admin/tickets/92310](https://trello.com/c/eXa9XJTs/4494-swagger-ui-seems-to-be-broken-https-supportorcidorg-admin-tickets-92310)

PR #4307, opened 2018-01-12
initialize reg form with null email additional and don't remove it du…

lizkrznarich:
…ring validation


lizkrznarich:
[Multi email on registration: trying to register without an email removes the additional email box](https://trello.com/c/EcE7llSB/4491-multi-email-on-registration-trying-to-register-without-an-email-removes-the-additional-email-box)

Issue #4316, opened 2018-01-22
Support ActivityPub outbox declaration in profile


csarven:
W3C [ActivityPub](https://www.w3.org/TR/activitypub/) (AP) species a protocol to deliver notifications and content for federated servers. Its [outbox](https://www.w3.org/TR/activitypub/#outbox) mechanism is intended to provide a collection of a profile's activities to its consumers.

AP's outbox is similar to [Linked Data Notifications](https://www.w3.org/TR/ldn/) inbox in that they both have to do with outgoing and incoming information that's relevant for the user. Hence, this issue is about providing the possibility for profiles to simply declare the URL of their outbox. No additional mechanism needs to be implemented.

Aside: This issue has the same architectural requirements as https://github.com/ORCID/ORCID-Source/issues/3416 and the corresponding PR https://github.com/ORCID/ORCID-Source/pull/3847 ie. users can optionally add the term "outbox" and its URL under "Edit websites" in the UI.

csarven:
PR merged and available in production. All good. Closing.

ashwinvis:
@csarven Does this allow for ActivityPub compatible services (mastodon, pleroma, gnu social etc.) to "follow"  orcid profiles?

csarven:
The ORCID server is not an AP implementation, so following will not work. However, a small part of the ORCID profile can be described like an AS2 Actor provided that it points to its outbox or inbox. See for example `curl -LH'Accept: text/turtle' https://orcid.org/0000-0002-0880-9125`. Putting content negotiation aside, an AP client can discover and interact with that eg. send or consume outbox activities, send or consume notifications from an inbox.

Aside: I think the `curl -LH'Accept: application/ld+json' https://orcid.org/0000-0002-0880-9125` is returning all information and correctly. Also from the AP perspective, JSON-LD would be required.

ashwinvis:
Thanks for the clarification. `ld+json` looks promising and in practice one can receive updates on research. I just tried mine:
```sh
curl -LH'Accept: application/ld+json' https://orcid.org/0000-0002-2979-6327
```

and the research "works" are available, which is what matters the most. It is also super nice that inbox and outbox supports interaction! Hopefully orcid can provide equivalent functionality to the likes of researchgate / academia.

csarven:
The JSON-LD representation has bugs: https://github.com/ORCID/ORCID-Source/issues/5452 .

[...]

PR #4317, opened 2018-01-22
Extending the profile description with personal outbox

csarven:
See issue https://github.com/ORCID/ORCID-Source/issues/4316 for description.


stain:
I think this is a simple addition and should go in. 

However I also wish we could document this somewhere, providing links to these upstream specifications and for what use cases they are appropriate.

Shall we just make a page in the [wiki](https://github.com/ORCID/ORCID-Source/wiki)? Their pages say "edited this page on Jun 20, 2013" ..

csarven:
+1 to documentation. I think there should be a separate issue/PR that sweeps through everything that is not documented.

PR #4320, opened 2018-01-24
MQ - Performance improvements on12 processing

amontenegro:


amontenegro:
[Message listener: Modify how we feed the 1.2 buckets](https://trello.com/c/jnTjvwUC/4466-message-listener-modify-how-we-feed-the-12-buckets)

PR #4321, opened 2018-01-25
fixed jackson mapping exceptions to throw 400 not 500

TomDemeranville:


TomDemeranville:
[Posting items with invalid json returns a 9008 500 server error, it should be a 9001 400 bad request error.](https://trello.com/c/BtDNQROo/4513-posting-items-with-invalid-json-returns-a-9008-500-server-error-it-should-be-a-9001-400-bad-request-error)

PR #4326, opened 2018-01-26
Create a read public access token for BB tests

amontenegro:


amontenegro:
[Include token during SetupClientsAndUsers used by for message-listener.properties](https://trello.com/c/p8wwJsET/4418-include-token-during-setupclientsandusers-used-by-for-message-listenerproperties)

PR #4328, opened 2018-01-29
added [] {} <> ; to list of valid DOI characters

TomDemeranville:


TomDemeranville:
[DOI grouping ignoring text in carrots](https://trello.com/c/JEjympql/4531-doi-grouping-ignoring-text-in-carrots)

PR #4330, opened 2018-01-30
added link to identifiers page from XSD

TomDemeranville:


TomDemeranville:
[Add link to list of identifiers to the external-identifier field documentation in the XSD](https://trello.com/c/O5gqyRHK/4534-add-link-to-list-of-identifiers-to-the-external-identifier-field-documentation-in-the-xsd)

PR #4331, opened 2018-01-30
Log invalid or null csrf token

amontenegro:


amontenegro:
Hi @wjrsimpson, @rcpeters,

This is the card we talked about early today, Spring catches invalid CSRF errors (which are 403) and return them as 405, so, it is affecting our error metrics.

With this PR, we will log any CSRF error and return the correct 403 error code.

Thanks

PR #4335, opened 2018-02-01
Fix RDF and identifiers

amontenegro:


amontenegro:
[RDF no longer works since switching the default public API version](https://trello.com/c/SB2EKbnw/4546-rdf-no-longer-works-since-switching-the-default-public-api-version)

orcid-pub-web/src/main/java/org/orcid/api/filters/PubApiDefaultVersionFilter.java
```diff
@@ -42,7 +42,7 @@
     protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain) throws ServletException, IOException {
         HttpServletRequest httpRequest = (HttpServletRequest) request;
         String path = httpRequest.getServletPath();
-        if (path.startsWith("/search/") || path.startsWith("/oauth/token")) {
+        if (path.startsWith("/search/") || path.startsWith("/oauth/token") || path.startsWith("/experimental_rdf_v1/") || path.startsWith("/static/identifierTypes.xsl")) {
```
TomDemeranville: Could we have just path.startsWith("/static/") ? The swagger.json is also in that directory.

Hmm.  This will be avoidable if you build path like /static/../whatever

PR #4337, opened 2018-02-02
Default api version filter for members and pub ap is

amontenegro:


amontenegro:
[Update the default members API version](https://trello.com/c/J6geMNxB/4549-update-the-default-members-api-version)

PR #4338, opened 2018-02-02
Website url for public clients now validated in same place as member …

bobcaprice:
…clients websites


bobcaprice:
[public API redirects: use same logic to validate redirects uris as is currently used for websites](https://trello.com/c/okMPKWNA/4536-public-api-redirects-use-same-logic-to-validate-redirects-uris-as-is-currently-used-for-websites)

PR #4339, opened 2018-02-05
Added deprecation method and deprecating admin to profile table

bobcaprice:


bobcaprice:
[Log deprecated method.](https://trello.com/c/7xfUw9Ys/4501-log-deprecated-method)

PR #4343, opened 2018-02-06
Fix password reset error msg

lizkrznarich:


lizkrznarich:
[Reset password sometimes shows error and success message at same time](https://trello.com/c/85OD9O51/4542-reset-password-sometimes-shows-error-and-success-message-at-same-time)

PR #4349, opened 2018-02-08
fix ng error that happens when blog feed hasn't finished loading yet

lizkrznarich:


lizkrznarich:
[Console errors on orcid.org](https://trello.com/c/p0YD658i/4565-console-errors-on-orcidorg)

PR #4351, opened 2018-02-09
update older browser alert msg

lizkrznarich:


lizkrznarich:
[Change 'older browser' alert message](https://trello.com/c/GvOs9RGk/4496-change-older-browser-alert-message)

PR #4353, opened 2018-02-09
Fix error where new aff types appears as Emplyments in the workspace

amontenegro:


amontenegro:
[New aff types appears as Employments in the workspace section](https://trello.com/c/zY3PzVKf/4584-new-aff-types-appears-as-employments-in-the-workspace-section)

PR #4355, opened 2018-02-09
Fixed self service org id search to use correct query.

wjrsimpson:


wjrsimpson:
[Add org ids to self-service](https://trello.com/c/Q0G2WmfP/4560-add-org-ids-to-self-service)

PR #4360, opened 2018-02-13
Script to go through access log files and collect basic statistics to…

bobcaprice:
… compare with google analytics, requires further testing


bobcaprice:
[Compare access logs and / or web logs for API hits with GA data to see if total event count is correct](https://trello.com/c/mWpS2bSI/4425-compare-access-logs-and-or-web-logs-for-api-hits-with-ga-data-to-see-if-total-event-count-is-correct)

PR #4363, opened 2018-02-15
Moved QR code generation into registry

bobcaprice:


bobcaprice:
[Move QR code generator into Registry](https://trello.com/c/6xLox3fQ/4550-move-qr-code-generator-into-registry)

PR #4365, opened 2018-02-15
When the email doesnt exists or it belongs to other record, return an…

amontenegro:
… error


amontenegro:
[Can't delete email addresses with a + sign](https://trello.com/c/mF5KvxIi/4599-cant-delete-email-addresses-with-a-sign)

PR #4371, opened 2018-02-19
Added server side validation of ajax dates

bobcaprice:


bobcaprice:
[Invalid end dates allowed for affiliations in UI](https://trello.com/c/UxE8Uu27/4544-invalid-end-dates-allowed-for-affiliations-in-ui)

PR #4385, opened 2018-02-23
Corrected works count method as only used by public profile page

bobcaprice:


bobcaprice:
[No 'no public info' message shown if user has no data except limited / private vis works](https://trello.com/c/dVYFgUfO/4615-no-no-public-info-message-shown-if-user-has-no-data-except-limited-private-vis-works)

PR #4389, opened 2018-02-27
We dont need a second query, we can update the password in the same q…

amontenegro:
…uery that claims the record


amontenegro:
[After claiming a new record you're taken to sign-in instead of to the record](https://trello.com/c/HZGLVjOP/4611-after-claiming-a-new-record-youre-taken-to-sign-in-instead-of-to-the-record)

PR #4395, opened 2018-02-28
Dont update the last modified on user login

amontenegro:


amontenegro:
[Add last_login to the profile table](https://trello.com/c/wZyBitJq/4624-add-lastlogin-to-the-profile-table)

PR #4397, opened 2018-03-01
Truncate org name, because there is a limit of 41 characters in SalesForce

wjrsimpson:


wjrsimpson:
[Consortia self-service: Limit org name to 41 chars](https://trello.com/c/DGBkFIEw/4629-consortia-self-service-limit-org-name-to-41-chars)

PR #4399, opened 2018-03-02
Resolve pids

TomDemeranville:
Adds a ResolverService, which checks if a PID resolves to something (a http 200 in most cases)

Not yet being used by anything.


TomDemeranville:
[ID resolution checker.](https://trello.com/c/3TLHn7ol/4623-id-resolution-checker)

PR #4405, opened 2018-03-06
Hide empty sections in the public page

amontenegro:


amontenegro:
[Affiliation sections don't display on a record if they are empty](https://trello.com/c/RxV4xCN8/4640-affiliation-sections-dont-display-on-a-record-if-they-are-empty)

PR #4412, opened 2018-03-08
added ability to html unescape DOI strings recursively

TomDemeranville:
until no more escaping to do


TomDemeranville:
[DOIs with "&lt;" and "&gt;" incorrectly group.](https://trello.com/c/umrMrJQw/4655-dois-with-lt-and-gt-incorrectly-group)

PR #4413, opened 2018-03-08
Group works backend

bobcaprice:


bobcaprice:
[2.a Make it easier for end users to group items in the record.](https://trello.com/c/mqtkPacU/4636-2a-make-it-easier-for-end-users-to-group-items-in-the-record)

PR #4417, opened 2018-03-09
IE don't show empty brackets if no affiliations in search result

lizkrznarich:


lizkrznarich:
[Affiliation in search results show empty brackets if there are no results on IE 11](https://trello.com/c/pkkhWT5C/4554-affiliation-in-search-results-show-empty-brackets-if-there-are-no-results-on-ie-11)

PR #4419, opened 2018-03-09
change container height to allow space for footer

lizkrznarich:


lizkrznarich:
[Update footer layout](https://trello.com/c/D9V4sFZh/3464-update-footer-layout)

PR #4425, opened 2018-03-13
update unsupported browsers list

lizkrznarich:


lizkrznarich:
[Update list of browsers which trigger the unsupported browser message](https://trello.com/c/aKD9AGwa/4618-update-list-of-browsers-which-trigger-the-unsupported-browser-message)

PR #4429, opened 2018-03-15
Fixed

amontenegro:


amontenegro:
[Helper text when the user has not added a distinction or invited position is out of order](https://trello.com/c/LgesCZI5/4663-helper-text-when-the-user-has-not-added-a-distinction-or-invited-position-is-out-of-order)

PR #4430, opened 2018-03-16
Fix update default visibility

lizkrznarich:


lizkrznarich:
[Account Settings visibility selector does not work](https://trello.com/c/PVnlKRGf/4689-account-settings-visibility-selector-does-not-work)

lizkrznarich:
[Both old and new visibility selector show on account settings when GDPR and ng2 are enabled](https://trello.com/c/PxrqYZpN/4680-both-old-and-new-visibility-selector-show-on-account-settings-when-gdpr-and-ng2-are-enabled)

PR #4432, opened 2018-03-16
fix default privacy and email frequency when ng2 enabled and gdpr ui …

lizkrznarich:
…disabled


lizkrznarich:
[ng2 register can't select default privacy & notifications are set to immediately if GDPR is not enabled](https://trello.com/c/Bl6ZvScY/4677-ng2-register-cant-select-default-privacy-notifications-are-set-to-immediately-if-gdpr-is-not-enabled)

PR #4437, opened 2018-03-19
include full url in more delegates id display

lizkrznarich:


lizkrznarich:
[The more delegates page links to ORCID iDs without https](https://trello.com/c/QKRKfkLK/4376-the-more-delegates-page-links-to-orcid-ids-without-https)

PR #4442, opened 2018-03-20
Ng2 migration.v.0.4

mjcalvo21:


mjcalvo21:
[Sign in doesn't load on IE 11](https://trello.com/c/EZy6rc5p/4696-sign-in-doesnt-load-on-ie-11)

PR #4446, opened 2018-03-20
Catch NoResultException so it doesn't throw a 500 error

amontenegro:


amontenegro:
[Refactor reset password so it doesn't return 500 when an email doesn't exists](https://trello.com/c/chdIXEuk/4684-refactor-reset-password-so-it-doesnt-return-500-when-an-email-doesnt-exists)

PR #4447, opened 2018-03-20
Limit the max number of works per record to 10,000

amontenegro:


amontenegro:
[Prevent records to have more than 10,000 works](https://trello.com/c/eYOvzywT/4701-prevent-records-to-have-more-than-10000-works)

PR #4454, opened 2018-03-22
Fix date validation issues on affiliations

amontenegro:


amontenegro:
[New affiliations adding an item without a start date won't save](https://trello.com/c/FATxNgeG/4670-new-affiliations-adding-an-item-without-a-start-date-wont-save)

PR #4465, opened 2018-03-27
Moving model attributes to the BaseWorkspaceController so the public …

amontenegro:
…page can access them


amontenegro:
[Public records do not load](https://trello.com/c/OiBuQTv0/4712-public-records-do-not-load)

PR #4466, opened 2018-03-27
Send the claim reminder email

amontenegro:


amontenegro:
[Resend claim page sending original claim email instead of reminder to claim email](https://trello.com/c/g3ogjnnV/4702-resend-claim-page-sending-original-claim-email-instead-of-reminder-to-claim-email)

PR #4467, opened 2018-03-27
Json value annotation on enums

amontenegro:


amontenegro:
[JSON switched to expecting lower case values](https://trello.com/c/4HjbXG6Y/4720-json-switched-to-expecting-lower-case-values)

PR #4468, opened 2018-03-28
added arxiv normalizer and url resolver

TomDemeranville:
Adds a normalizer and URL resolver for arxiv, using info found here: https://arxiv.org/help/arxiv_identifier_for_services


TomDemeranville:
[Normalize arxiv identifiers](https://trello.com/c/Qax8BBfo/4709-normalize-arxiv-identifiers)

PR #4469, opened 2018-03-28
Fix bug

amontenegro:


amontenegro:
[Email and keywords modals do not load content](https://trello.com/c/azFQNvyz/4722-email-and-keywords-modals-do-not-load-content)

PR #4471, opened 2018-03-28
When opening an oauth2Authorize tab, save the authorization endpoint …

amontenegro:
…in the session, so, you can pick it in case the saved request get lost


amontenegro:

PR #4475, opened 2018-03-29
remove ng1 search and use ng2

lizkrznarich:


lizkrznarich:
[Affiliation in search results show empty brackets if there are no results on IE 11](https://trello.com/c/pkkhWT5C/4554-affiliation-in-search-results-show-empty-brackets-if-there-are-no-results-on-ie-11)

PR #4477, opened 2018-04-02
If two tabs, one with oauth login and other with login, and the user …

amontenegro:
…login, then the first tab redirects to oauth authorize page


amontenegro:

PR #4480, opened 2018-04-03
added separate DOI resolver that looks for metadata

TomDemeranville:
(to cope with services that require cookies!)


TomDemeranville:

PR #4482, opened 2018-04-03
modified user info to use configured baseUri for issuer

TomDemeranville:
Modified id_token to use configured baseUri as issuer


TomDemeranville:
[Change issuer on sandbox to be the sandbox domain.](https://trello.com/c/dPoyiKW7/4734-change-issuer-on-sandbox-to-be-the-sandbox-domain)

PR #4486, opened 2018-04-04
Fix force sign out case

amontenegro:


amontenegro:

PR #4490, opened 2018-04-06
Rolled back all oauth urls requiring https

bobcaprice:


bobcaprice:
[OAuth will not load- redirect error](https://trello.com/c/6naQEbgg/4743-oauth-will-not-load-redirect-error)

PR #4497, opened 2018-04-10
Dont index unclaimed records

amontenegro:


amontenegro:
[Research active mq for new message listener implementation](https://trello.com/c/LDkSdKjV/4593-research-active-mq-for-new-message-listener-implementation)

PR #4504, opened 2018-04-13
Hard coded concurrency of topic listeners to a single thread

bobcaprice:


bobcaprice:
[Multiple threads consuming same jms topic message](https://trello.com/c/i9QTQUw1/4766-multiple-threads-consuming-same-jms-topic-message)

PR #4506, opened 2018-04-16
Id resolver ui

TomDemeranville:
basic existence resolution support for: 
DOI
OCLC
ISBN
ARXIV
PMID
PMC
RRID
RFC
PDB

Other types will generate URLs if a prefix exists, using a normalized identifier if a normalizer exists.


TomDemeranville:

PR #4508, opened 2018-04-16
Disable v1.1 from the members and public api

amontenegro:


amontenegro:
[Disable v1.1 of the API](https://trello.com/c/BDmTOSD0/4746-disable-v11-of-the-api)

amontenegro:
Need to add togglz to it

PR #4509, opened 2018-04-16
Prevent clients to try to login

amontenegro:


amontenegro:
[Prevent client_ids be used to login](https://trello.com/c/40Hrznro/4739-prevent-clientids-be-used-to-login)

PR #4513, opened 2018-04-17
Create a thread pool on every listener, so, we can handle how many th…

amontenegro:
…reads are processed at the same time


amontenegro:
[Thread pool to control how many threads runs at the same time in the message listener](https://trello.com/c/WmVnp1sk/4769-thread-pool-to-control-how-many-threads-runs-at-the-same-time-in-the-message-listener)

amontenegro:
Hi @TomDemeranville, could you please take a look?

Thanks

PR #4524, opened 2018-04-20
On the DB we are storing the Visibility.name not the Visibility.value

amontenegro:


amontenegro:
[Setting an affiliation to private or limited causes all affiliations to fail to load](https://trello.com/c/B6odhIbA/4776-setting-an-affiliation-to-private-or-limited-causes-all-affiliations-to-fail-to-load)

PR #4525, opened 2018-04-20
Fix error, string.equals(enum)

amontenegro:


amontenegro:
[Admin switch user does not work](https://trello.com/c/RCM0k9T5/4780-admin-switch-user-does-not-work)

PR #4530, opened 2018-04-24
Fix worksPaginator filter, so, for the public page, it filters the wo…

amontenegro:
…rks based on the element visibility and not on the primary work group visibility


amontenegro:
[Groups with works that include different visibilities all works take the visibility of the primary work in the UI](https://trello.com/c/JBIs7vFX/4789-groups-with-works-that-include-different-visibilities-all-works-take-the-visibility-of-the-primary-work-in-the-ui)

PR #4539, opened 2018-04-26
added schema.org to displayMyID

TomDemeranville:


TomDemeranville:
[Add semantic markup to "Display my iD"](https://trello.com/c/OWWbYa2k/4238-add-semantic-markup-to-display-my-id)

PR #4540, opened 2018-04-26
DisambiguatedOrgs now validated in v2 API when supplied

bobcaprice:


wjrsimpson:
[Use org disabiguated ID when supplied in the v2 API](https://trello.com/c/Os9x55ri/4920-use-org-disabiguated-id-when-supplied-in-the-v2-api)

PR #4549, opened 2018-05-03
ng5: removing qa tags for some components

mjcalvo21:


mjcalvo21:
[Header](https://trello.com/c/P18zfcRy/4824-header)

PR #4556, opened 2018-05-03
make footer full width

lizkrznarich:


lizkrznarich:
[GDPR foot isn't full width of window](https://trello.com/c/D0MzxydX/4856-gdpr-foot-isnt-full-width-of-window)

PR #4559, opened 2018-05-04
default_visibility return json

lizkrznarich:


lizkrznarich:
[Unable to update visibility preferences error](https://trello.com/c/S17j487R/4863-unable-to-update-visibility-preferences-error)

PR #4560, opened 2018-05-04
show bulk visibility edit on also known as form

lizkrznarich:


lizkrznarich:
[Other names is missing bulk edit](https://trello.com/c/rBmeUyK7/4859-other-names-is-missing-bulk-edit)

PR #4561, opened 2018-05-04
Delete country issues

lizkrznarich:


lizkrznarich:
[Country: delete not working right](https://trello.com/c/9IRvplo4/4861-country-delete-not-working-right)

PR #4568, opened 2018-05-07
Corrected biography visibility to use defaultVis.name() rather than (…

bobcaprice:
…lower case) value()


bobcaprice:
[Biography visibility set to lower case value](https://trello.com/c/oAKhTB5N/4879-biography-visibility-set-to-lower-case-value)

PR #4578, opened 2018-05-08
4875 inbox header is missing count of number of unread notifications

mjcalvo21:


mjcalvo21:
[Inbox header is missing count of number of unread notifications](https://trello.com/c/QnSE0DKw/4875-inbox-header-is-missing-count-of-number-of-unread-notifications)

PR #4581, opened 2018-05-10
adding auto property for left, top, right, bottom and margin because …

mjcalvo21:
…ie11 doesnt recognize initial value


mjcalvo21:
[IE 11 personal info visibility pop-up blocks changing visibility](https://trello.com/c/9eyGboTu/4886-ie-11-personal-info-visibility-pop-up-blocks-changing-visibility)

PR #4584, opened 2018-05-10
languages: setting english as the default language instead of first i…

mjcalvo21:
…tem in array

https://trello.com/c/svgKH7UY/4890-language-selector-show-arabic-as-default-choice-rather-than-the-language-the-page-is-displayed-in


mjcalvo21:
[Language selector show Arabic as default choice rather than the language the page is displayed in](https://trello.com/c/svgKH7UY/4890-language-selector-show-arabic-as-default-choice-rather-than-the-language-the-page-is-displayed-in)

PR #4593, opened 2018-05-11
update GDPR deactivate text to reference all email addresses

lizkrznarich:


lizkrznarich:
[GDPR Deactivate: Says we only keep primary email address but we keep all email addresses](https://trello.com/c/yoCyaKwA/4904-gdpr-deactivate-says-we-only-keep-primary-email-address-but-we-keep-all-email-addresses)

PR #4602, opened 2018-05-14
Grettings and footer should not be Translated

amontenegro:


amontenegro:
[Service Announcement email: greeting and footer are translated](https://trello.com/c/NbGNK8wh/4913-service-announcement-email-greeting-and-footer-are-translated)

PR #4606, opened 2018-05-14
GDPR Email frequency on claim page

amontenegro:


amontenegro:
Hey @lizkrznarich, it looks like the ng2 claim page is not up to date, it doesnt include the new visibility settings. Could you please review  this (which only modifies claim.ftl) and see if we should also update the claim ng2 template as well?

Thanks

PR #4616, opened 2018-05-16
4922 cant select day with affiliations

mjcalvo21:


mjcalvo21:
[Can't select day with affiliations](https://trello.com/c/iadw42VQ/4922-cant-select-day-with-affiliations)

PR #4617, opened 2018-05-17
Added deprecated orcid param to OrcidDeprecatedException

bobcaprice:


bobcaprice:
[3.0 reading a locked or deprecated record returns the record instead of an error](https://trello.com/c/AQqpQ756/4895-30-reading-a-locked-or-deprecated-record-returns-the-record-instead-of-an-error)

PR #4618, opened 2018-05-17
Changed url of qr code so doesn't pull my-orcid templates in

bobcaprice:


bobcaprice:
[my QR code: nothing loads](https://trello.com/c/MMGjvrhT/4924-my-qr-code-nothing-loads)

PR #4619, opened 2018-05-17
Affiliations v3 sort

lizkrznarich:


lizkrznarich:
[V3 Affiliations: Sort within section instead of all affiliations](https://trello.com/c/Jw9KNH9d/4794-v3-affiliations-sort-within-section-instead-of-all-affiliations)

PR #4625, opened 2018-05-18
blog feed configure http client request to receive XML

lizkrznarich:


lizkrznarich:
[Blog feed on home page is broken](https://trello.com/c/XZKHHT9R/4936-blog-feed-on-home-page-is-broken)

PR #4627, opened 2018-05-18
Orgs now matched on address plus disambiguated org id in v2 to bring …

bobcaprice:
…in line with v3


bobcaprice:
[2.0+ Always user org identifier provided with the affiliation](https://trello.com/c/E1yge8Jt/4932-20-always-user-org-identifier-provided-with-the-affiliation)

PR #4628, opened 2018-05-18
Index new records

amontenegro:


amontenegro:
[New records aren't being indexed (or are being indexed slowly)](https://trello.com/c/nL2Pjf9C/4939-new-records-arent-being-indexed-or-are-being-indexed-slowly)

PR #4629, opened 2018-05-18
Ng2 header move search logic to angular

lizkrznarich:


lizkrznarich:
[Search the ORCID website is broken](https://trello.com/c/sEOORYjf/4935-search-the-orcid-website-is-broken)

PR #4642, opened 2018-05-28
Check for null source name content before creating source name object

bobcaprice:


bobcaprice:
[3.0 error causing distinction section to be cut off in API](https://trello.com/c/3fQvqttu/4943-30-error-causing-distinction-section-to-be-cut-off-in-api)

bobcaprice:
[Api analysis script bug - if final item in array of log file names does not exist script falls over](https://trello.com/c/9he3m44f/4951-api-analysis-script-bug-if-final-item-in-array-of-log-file-names-does-not-exist-script-falls-over)

mjcalvo21:
[/account: Security question](https://trello.com/c/W5cJz6Br/4803-account-security-question)

PR #4644, opened 2018-05-29
Record name utils now returns null where name not public

bobcaprice:


bobcaprice:
[Display name should follow trusted individual users preferences.](https://trello.com/c/sj4PhzCg/4954-display-name-should-follow-trusted-individual-users-preferences)

PR #4647, opened 2018-05-31
made claims_type_supported a boolean instead of a string.

TomDemeranville:


TomDemeranville:
[remove claims_parameter_supported from openid well known config](https://trello.com/c/tWDYzbZb/4942-remove-claimsparametersupported-from-openid-well-known-config)

PR #4650, opened 2018-06-01
some minor improvements to the readability of the manual test doc

dreamofaredbird:


caoyler:
@dreamofaredbird For step 47 (add a funding item) the city and country should populate automatically. Please double check.

dreamofaredbird:
Agreed the city and country auto-populate.

PR #4651, opened 2018-06-01
Dont store the notification if the frequency is set to NEVER

amontenegro:


amontenegro:
[Notification frequencies set to Never are still delivered to the inbox](https://trello.com/c/HHuTAPz1/4946-notification-frequencies-set-to-never-are-still-delivered-to-the-inbox)

PR #4660, opened 2018-06-06
Return 409 with an error

amontenegro:


amontenegro:
[When posting a permission notification return a 409 error message](https://trello.com/c/goq0utfj/4966-when-posting-a-permission-notification-return-a-409-error-message)

PR #4661, opened 2018-06-06
Post social sign in url now set to absolute url including https schema

bobcaprice:


bobcaprice:
[Change destination header to absolute URL for social sign in](https://trello.com/c/jLGbWNPM/4971-change-destination-header-to-absolute-url-for-social-sign-in)

PR #4662, opened 2018-06-06
Custom db connection pools for the internal api so we can reduce the …

amontenegro:
…number of open connections


amontenegro:
[Create custom db connection pool for the internal api](https://trello.com/c/lgtTkU9a/4740-create-custom-db-connection-pool-for-the-internal-api)

PR #4673, opened 2018-06-11
Fix for missing item in header

mjcalvo21:
https://trello.com/c/rwKvWFvO/4978-member-tools-menu-option-missing-on-records-with-self-service-enabled


mjcalvo21:
[Member Tools menu option missing on records with self-service enabled](https://trello.com/c/rwKvWFvO/4978-member-tools-menu-option-missing-on-records-with-self-service-enabled)

PR #4681, opened 2018-06-14
clears research resources when record deactivated/depreciated

TomDemeranville:


TomDemeranville:
[Research-resources are not deleted when a record is deactivated](https://trello.com/c/em8KneTu/4992-research-resources-are-not-deleted-when-a-record-is-deactivated)

PR #4683, opened 2018-06-14
move aka, country, keywords and websites into one module

lizkrznarich:


wjrsimpson:
@lizkrznarich Can you take a look at the conflicts (created when i merged Manuel's upgrade to ng6)?

PR #4687, opened 2018-06-18
fixed path to be relative

TomDemeranville:


TomDemeranville:
[Fix path in activities xml sample](https://trello.com/c/uN2oDZVm/5001-fix-path-in-activities-xml-sample)

PR #4691, opened 2018-06-19
Work group suggestions

bobcaprice:


bobcaprice:
[2.b[IN PROGRESS] Suggestion engine for user driven groupings.](https://trello.com/c/NBAVc0ps/4963-2bin-progress-suggestion-engine-for-user-driven-groupings)

PR #4693, opened 2018-06-19
fixing npm version issue when deploying

mjcalvo21:


mjcalvo21:
[The deployment script fails to find the javascript file when building on QA](https://trello.com/c/UEH5VKXt/5003-the-deployment-script-fails-to-find-the-javascript-file-when-building-on-qa)

PR #4694, opened 2018-06-19
Get submember contacts list

wjrsimpson:


wjrsimpson:
[Add self service functionality for consortium leads to generate list of main and technical contacts for their members. This should be an on demand feature.](https://trello.com/c/QU53gFvh/4765-add-self-service-functionality-for-consortium-leads-to-generate-list-of-main-and-technical-contacts-for-their-members-this-shoul)

PR #4696, opened 2018-06-20
made orcid-model independent from orcid-parent (which now has it as a…

TomDemeranville:
… dependency)

rationalised orcid-model dependencies


TomDemeranville:
[Create stand alone XML-JSON converter.](https://trello.com/c/lSap8hxU/4985-create-stand-alone-xml-json-converter)

PR #4699, opened 2018-06-22
Changed so the call to geonames is retried once if it fails.

wjrsimpson:


wjrsimpson:
[Fundref loader: Retry geonames call](https://trello.com/c/uZfkBJpm/5002-fundref-loader-retry-geonames-call)

PR #4700, opened 2018-06-22
Added profile checks to notification endpoints

bobcaprice:


bobcaprice:
[Return a 404 error with invalid ORCID iDs with the /notification-permission endpoint](https://trello.com/c/aZo9AtJ0/4999-return-a-404-error-with-invalid-orcid-ids-with-the-notification-permission-endpoint)

PR #4703, opened 2018-06-25
Added client ID as a request attribute so that it can be logged by tomcat.

wjrsimpson:


wjrsimpson:
[Add client ID to member and pub API tomcat access logs](https://trello.com/c/Q8oVlW87/5028-add-client-id-to-member-and-pub-api-tomcat-access-logs)

PR #4708, opened 2018-06-26
Fix the KB link that talks about the notifications preferences

amontenegro:


amontenegro:
[Dead KB article link in the disclaimer of the digest email](https://trello.com/c/zHMFfmJ6/5014-dead-kb-article-link-in-the-disclaimer-of-the-digest-email)

PR #4710, opened 2018-06-27
Added IF EXISTS clause to sql statement to see if fixes ci server

bobcaprice:


bobcaprice:
Doesn't work, HSQL doesn't seem to support IF EXISTS

PR #4715, opened 2018-06-28
fixing setting page

mjcalvo21:


mjcalvo21:
[Account Settings doesn't fully load](https://trello.com/c/JWzz3mF1/5016-account-settings-doesnt-fully-load)

PR #4718, opened 2018-06-28
Limit 1.2 profile cache size in megabytes.

wjrsimpson:


wjrsimpson:
[Limit 1.2 profile cache size in megabytes](https://trello.com/c/UHYICDH7/5021-limit-12-profile-cache-size-in-megabytes)

PR #4719, opened 2018-06-28
added short docs to core

TomDemeranville:


TomDemeranville:
[Document how to add a new activity type](https://trello.com/c/CMw6vUib/4993-document-how-to-add-a-new-activity-type)

PR #4723, opened 2018-06-29
removed slfj dependency

TomDemeranville:


TomDemeranville:
[Create stand alone XML-JSON converter.](https://trello.com/c/lSap8hxU/4985-create-stand-alone-xml-json-converter)

PR #4727, opened 2018-07-03
Error message changed for logged out users where client id invalid

bobcaprice:


bobcaprice:
[Different OAuth error message if you are logged in or not](https://trello.com/c/IIvSvkec/5023-different-oauth-error-message-if-you-are-logged-in-or-not)

PR #4730, opened 2018-07-04
arxiv only extracted from bibtex if eprinttype=arxiv

TomDemeranville:
Example of this here: https://tex.stackexchange.com/questions/49757/what-should-bibtex-for-arxiv-entries-look-like


TomDemeranville:
[BibTeX Importer: ignore eprint field (currently being saved as ARXIV identifier)](https://trello.com/c/hb4aw3ZR/5032-bibtex-importer-ignore-eprint-field-currently-being-saved-as-arxiv-identifier)

PR #4746, opened 2018-07-10
Fix ng2 deprecate

lizkrznarich:


lizkrznarich:
[/account: Remove duplicate record](https://trello.com/c/Dh7BsjFE/4805-account-remove-duplicate-record)

PR #4751, opened 2018-07-12
Allow email frequency to be null (to allow for unclaimed records).

wjrsimpson:


wjrsimpson:
[Using the API to create a record fails](https://trello.com/c/r5z39J7X/5037-using-the-api-to-create-a-record-fails)

PR #4765, opened 2018-07-16
Hide the header when there is no sign in account available

amontenegro:


amontenegro:
[Add list of records user has Trusted Individual access to on /account](https://trello.com/c/OKQK1IQG/4773-add-list-of-records-user-has-trusted-individual-access-to-on-account)

PR #4766, opened 2018-07-16
Allow an org to update the sub type

amontenegro:


amontenegro:
[Dont display the preferred disambiguated org ext id twice on affiliations](https://trello.com/c/yQampT1n/5063-dont-display-the-preferred-disambiguated-org-ext-id-twice-on-affiliations)

PR #4767, opened 2018-07-17
Affiliation external IDs now normalised

bobcaprice:


bobcaprice:
[Should external identifiers on affiliations be normalized](https://trello.com/c/F863gh0m/4959-should-external-identifiers-on-affiliations-be-normalized)

PR #4768, opened 2018-07-17
Prevent NPE when the org belongs to the European Union

amontenegro:


amontenegro:
[Fundref loading breaking on EU](https://trello.com/c/6amHai7q/5061-fundref-loading-breaking-on-eu)

PR #4769, opened 2018-07-17
Allow certain members to use 1.2 api while it is disabled for all others

amontenegro:


amontenegro:
[Create a filter that allows 1.2 calls for just specific clients](https://trello.com/c/xuvbDCbG/5034-create-a-filter-that-allows-12-calls-for-just-specific-clients)

PR #4774, opened 2018-07-18
fix delete multiple sources args

lizkrznarich:


lizkrznarich:
[Works: No response when clicking trash icon for works w multiple sources](https://trello.com/c/nT5VpJhi/5068-works-no-response-when-clicking-trash-icon-for-works-w-multiple-sources)

PR #4777, opened 2018-07-18
works section use groups instead of workGroups

lizkrznarich:


lizkrznarich:
[No works display on the public view](https://trello.com/c/MLNy1Xjh/5070-no-works-display-on-the-public-view)

PR #4778, opened 2018-07-18
Fetch works in batches from db

amontenegro:


amontenegro:
Hi @wjrsimpson, 

Please take a look, with this we will fetch work entities in batches from the DB, reducing the number of times we hit it per every bulk request.

Thanks

amontenegro:
[Read work bulk improvements: Fetch works entities in batches from the DB](https://trello.com/c/BLlUMVQ3/5046-read-work-bulk-improvements-fetch-works-entities-in-batches-from-the-db)

PR #4780, opened 2018-07-19
fixed ordering in WorkSummary and added round trip test with schema v…

TomDemeranville:
…alidation


TomDemeranville:
[V3 bug - cannot deserialize using our own model](https://trello.com/c/82yVDt6k/5058-v3-bug-cannot-deserialize-using-our-own-model)

PR #4784, opened 2018-07-20
Script for building model and test before parent

bobcaprice:


bobcaprice:
[script for building model and test before parent](https://trello.com/c/01XnpDWE/5084-script-for-building-model-and-test-before-parent)

jpeerz:
I've updated the dev doc too
https://github.com/ORCID/ORCID-Source/pull/4775

PR #4789, opened 2018-07-21
Rollback artifact by version

jpeerz:


jpeerz:
found a less intrusive approach using versions maven plugin, pom files are updated on the fly.

    mvn -o versions:set -DnewVersion=1.0.0-perez


@wjrsimpson @amontenegro @rcpeters

PR #4792, opened 2018-07-23
works fix group item delete and bibtex export

lizkrznarich:


lizkrznarich:
[Works: deleting one source on a group results in no works displaying on the record.](https://trello.com/c/yKgIiO1H/5090-works-deleting-one-source-on-a-group-results-in-no-works-displaying-on-the-record)

lizkrznarich:
[BibTeX export doesn't generate file](https://trello.com/c/QSyrrRxw/5091-bibtex-export-doesnt-generate-file)

PR #4794, opened 2018-07-23
id normalizer endpoint example:///orcid-web/identifiers/norm/doi?value=10.1/11

TomDemeranville:
…i?value=10.1/111


TomDemeranville:
[Refactor external ID generation on UI to use server side generated normalized URLs.  Remove javascript version.](https://trello.com/c/h9WGGat8/4995-refactor-external-id-generation-on-ui-to-use-server-side-generated-normalized-urls-remove-javascript-version)

PR #4806, opened 2018-07-25
Fixed duplicated add some now message.

wjrsimpson:


wjrsimpson:
[Records with no works have "add some now" text duplicated](https://trello.com/c/2wcINQWL/5103-records-with-no-works-have-add-some-now-text-duplicated)

PR #4808, opened 2018-07-25
Id validator chrome issue

TomDemeranville:


TomDemeranville:
[Work identifier validator doesn't work on Chrome or Safari](https://trello.com/c/RYLmZy2Y/5093-work-identifier-validator-doesnt-work-on-chrome-or-safari)

PR #4809, opened 2018-07-25
fix work type not saving in manual entry form

lizkrznarich:


lizkrznarich:
[Manually adding works the type always saves as the default for the category](https://trello.com/c/PZ0uvlmu/5104-manually-adding-works-the-type-always-saves-as-the-default-for-the-category)

lizkrznarich:
[Manually adding works the type always saves as the default for the category](https://trello.com/c/PZ0uvlmu/5104-manually-adding-works-the-type-always-saves-as-the-default-for-the-category)

PR #4810, opened 2018-07-25
fix issue sorting external ids in bibtex import

lizkrznarich:


lizkrznarich:
[BibTeX loader is very slow to load works and shows them partially loaded](https://trello.com/c/pjRuiRtJ/5102-bibtex-loader-is-very-slow-to-load-works-and-shows-them-partially-loaded)

PR #4816, opened 2018-07-26
Password reset now validated in same way as registration password

bobcaprice:


bobcaprice:
[Reset password: doesn't check for common passwords or matching passwords](https://trello.com/c/hxhpbZw6/4937-reset-password-doesnt-check-for-common-passwords-or-matching-passwords)

PR #4818, opened 2018-07-26
Works with equal dates sorted by title

bobcaprice:


bobcaprice:
[Works with the same date should be in order by title alphabetically](https://trello.com/c/lF14b3Mi/5101-works-with-the-same-date-should-be-in-order-by-title-alphabetically)

PR #4819, opened 2018-07-26
bibtex import show error if file does not parse

lizkrznarich:


lizkrznarich:
[BibTeX importer return an error with invalid files](https://trello.com/c/OXBI0Aa0/5110-bibtex-importer-return-an-error-with-invalid-files)

PR #4821, opened 2018-07-27
Made sure status check returns 500 if it fails.

wjrsimpson:


wjrsimpson:
@amontenegro Can you take a look at the changes to Disable12ApiFilter.java. I changed it to stop togglz being called before the status check runs (so that we get the correct response from the status check if the db is down).

PR #4825, opened 2018-07-27
On 1.2 PUT, load all affiliations, including the new aff types, so, t…

amontenegro:
…hey don't get deleted by the orphan removal


amontenegro:
[Manually added new affiliations are deleted when you update affiliations with 1.2](https://trello.com/c/frEiAu5h/5114-manually-added-new-affiliations-are-deleted-when-you-update-affiliations-with-12)

PR #4828, opened 2018-07-31
small typos

dreamofaredbird:


caoyler:
Looks good, but since you're on the page anyway- why don't you update it to mark 1.2 as entirely sunset as of Aug 1.

dreamofaredbird:
added sunset dates

PR #4829, opened 2018-07-31
prevented empty biographies in DB

TomDemeranville:


TomDemeranville:
[Empty biography should result in no element in API responses. Stop making them empty.](https://trello.com/c/ymVMcH6T/5117-empty-biography-should-result-in-no-element-in-api-responses-stop-making-them-empty)

PR #4834, opened 2018-07-31
remove extra borders on peer review items

lizkrznarich:


lizkrznarich:
[peer-review has double lines on the box edges when view is expanded](https://trello.com/c/7ozpe0XL/5119-peer-review-has-double-lines-on-the-box-edges-when-view-is-expanded)

lizkrznarich:
[Public view show an empty work section if there is a researcher-resource item, but no works](https://trello.com/c/NgTD2lUC/5130-public-view-show-an-empty-work-section-if-there-is-a-researcher-resource-item-but-no-works)

PR #4835, opened 2018-07-31
hide works and research resources sections from public record if empty

lizkrznarich:


lizkrznarich:
[Public view show an empty work section if there is a researcher-resource item, but no works](https://trello.com/c/NgTD2lUC/5130-public-view-show-an-empty-work-section-if-there-is-a-researcher-resource-item-but-no-works)

PR #4837, opened 2018-08-01
Migrated 2FA state and setup to ng2

bobcaprice:


bobcaprice:
[/account: Two-factor authentication](https://trello.com/c/5E5DKG5a/4806-account-two-factor-authentication)

PR #4838, opened 2018-08-01
add ResetPasswordCtrlModal controller back to angular app

lizkrznarich:


lizkrznarich:
[Admin tools are listed twice, but can't be accessed](https://trello.com/c/jPx3T4K5/5133-admin-tools-are-listed-twice-but-cant-be-accessed)

PR #4841, opened 2018-08-01
change pom to build orcid-test and orcid-model

rcpeters:
Not sure why this isn't set up this way, there must be reason I don't understand?

### DIFF
--- a/DEVSETUP.md
+++ b/DEVSETUP.md
@@ -2,7 +2,7 @@
 
 ## Prerequisites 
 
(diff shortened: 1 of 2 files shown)


TomDemeranville:
I've dug a little deeper and this works great.  I thought that you could not have a module that didn't reference the parent, but that's not the case.  You can have a parent aggregate non-child modules.  FYI the change was made so that we could build model separate from parent because we were talking about adding it to mvn central.  With this change, we can still do that if we want.

https://stackoverflow.com/questions/1725579/maven-making-child-projects-that-can-be-independent-of-their-parent

Nice one Rob!

PR #4842, opened 2018-08-02
Show more details on public page also loads new peer review data from…

bobcaprice:
… server, both public and my-orcid now use callback function so more details only shown once new data is loaded.


bobcaprice:
[Peer review doesn't display subject in UI details view](https://trello.com/c/sVA5FbQe/5131-peer-review-doesnt-display-subject-in-ui-details-view)

PR #4847, opened 2018-08-03
affiliations use activePutCode to show/hide details instead of groupId

lizkrznarich:


lizkrznarich:
[New affiliations: Hovering over the visibility selector for one item brings it up for others](https://trello.com/c/MrmDBrg8/5126-new-affiliations-hovering-over-the-visibility-selector-for-one-item-brings-it-up-for-others)

PR #4849, opened 2018-08-06
added normalizedUrl to ActivityExternalIdentifier (hidden in API)

TomDemeranville:


TomDemeranville:
[Return normalized ID with work form. Keep existing endpoint.](https://trello.com/c/41dshBay/5125-return-normalized-id-with-work-form-keep-existing-endpoint)

PR #4851, opened 2018-08-06
show correct deactivated error on /reset-password

lizkrznarich:


lizkrznarich:
[Fix error message with deactivated records on reset-password](https://trello.com/c/mzVUiGmy/5160-fix-error-message-with-deactivated-records-on-reset-password)

PR #4854, opened 2018-08-07
Password reset send email in all cases

lizkrznarich:


lizkrznarich:
[Registry email / UI: Password reset messages](https://trello.com/c/bjkb713N/4586-registry-email-ui-password-reset-messages)

PR #4855, opened 2018-08-07
Add new endpoints to frontend security config

lizkrznarich:
Endpoints for affiliations, peer review, research resources and identifiers


lizkrznarich:
[New affiliations don't display on public record with http](https://trello.com/c/wnih0iaF/5159-new-affiliations-dont-display-on-public-record-with-http)

PR #4864, opened 2018-08-08
Ext ids get norm url from backend

lizkrznarich:


lizkrznarich:
[External Ids: Use normalized endpoint from pojo instead of sending request to API endpoint](https://trello.com/c/yPXBqFzj/5161-external-ids-use-normalized-endpoint-from-pojo-instead-of-sending-request-to-api-endpoint)

PR #4873, opened 2018-08-14
Set display index on creation

amontenegro:


amontenegro:
[Affiliations manually added to the new sections don't get a display index](https://trello.com/c/Zf7MudYO/5182-affiliations-manually-added-to-the-new-sections-dont-get-a-display-index)

PR #4875, opened 2018-08-14
Set it on 1.2 and 2.0

amontenegro:


amontenegro:
[Affiliations added with 1.2 or 2.0 don't get a display index](https://trello.com/c/ipeJRWHZ/5183-affiliations-added-with-12-or-20-dont-get-a-display-index)

PR #4876, opened 2018-08-15
Work in two groups

bobcaprice:


bobcaprice:
[Same work appears in two groups](https://trello.com/c/7OaUVSe4/5150-same-work-appears-in-two-groups)

PR #4879, opened 2018-08-16
added missing 2faSetupModule to orcid app

bobcaprice:


bobcaprice:
[Enabling 2FA now results in a blank screen](https://trello.com/c/o4ZgCFyB/5192-enabling-2fa-now-results-in-a-blank-screen)

PR #4880, opened 2018-08-16
Work type 'SOFTWARE' should map to 'OTHER' in 1.2

amontenegro:


amontenegro:
[1.2 API post/update calls return an error about software type.](https://trello.com/c/fDrpa5jA/5191-12-api-post-update-calls-return-an-error-about-software-type)

PR #4885, opened 2018-08-17
Public client fix

amontenegro:


amontenegro:
[Public API clients edited in manage-members interface cannot be edited in user developers tools](https://trello.com/c/sMB1uPEb/5195-public-api-clients-edited-in-manage-members-interface-cannot-be-edited-in-user-developers-tools)

PR #4886, opened 2018-08-20
show 'add some now' text on /my-orcid if no funding groups

lizkrznarich:


lizkrznarich:
[Empty funding section is missing "Add some now" link](https://trello.com/c/TNXy72SD/5203-empty-funding-section-is-missing-add-some-now-link)

PR #4888, opened 2018-08-20
Add some error messages for when the admin miss the email address or …

amontenegro:
…specify an email address that doesnt exists yet


amontenegro:
[Reactivate account from admin does not work](https://trello.com/c/vulSqFyV/5231-reactivate-account-from-admin-does-not-work)

PR #4895, opened 2018-08-21
When adding a new email, check the email hash so we can know if a dea…

amontenegro:
…ctivate orcid is available


amontenegro:
[Adding an email address associated with a deactivated record to an existing record fails without an error](https://trello.com/c/mNucSeRV/5239-adding-an-email-address-associated-with-a-deactivated-record-to-an-existing-record-fails-without-an-error)

PR #4896, opened 2018-08-21
Social signin wasn't using the same components as signin and oauthsig…

amontenegro:
…n in


amontenegro:
[Trying to use a deactivated email address when setting up social/inst sign in fails without an error](https://trello.com/c/IcJGUkVr/5240-trying-to-use-a-deactivated-email-address-when-setting-up-social-inst-sign-in-fails-without-an-error)

PR #4898, opened 2018-08-22
sendReactivationEmail function will get the email as param

amontenegro:


amontenegro:
[link to reactivate record on registration does not work](https://trello.com/c/1ZB0lg4W/5244-link-to-reactivate-record-on-registration-does-not-work)

PR #4900, opened 2018-08-23
Bibtex import loading icon

lizkrznarich:


lizkrznarich:
[Large BibTeX files only save some works](https://trello.com/c/7kP7qbnD/5237-large-bibtex-files-only-save-some-works)

PR #4901, opened 2018-08-23
add funding import wizard functions to ng2 funding component

lizkrznarich:


lizkrznarich:
[The Uberwizard no longer responds to user input.](https://trello.com/c/osMfhGBx/5248-the-uberwizard-no-longer-responds-to-user-input)

PR #4902, opened 2018-08-23
init null funding ext id urls when loading funding form

lizkrznarich:


lizkrznarich:
[When editing an existing a funding item, the identifier doesn't show self/part-of and more ids can't be added](https://trello.com/c/1yGBMND0/5241-when-editing-an-existing-a-funding-item-the-identifier-doesnt-show-self-part-of-and-more-ids-cant-be-added)

PR #4904, opened 2018-08-27
Research resource pojo fixes

lizkrznarich:


lizkrznarich:
[Research-resources with two hosts, don't show either org identifiers](https://trello.com/c/ZzFP6i0K/5261-research-resources-with-two-hosts-dont-show-either-org-identifiers)

PR #4905, opened 2018-08-28
Add sample work json file

alainna:
per member request

### DIFF
--- a/orcid-model/src/main/resources/record_2.1/samples/write_sample/work-full-2.1.json
+++ b/orcid-model/src/main/resources/record_2.1/samples/write_sample/work-full-2.1.json
@@ -1,0 +1,73 @@
+{
+  "language-code": "en", 
+  "contributors": {
+    "contributor": [
+      {
+        "contributor-orcid": {
+          "path": "0000-0001-5109-3700", 
+          "host": "orcid.org", 
+          "uri": "https://orcid.org/0000-0001-5109-3700"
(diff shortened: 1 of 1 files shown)


caoyler:
What's the reason for adding this file? Could we not just point users to https://github.com/ORCID/orcid-conversion-lib instead?  We've resisted adding JSON samples before because we don't want the additional burden of maintaining them (keeping the XML valid is hard enough!)

alainna:
We've had direct requests from members who asked why we offered the bulk works json but not individual works json. They've been referred to the converter, but still request the file. 

Perhaps another option could be to remove all json and add a README.md that refers to the converter. Thoughts?

caoyler:
We provided the bulk json in works as the json is significantly different when writing vs reading, so it's the once section that a member couldn't be expected to figure out on their own before we had the convertor. If there's a high demand for a single work json I think it's fine to add one file- but we do not want to be adding json examples for all the write options. What would you rather offer- json write samples for works but having to explain that we will not provide samples for other sections, or answering questions from members bulk posting works who can't figure out the json since the write doesn't look like read. 

I'll add a link to the convertor from the readmes.

alainna:
I'd rather provide the samples that people frequently request as it's faster than explaining why we don't have them and what to use instead. ; )

rcpeters:
This will have to be automated if we decide to process Closing for now.

PR #4909, opened 2018-08-28
Regex to match the verify email link shold include the lang

amontenegro:


amontenegro:
[Update the verify email page so it doesnt require login](https://trello.com/c/8YhKveWh/5149-update-the-verify-email-page-so-it-doesnt-require-login)

PR #4910, opened 2018-08-28
Since we have several invalid email addresses in prod, lets not valid…

amontenegro:
…ate the email address for updating the current/past status


amontenegro:
[Changing current/past on email address doesn't save](https://trello.com/c/RLYLTWQS/5263-changing-current-past-on-email-address-doesnt-save)

PR #4911, opened 2018-08-28
OrcidNoResultException was not mapped in the ExceptionMapper

amontenegro:


amontenegro:
[3.0 Pub API calling a record that doesn't exists returns a 500 error](https://trello.com/c/4A8XGMxD/5260-30-pub-api-calling-a-record-that-doesnt-exists-returns-a-500-error)

PR #4913, opened 2018-08-29
Funding backend grouping ui

lizkrznarich:


lizkrznarich:
[Text not saved as a string](https://trello.com/c/1xwOE3xQ/5264-text-not-saved-as-a-string)

PR #4916, opened 2018-08-29
Help icons all blue

lizkrznarich:


lizkrznarich:
[Some help popovers are wrong color](https://trello.com/c/FH2tPGjr/4944-some-help-popovers-are-wrong-color)

PR #4917, opened 2018-08-30
fixed employment funding mixup in swagger annotation

TomDemeranville:


TomDemeranville:
[Fix swagger employments in public api.](https://trello.com/c/wyrs473o/5170-fix-swagger-employments-in-public-api)

PR #4920, opened 2018-08-30
fix issues with funding ui ext ids, save and make a copy and edit

lizkrznarich:


lizkrznarich:
[Funding dates have an extra pipe after them](https://trello.com/c/MeH5klex/5273-funding-dates-have-an-extra-pipe-after-them)

lizkrznarich:
[API added funding items all tell you to edit your version instead of make a copy and edit, even when there is no user added version](https://trello.com/c/Sk5hoKo0/5272-api-added-funding-items-all-tell-you-to-edit-your-version-instead-of-make-a-copy-and-edit-even-when-there-is-no-user-added-versi)

lizkrznarich:
[Identifier section is missing on funding add manually form](https://trello.com/c/EXblTlMe/5271-identifier-section-is-missing-on-funding-add-manually-form)

lizkrznarich:
[Manually add funding doesn't save](https://trello.com/c/OJpSTzW5/5270-manually-add-funding-doesnt-save)

PR #4929, opened 2018-09-05
removes Consortia duplicated text

leomendoza123:


leomendoza123:
[Consortia page "organizations" written twice in the definition text. https://orcid.org/consortia](https://trello.com/c/ND5TmLpN/4961-consortia-page-organizations-written-twice-in-the-definition-text-https-orcidorg-consortia)

PR #4934, opened 2018-09-06
CSS tweak for bulk actions privacy popover

bobcaprice:


bobcaprice:
[The 'who can see this' pop up has a styling alignment issue](https://trello.com/c/T83jynPM/5276-the-who-can-see-this-pop-up-has-a-styling-alignment-issue)

PR #4941, opened 2018-09-11
Updates non existing URLs

leomendoza123:


leomendoza123:
[Registration form: Update link to passwords we don't allow](https://trello.com/c/EY5UEypj/5162-registration-form-update-link-to-passwords-we-dont-allow)

PR #4942, opened 2018-09-11
Fixes incorrect contact-us URL on public record header

leomendoza123:


leomendoza123:
[Contact Us link in public record header leads to incorrect address](https://trello.com/c/eGTnstfW/5197-contact-us-link-in-public-record-header-leads-to-incorrect-address)

PR #4943, opened 2018-09-12
Fixes deleting identifiers 

leomendoza123:


leomendoza123:
[Edit works, deleting identifiers always deletes the first one in the list](https://trello.com/c/zgQgdNpv/5283-edit-works-deleting-identifiers-always-deletes-the-first-one-in-the-list)

PR #4955, opened 2018-09-19
Refactor slow queries delete email event

amontenegro:


amontenegro:
[Round of SQL optimizations - deleting an email is a slow query](https://trello.com/c/OKz2D4aE/5317-round-of-sql-optimizations-deleting-an-email-is-a-slow-query)

PR #4957, opened 2018-09-19
On the public page, check if there are public peer reviews to display

amontenegro:


wjrsimpson:
@amontenegro merge monster!

PR #4972, opened 2018-09-28
remove custom emails functionality since no members have access to it

lizkrznarich:


lizkrznarich:
[/custom-emails](https://trello.com/c/O28yxyCb/5210-custom-emails)

PR #4988, opened 2018-10-03
Added client id to access logs for token endpoint.

wjrsimpson:


wjrsimpson:
[Add client id to logs for /oauth/token](https://trello.com/c/v6iGYGx7/5352-add-client-id-to-logs-for-oauth-token)

PR #4989, opened 2018-10-03
Create two different config variables, one for the max number of elem…

amontenegro:
…ents to post per bulk, and, other one for the max number of elements to get per bulk


amontenegro:
[Have different config values for read/write works bulk](https://trello.com/c/tcursDUJ/5109-have-different-config-values-for-read-write-works-bulk)

PR #4994, opened 2018-10-05
hide import wizard when works or funding section closed

lizkrznarich:


lizkrznarich:
[Funding chevron does not close search and link wizard](https://trello.com/c/YgbsM6s2/5355-funding-chevron-does-not-close-search-and-link-wizard)

PR #4997, opened 2018-10-08
Ensured names are not duplicated on institutional sign in registration

wjrsimpson:


wjrsimpson:
[Ensure only one name auto-populated during institution sign in registration](https://trello.com/c/xFRJJwql/5365-ensure-only-one-name-auto-populated-during-institution-sign-in-registration)

PR #5001, opened 2018-10-09
Link social show oauth

lizkrznarich:


lizkrznarich:
[Oauth With Social Sign in is not working correctly](https://trello.com/c/C2RrAzOB/5359-oauth-with-social-sign-in-is-not-working-correctly)

PR #5002, opened 2018-10-10
Fixed JPA problem removing access for self service contact.

wjrsimpson:


dreamofaredbird:
[Remove contact in self service is not working](https://trello.com/c/YDqat3dN/5369-remove-contact-in-self-service-is-not-working)

PR #5012, opened 2018-10-22
remove duplicate last modified text from public record

lizkrznarich:


lizkrznarich:
["Record last modified" is duplicated on public record](https://trello.com/c/JpTDcAsr/5381-record-last-modified-is-duplicated-on-public-record)

PR #5015, opened 2018-10-23
Affiliations no dates

lizkrznarich:


lizkrznarich:
[Affiliations with no dates display as "to present"](https://trello.com/c/glS77mDW/5386-affiliations-with-no-dates-display-as-to-present)

PR #5018, opened 2018-10-24
Admin actions fixes

amontenegro:


amontenegro:
[When adding a delegate email the validator sees a space as a character and does not trigger an error message](https://trello.com/c/RKkKT4LN/5380-when-adding-a-delegate-email-the-validator-sees-a-space-as-a-character-and-does-not-trigger-an-error-message)

amontenegro:
[Admin find records can't find if email has a plus sign](https://trello.com/c/iE6KluO6/5382-admin-find-records-cant-find-if-email-has-a-plus-sign)

PR #5020, opened 2018-10-24
Use the userStatus.json endpoint to verify if a user is logged in or …

amontenegro:
…not, if it is, then, fetch the public name, if not, just ignore doing the names call

### DIFF
--- a/orcid-web/src/main/webapp/static/javascript/ng1Orcid/app/modules/oauthAuthorization/oauthAuthorization.component.ts
+++ b/orcid-web/src/main/webapp/static/javascript/ng1Orcid/app/modules/oauthAuthorization/oauthAuthorization.component.ts
@@ -737,28 +737,41 @@ export class OauthAuthorizationComponent implements AfterViewInit, OnDestroy, On
             }
         );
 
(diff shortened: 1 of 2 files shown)


amontenegro:
Hi @lizkrznarich, could you please review this code?

The code that fetches the public name on the signin page makes spring believe that the last user call was to `/account/names/public`, which makes us lose the real url the user want to visit after login, here is a card describing the problem:

https://trello.com/c/7UIpCe50/5378-clicking-on-the-orcid-inbox-link-in-the-update-email-when-signed-out-does-not-redirect-you-to-inbox-when-you-sign-in 

With this PR, we will call the userStatus.json endpoint to check if the user is logged in or not, and, if it is, it will then call the  `/account/names/public` endpoint to get the public name of the authenticated user.

Please let me know if you have any question, comment, concern.

Thanks

amontenegro:
[Clicking on the Orcid Inbox link in the update email when signed out does not redirect you to inbox when you sign in](https://trello.com/c/7UIpCe50/5378-clicking-on-the-orcid-inbox-link-in-the-update-email-when-signed-out-does-not-redirect-you-to-inbox-when-you-sign-in)

PR #5021, opened 2018-10-25
Tweak to how v3 package names differ according to minor versions

bobcaprice:


bobcaprice:
[3.0_rc1 is broken](https://trello.com/c/lctJOXum/5392-30rc1-is-broken)

PR #5022, opened 2018-10-25
make sure specials chars are escaped

lizkrznarich:


lizkrznarich:
[Public record: PublicRecordCtrl](https://trello.com/c/3YBMqPxV/5220-public-record-publicrecordctrl)

PR #5023, opened 2018-10-26
Changed create work group algorithm to handle multiple user source wo…

bobcaprice:
…rks in single group


bobcaprice:
[merging with PART OF relationship ext ids](https://trello.com/c/2V8HKFzP/5373-merging-with-part-of-relationship-ext-ids)

PR #5024, opened 2018-10-26
Fixed externalIdsPresent field so components pick up value regardless…

bobcaprice:
… of whether true or false


bobcaprice:
[merging work bug](https://trello.com/c/klQ6SuVF/5372-merging-work-bug)

PR #5027, opened 2018-10-26
Dissertation thesis

amontenegro:


amontenegro:
[Migrate work type dissertation to dissertation-thesis in 3.0_rc2](https://trello.com/c/k55LUUIr/5398-migrate-work-type-dissertation-to-dissertation-thesis-in-30rc2)

PR #5030, opened 2018-10-30
fix email prefs line breaks and mobile width

lizkrznarich:


lizkrznarich:
[Line break is missing on notification frequency selector](https://trello.com/c/hzlPWrh6/5404-line-break-is-missing-on-notification-frequency-selector)

PR #5034, opened 2018-10-30
make /claim UI match /register UI

lizkrznarich:


lizkrznarich:
[Styling differences on Claim vs Register](https://trello.com/c/UI0aiMGT/5379-styling-differences-on-claim-vs-register)

PR #5037, opened 2018-10-30
Validate the id before showing the reset password confirm

amontenegro:


amontenegro:
[Could we make the Notification happen earlier for incorrect ORCIDs on reset password?](https://trello.com/c/RFl9vzyQ/5375-could-we-make-the-notification-happen-earlier-for-incorrect-orcids-on-reset-password)

PR #5041, opened 2018-10-31
Removed duplicate work validation and tidied up incorrect version con…

bobcaprice:
…version in delegator (which wasn't doing anything yet but shouldn't be looking for v2.1)


bobcaprice:
[3.0 bulk works post returns an error](https://trello.com/c/m9CAKliM/5406-30-bulk-works-post-returns-an-error)

PR #5044, opened 2018-11-01
Profile history event for password reset

bobcaprice:


bobcaprice:
[Record password reset datetime](https://trello.com/c/q27uNomY/5388-record-password-reset-datetime)

PR #5046, opened 2018-11-02
Corrected setting of client url

bobcaprice:


bobcaprice:
[3.0 "/client" is no longer in source uris for API clients](https://trello.com/c/e6faHjlC/5419-30-client-is-no-longer-in-source-uris-for-api-clients)

PR #5051, opened 2018-11-02
wrap-fix

leomendoza123:


leomendoza123:
[Arabic language nav is partially obscured](https://trello.com/c/d0XIsc7D/5371-arabic-language-nav-is-partially-obscured)

PR #5052, opened 2018-11-02
Improvements on public page

amontenegro:


amontenegro:
[Refactor public page so it loads faster](https://trello.com/c/1zuS4DeL/5320-refactor-public-page-so-it-loads-faster)

PR #5054, opened 2018-11-05
don't use {} in translated strings

lizkrznarich:


lizkrznarich:
[When clicking a verify email when switched to french language the /myorcid page fails to load](https://trello.com/c/I9GWZVbH/5428-when-clicking-a-verify-email-when-switched-to-french-language-the-myorcid-page-fails-to-load)

PR #5057, opened 2018-11-05
use correct  var in peer review showMoreDetails

lizkrznarich:


lizkrznarich:
[Convening org & created date not displayed in peer review](https://trello.com/c/3YRrQ7aB/5431-convening-org-created-date-not-displayed-in-peer-review)

PR #5061, opened 2018-11-06
Removed call to byDefault() in addV2CommonFields / addV3CommonFields …

bobcaprice:
…methods, moved to calling methods


bobcaprice:
[remove byDefault() call from mapper facade add*CommonFields method](https://trello.com/c/NyepYXuE/5422-remove-bydefault-call-from-mapper-facade-addcommonfields-method)

PR #5063, opened 2018-11-06
Inbox ng2 fixes

lizkrznarich:


lizkrznarich:
[Inbox archive link within messages does not work](https://trello.com/c/IA9tbXoh/5439-inbox-archive-link-within-messages-does-not-work)

lizkrznarich:
[Inbox doesn't load if there is a trusted individual message](https://trello.com/c/tXt7sNxp/5437-inbox-doesnt-load-if-there-is-a-trusted-individual-message)

lizkrznarich:
[Inbox includes "ORCID" in all from fields](https://trello.com/c/CgToAVg3/5438-inbox-includes-orcid-in-all-from-fields)

PR #5064, opened 2018-11-07
Bulk works rejected if POSTed with bulk elements of type error

bobcaprice:


bobcaprice:
[don't allow BulkElement of type error to be posted to bulk works endpoint](https://trello.com/c/hI6jFcNP/5416-dont-allow-bulkelement-of-type-error-to-be-posted-to-bulk-works-endpoint)

PR #5066, opened 2018-11-07
Fixed problem where names were hidden in self service, even when public.

wjrsimpson:


wjrsimpson:
[Self-service--update contacts not working properly](https://trello.com/c/FyfogeAy/5421-self-service-update-contacts-not-working-properly)

PR #5070, opened 2018-11-07
ext ids handle case where url field exists but has no value

lizkrznarich:


lizkrznarich:
[identifiers missing in peer-review UI](https://trello.com/c/ATr5WXN4/5447-identifiers-missing-in-peer-review-ui)

PR #5071, opened 2018-11-07
Admin should be able to switch to disabled records

amontenegro:


amontenegro:
[Unable to admin into locked records](https://trello.com/c/pAY4CObj/5444-unable-to-admin-into-locked-records)

PR #5073, opened 2018-11-07
Fix action executed when hitting enter key on switch user, reactivate…

amontenegro:
… and look up ids


amontenegro:
[Hitting enter on admin switch user returns a not found error](https://trello.com/c/Sy9gGvrA/5418-hitting-enter-on-admin-switch-user-returns-a-not-found-error)

PR #5079, opened 2018-11-08
Fix loading id section

leomendoza123:


orcid-web/src/main/resources/freemarker/includes/ng2_templates/id-banner-ng2-template.ftl
```diff
@@ -0,0 +1,38 @@
+<script type="text/ng-template" id="id-banner-ng2-template">
+    <div class="id-banner <#if inDelegationMode>delegation-mode</#if>"> 
+        <#if inDelegationMode><span class="delegation-mode-warning">${springMacroRequestContext.getMessage("delegate.managing_record")}</span></#if>
+        <!-- Name -->
+
+        <#if springMacroRequestContext.requestUri?contains("/my-orcid")>
```
leomendoza123: @lizkrznarich Hi! 
I am using this conditional to check if I am on `/my-orcid` or in the public version. Is this a good way, is there any corner case that might break this?

leomendoza123:
[Id section displays while page is loading](https://trello.com/c/rUTsP3cb/5430-id-section-displays-while-page-is-loading)

PR #5080, opened 2018-11-08
When admin into a locked record, show the switch user form with the a…

amontenegro:
…dmin on it so you can switch back


amontenegro:
[When admined into a locked record there is no option to switch back to your own record](https://trello.com/c/FCxCgET9/5452-when-admined-into-a-locked-record-there-is-no-option-to-switch-back-to-your-own-record)

PR #5081, opened 2018-11-09
activities ui handle case where ext id relationship is null

lizkrznarich:


lizkrznarich:
[identifiers missing in peer-review UI](https://trello.com/c/ATr5WXN4/5447-identifiers-missing-in-peer-review-ui)

PR #5082, opened 2018-11-09
Exclude lei from orgs drop down

bobcaprice:


bobcaprice:
[Add LEI](https://trello.com/c/VC3OhMUv/5366-add-lei)

PR #5083, opened 2018-11-09
Affiliations, fundings, and research resources now require a start da…

bobcaprice:
…te in V3 if an end date is supplied


bobcaprice:
[Require start date if end date is present in 3.0+ including RC1](https://trello.com/c/C9ntCyuw/5453-require-start-date-if-end-date-is-present-in-30-including-rc1)

PR #5088, opened 2018-11-12
fix-works-id

leomendoza123:


leomendoza123:
[Identifier type is missing from UI](https://trello.com/c/4jaRupn0/5455-identifier-type-is-missing-from-ui)

leomendoza123:
[Identifier type is missing from UI](https://trello.com/c/4jaRupn0/5455-identifier-type-is-missing-from-ui)

PR #5090, opened 2018-11-13
Notification alerts fix auth url

lizkrznarich:


lizkrznarich:
[Link in alert for institutional round trip does not work](https://trello.com/c/kLuRDNZc/5458-link-in-alert-for-institutional-round-trip-does-not-work)

PR #5094, opened 2018-11-14
fix-link-works-url

leomendoza123:


leomendoza123:
[missing url on Search and Link Works text.](https://trello.com/c/yGEvNlil/5464-missing-url-on-search-and-link-works-text)

PR #5095, opened 2018-11-15
Added debug for updating label

bobcaprice:


bobcaprice:
[Fix display for "Conference title" when adding conference info manually](https://trello.com/c/h1Lpiujs/5450-fix-display-for-conference-title-when-adding-conference-info-manually)

PR #5096, opened 2018-11-16
Activities add created last mod

lizkrznarich:


lizkrznarich:
[Date created and last modified should be exposed in all activity details UI](https://trello.com/c/B0arX1lV/5393-date-created-and-last-modified-should-be-exposed-in-all-activity-details-ui)

PR #5097, opened 2018-11-16
New endpoint in workspace service for getting map of country names ma…

bobcaprice:
…pped to country codes to correct ordering of country names in ui


bobcaprice:
[Countries: Display regions in alphabetic order based on name](https://trello.com/c/EEN7YPNE/5451-countries-display-regions-in-alphabetic-order-based-on-name)

PR #5102, opened 2018-11-19
save-all-for-large-bibtex add timeout for each request

leomendoza123:


leomendoza123:
[Large BibTeX files only save some works](https://trello.com/c/7kP7qbnD/5237-large-bibtex-files-only-save-some-works)

leomendoza123:
@lizkrznarich Uploading works in batches of 20. We might create a new ticket to solve this later using an endpoint that supports a bulk of works as you propose.

PR #5105, opened 2018-11-20
print-bio-fix remove angularJs hide

leomendoza123:


leomendoza123:
[Biography section missing from print view](https://trello.com/c/Q4oVo29Y/5456-biography-section-missing-from-print-view)

PR #5106, opened 2018-11-20
Lowercase email before looking at accounts to reactivate

amontenegro:


amontenegro:
[Error reactivating email addresses with different cases.](https://trello.com/c/iYTjQZM4/5469-error-reactivating-email-addresses-with-different-cases)

amontenegro:
[Refactor email hash creation so it happens in a single manager instead of happening in different places](https://trello.com/c/UsaAEG9J/5475-refactor-email-hash-creation-so-it-happens-in-a-single-manager-instead-of-happening-in-different-places)

PR #5108, opened 2018-11-21
Token exchange

TomDemeranville:


TomDemeranville:
[Create token exchange endpoint](https://trello.com/c/0t4n3eyL/5350-create-token-exchange-endpoint)

PR #5111, opened 2018-11-26
Maintain list of member chosen orgs

bobcaprice:


bobcaprice:
[Provide a list of member chosen ORG IDs in the registry.](https://trello.com/c/Kc5IziHu/4841-provide-a-list-of-member-chosen-org-ids-in-the-registry)

PR #5115, opened 2018-11-27
Fix bio and name variables

leomendoza123:


leomendoza123:
[Names are missing on the public ORCID view](https://trello.com/c/zPxM7Ynx/5484-names-are-missing-on-the-public-orcid-view)

PR #5119, opened 2018-11-27
remove top line

leomendoza123:


leomendoza123:
[Id section displays while page is loading](https://trello.com/c/rUTsP3cb/5430-id-section-displays-while-page-is-loading)

PR #5121, opened 2018-11-28
don't display email addresses as links on public record

lizkrznarich:


lizkrznarich:
[Print view has email addresses listed twice](https://trello.com/c/32JU5Ho1/5491-print-view-has-email-addresses-listed-twice)

PR #5124, opened 2018-11-28
add missing noto font css back to html-head

lizkrznarich:


lizkrznarich:
[Noto font missing if not installed locally](https://trello.com/c/xvKO6PCt/5495-noto-font-missing-if-not-installed-locally)

PR #5127, opened 2018-11-30
fix switch for locked users

leomendoza123:


leomendoza123:
[Fix switch for locked users](https://trello.com/c/mGBFERGy/5503-fix-switch-for-locked-users)

PR #5130, opened 2018-12-05
Removing ActivitiesStatusDaoTest's test-order dependency

OrDTesters:
ActivitiesStatusDaoTest fails when run after ActivitiesStatusManagerTest. More specifically, ActivitiesStatusDaoTest.getFailedElementsTest fails when run after ActivitiesStatusManagerTest.markAsFailedTest since markAsFailedTest adds a new failed element and getFailedElementsTest counts the extra failed element, leading to an assertion failure. The proposed patch is to set the element to be sent at the end of markAsFailedTest. Removing the test-order dependency allows these tests to be run in any order.

Ideally, it would be good to remove the added elements entirely at the end of the test, but I could not find the relevant API call to do this. I would be happy to discuss the patch more.


wjrsimpson:
@OrDTesters Thank you for the pull request! We'd love to know more about you and about interesting stuff you are doing with ORCID.

OrDTesters:
I am a researcher working on detecting and removing test-order dependencies. Our tooling found this dependency between tests in ORCID and helped me with the fix. There are a few more I plan on fixing and sending pull requests for.

PR #5133, opened 2018-12-05
Merge works ui

lizkrznarich:


lizkrznarich:
[2.a Make it easier for end users to group items in the record.](https://trello.com/c/mqtkPacU/4636-2a-make-it-easier-for-end-users-to-group-items-in-the-record)

PR #5134, opened 2018-12-06
Removing error:error from write samples

dreamofaredbird:


dreamofaredbird:
[Review bulk write samples and remove error text](https://trello.com/c/2O9ITkoE/5515-review-bulk-write-samples-and-remove-error-text)

PR #5136, opened 2018-12-06
Fix popover position

leomendoza123:


leomendoza123:
[Privacy selector popover is misaligned in activities sections](https://trello.com/c/OMuANnzK/5518-privacy-selector-popover-is-misaligned-in-activities-sections)

PR #5145, opened 2018-12-14
Fix mange members find member

leomendoza123:


leomendoza123:
[/manage-members find member cannot find member by anything other client id, member id and email result in nothing happening](https://trello.com/c/EctZwrIM/5525-manage-members-find-member-cannot-find-member-by-anything-other-client-id-member-id-and-email-result-in-nothing-happening)

PR #5146, opened 2018-12-17
Use master to get final list of notifications to send to user

wjrsimpson:


wjrsimpson:
[Use master DB to read notifications to send](https://trello.com/c/wS0uwdhm/5533-use-master-db-to-read-notifications-to-send)

PR #5147, opened 2018-12-17
manage members switch account fix

leomendoza123:


leomendoza123:
[click on the 'here' to switch to member account when you have created it in /manage-members fails](https://trello.com/c/yYQDoZIa/5524-click-on-the-here-to-switch-to-member-account-when-you-have-created-it-in-manage-members-fails)

PR #5148, opened 2018-12-17
Removing RecordStatusManagerTest's test-order dependency

OrDTesters:
Similar to #5130, RecordStatusDaoTest fails when run after RecordStatusManagerTest. The failure is due to same reason as in #5130, concerning extra failed elements added in RecordStatusManagerTest.markAsFailedTest. The proposed patch is similar to what was proposed in #5130 as well.


lizkrznarich:
[Removing RecordStatusManagerTest's test-order dependency](https://trello.com/c/uwV6DRN9/5567-removing-recordstatusmanagertests-test-order-dependency)

PR #5150, opened 2018-12-18
Flag notifications as read individually

wjrsimpson:


wjrsimpson:
[Flag notifications as sent individually](https://trello.com/c/2RLdOfw4/5535-flag-notifications-as-sent-individually)

PR #5152, opened 2018-12-19
changed id_token to expire in 20 years

TomDemeranville:


TomDemeranville:
[id_token exchange: id tokens from access token say they expire in 10 minutes](https://trello.com/c/amBuCZDM/5536-idtoken-exchange-id-tokens-from-access-token-say-they-expire-in-10-minutes)

PR #5155, opened 2018-12-19
Fix locked users name

leomendoza123:


leomendoza123:
[Name displays on locked records](https://trello.com/c/L6THOjNs/5530-name-displays-on-locked-records)

PR #5165, opened 2018-12-24
Fix bibtex second upload

leomendoza123:


leomendoza123:
[Once you have deleted a work if you try to add one via bibtex importer the file will not load](https://trello.com/c/B3Kl4knO/5550-once-you-have-deleted-a-work-if-you-try-to-add-one-via-bibtex-importer-the-file-will-not-load)

PR #5166, opened 2018-12-28
Remove empty Act typ and geo area on client redirect uris, so, it is …

amontenegro:
…null not an empty string


amontenegro:
[Remove empty act type and geo area from manage members edit client](https://trello.com/c/xqx3yqdJ/5554-remove-empty-act-type-and-geo-area-from-manage-members-edit-client)

PR #5168, opened 2019-01-02
Fix bibtex dissertation thesis upload

leomendoza123:


leomendoza123:
[Unable to save bibtex file if it has @phdthesis as the work type](https://trello.com/c/IogVLtY5/5551-unable-to-save-bibtex-file-if-it-has-phdthesis-as-the-work-type)

PR #5169, opened 2019-01-02
Fix other ids wrap

leomendoza123:


leomendoza123:
[Other iDs section is missing line breaks in public view](https://trello.com/c/dtlGzkI7/5556-other-ids-section-is-missing-line-breaks-in-public-view)

PR #5173, opened 2019-01-04
add label, name and id to form input so autofill can bind data

lizkrznarich:


lizkrznarich:
[Unable to add another email if autocomplete turned on in Chrome.](https://trello.com/c/O0Kp5jAK/5546-unable-to-add-another-email-if-autocomplete-turned-on-in-chrome)

PR #5174, opened 2019-01-07
Corrected check for read-only scopes to check for openid / read-publi…

bobcaprice:
…c / authenticate scopes


bobcaprice:
[Implicit process now allows read-limited scope](https://trello.com/c/EX2U98zM/5553-implicit-process-now-allows-read-limited-scope)

PR #5177, opened 2019-01-07
On extractSourceFromEntity check if the source_id field contains a le…

amontenegro:
…gacy id


amontenegro:
Hi @TomDemeranville, could you please take a look at this PR, it affects how we handle the legacy client ids in the source_id field and affects 3.0+, the side effect will be that, when editing an element with a legacy client_id in the source_id field, the code will copy that id from the source_id to the client_source_id field.

Thanks

amontenegro:
[3.0 returning client source in source-id field on one record](https://trello.com/c/Nm3dhJ45/5548-30-returning-client-source-in-source-id-field-on-one-record)

TomDemeranville:
This will work.

However, I worry that whatever is calling extractSourceFromEntity(SourceAwareEntity) is passing in entities that have client IDs in the wrong place.  The entities should not be created /wrong/ in the first place.

PR #5178, opened 2019-01-08
Add info about how to add consortium logo on self service

leomendoza123:


leomendoza123:
[Self Service - Add text to the Public Display section for member to send logo to support@orcid.org (0.5)](https://trello.com/c/sHvClwAN/4451-self-service-add-text-to-the-public-display-section-for-member-to-send-logo-to-supportorcidorg-05)

PR #5179, opened 2019-01-08
Make additional org urls into links

leomendoza123:


leomendoza123:
[Make additional org urls into links (2)](https://trello.com/c/msgVr7UL/5086-make-additional-org-urls-into-links-2)

PR #5180, opened 2019-01-08
Get funding import wizards from endpoint

lizkrznarich:


lizkrznarich:
[Template parse error on my-orcid](https://trello.com/c/AEQtXYI2/5561-template-parse-error-on-my-orcid)

PR #5181, opened 2019-01-09
Ma test edits

dreamofaredbird:
Added browserstack ,  mobile and link to self service tests to MA tests


lizkrznarich:
@dreamofaredbird merge monster!

PR #5182, opened 2019-01-09
Fix adding multiple URLs of the same type

leomendoza123:
- add missing unsubscriptions

### DIFF
--- a/orcid-web/src/main/webapp/static/javascript/ng1Orcid/app/modules/manageMembers/manageMembersFindClient.component.ts
+++ b/orcid-web/src/main/webapp/static/javascript/ng1Orcid/app/modules/manageMembers/manageMembersFindClient.component.ts
@@ -33,6 +33,7 @@ export class ManageMembersFindClientComponent
   _client;
   scopes = [];
   selectedScopes = [];
+  modalSubscription : Subscription
   actTypeList = ["Articles", "Books", "Data", "Student Publications"].map(
     geo => {
       return { name: geo };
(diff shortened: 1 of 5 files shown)


orcid-web/src/main/webapp/static/javascript/ng1Orcid/app/modules/manageMembers/manageMembersFindClient.component.ts
```diff
@@ -128,13 +128,25 @@ export class ManageMembersFindClientComponent
...
       ) {
+        
+        if (data.object.redirectUris) {
+          data.object.redirectUris.forEach(element => {
+            if (element.actType && !element.actType.value) {
+              element.actType.value = "\"\""
+            }
+            if (element.geoArea && !element.geoArea.value) {
+              element.geoArea.value = "\"\""
+            }
```
leomendoza123: @lizkrznarich This was particularly hard to find. The backend needs quotes on empty strings, and this problem only happens if multiple URL of the same type are been added :mindblow:.
I am temporally fixing this on the front end, but maybe we can solve this on the backend later.

leomendoza123:
[Unable to update Client details from /manage-members. Changes do not save.](https://trello.com/c/FTZjNt94/5560-unable-to-update-client-details-from-manage-members-changes-do-not-save)

orcid-web/src/main/webapp/static/javascript/ng1Orcid/app/modules/manageMembers/manageMembersFindClient.component.ts
lizkrznarich: @leomendoza123 Agree that we need to fix this up. It's a json field in the DB, but the data there is not consistent (values for default redirects vary between null, "", {"import-works-wizard":[]}, and{"import-works-wizard":[]} and {"import-works-wizard":["Global"]}. Will file some cards to clean up existing data and make sure geo are and act type are added consistently in future.

PR #5195, opened 2019-01-16
only generate grouping suggestions if GROUPING_SUGGESTIONS togglz ena…

lizkrznarich:
…bled


lizkrznarich:
[Records with no works are missing the "0 of 0" count](https://trello.com/c/EbTaG7XJ/5580-records-with-no-works-are-missing-the-0-of-0-count)

PR #5199, opened 2019-01-21
Merge works error if missing api self ext

lizkrznarich:


lizkrznarich:
[Merge works with only part-of identifiers](https://trello.com/c/SewaQwMf/5540-merge-works-with-only-part-of-identifiers)

lizkrznarich:
[500 error when trying to merge a work with no identifier added by a client](https://trello.com/c/Dn5Jj543/5542-500-error-when-trying-to-merge-a-work-with-no-identifier-added-by-a-client)

PR #5204, opened 2019-01-22
Fixed failure if metadata language value doesn't match our language c…

bobcaprice:
…ode value


bobcaprice:
[Works metadata resolver: Is it possible to have a different error when we can't pull metadata vs. when an identifier is incorrect](https://trello.com/c/YaFUz9eE/5599-works-metadata-resolver-is-it-possible-to-have-a-different-error-when-we-cant-pull-metadata-vs-when-an-identifier-is-incorrect)

PR #5205, opened 2019-01-22
ISSN external id now defaults to relationship PART_OF

bobcaprice:


bobcaprice:
[Works metadata resolver: Have issns default to part-of](https://trello.com/c/9VQ0hSIG/5592-works-metadata-resolver-have-issns-default-to-part-of)

PR #5206, opened 2019-01-22
If DOI data subtitle is encased with [], brackets are removed. If res…

bobcaprice:
…ult is encased in double quotes, these are also removed


bobcaprice:
[DOI importer: subtitles have extra brackets](https://trello.com/c/iNT5gTeS/5598-doi-importer-subtitles-have-extra-brackets)

PR #5207, opened 2019-01-22
Added default type of PREPRINT to arxiv importer

bobcaprice:


bobcaprice:
[ARXIV importer does not include type](https://trello.com/c/1a7FjQKr/5597-arxiv-importer-does-not-include-type)

PR #5211, opened 2019-01-22
Fix missing interpolation bracket

leomendoza123:


leomendoza123:
[Works metadata resolver: update error text](https://trello.com/c/zajCiMa7/5600-works-metadata-resolver-update-error-text)

PR #5214, opened 2019-01-23
Update pubMed placeholder

leomendoza123:


leomendoza123:
[Update PubMed import helper text to include PMC example](https://trello.com/c/GPQqLGJs/5606-update-pubmed-import-helper-text-to-include-pmc-example)

PR #5215, opened 2019-01-23
PIDs now URLEncoded before fetching metadata

bobcaprice:


bobcaprice:
[DOI import tool: check logic we use to determine if a DOI is valid](https://trello.com/c/2oYrM01t/5591-doi-import-tool-check-logic-we-use-to-determine-if-a-doi-is-valid)

lizkrznarich:
@bobcaprice Can you have a look at the PIDControllerTest fail?

PR #5219, opened 2019-01-24
DOI resolver now populates publicationDate according to 'issued' fiel…

bobcaprice:
…d if no other published dates available


bobcaprice:
[DOI importer parse dates from DataCite DOIs](https://trello.com/c/YsmUjHNV/5607-doi-importer-parse-dates-from-datacite-dois)

PR #5220, opened 2019-01-24
Pub med resolver default work type of journal article

bobcaprice:


bobcaprice:
[PubMed importer have type default to journal article](https://trello.com/c/lNhhP6k7/5604-pubmed-importer-have-type-default-to-journal-article)

PR #5223, opened 2019-01-28
Improved url validation such that path is now url encoded

bobcaprice:


bobcaprice:
[DOI import tool: check logic we use to determine if a DOI is valid](https://trello.com/c/2oYrM01t/5591-doi-import-tool-check-logic-we-use-to-determine-if-a-doi-is-valid)

PR #5225, opened 2019-01-29
allow setting locale_v3 cookie to de (german)

lizkrznarich:


lizkrznarich:
[Add German to Drupal lang selector on QA (2)](https://trello.com/c/HnScHcs9/5615-add-german-to-drupal-lang-selector-on-qa-2)

PR #5227, opened 2019-01-29
PIDs returning 301 now considered valid

bobcaprice:


bobcaprice:
[Why is this DOI invalid?](https://trello.com/c/2xb6I0ee/5617-why-is-this-doi-invalid)

PR #5224, opened 2019-01-28
Remove12 code from search

amontenegro:

PR #5229, opened 2019-01-29
Revert "Remove12 code from search"

lizkrznarich:
Reverts ORCID/ORCID-Source#5224


amontenegro:
@lizkrznarich this is ok! so, we can merge it as soon as the CI finish.

PR #5234, opened 2019-01-29
Manual test jan19

caoyler:
@dreamofaredbird Is there a reason self-service step 14-18 involve logging back into the consortium lead instead of just logging into the account that was added and checking that Member Tools can be accessed from it?


dreamofaredbird:
@caoyler I think this is my mistake as I just pulled master and it still has self service tests in. It shouldn't. You can just remove this section. Or I can. I think it got messed up in a merge conflict.

PR #5238, opened 2019-01-31
Fix bug where records with CS local fail to load on 2.0 and 2.1

amontenegro:


amontenegro:
[Records with CS locale return an error on 2.0 and 2.1, also 3.0_rc1 and 3.0_rc2 returns EN as the locale for all records](https://trello.com/c/ZBu3RlM8/5643-records-with-cs-locale-return-an-error-on-20-and-21-also-30rc1-and-30rc2-returns-en-as-the-locale-for-all-records)

PR #5240, opened 2019-01-31
Fix notification with null source-name

leomendoza123:
For this solution if the `notification.source.sourceName` is null the displayed name is going to be "Orcid". As apparently was initially intended when the verification for null `notification.source` displayed "Orcid". 

The main difference is that  is that if `sourceName` is null, this is going to display "Orcid" no matter if `notification.overwrittenSourceName`  is true or false.


leomendoza123:
[Some notifications don't load in the UI](https://trello.com/c/T87ruQ6V/5661-some-notifications-dont-load-in-the-ui)

PR #5251, opened 2019-02-07
Ensured null never put into grouping suggestions cache

bobcaprice:


bobcaprice:
[Return up to 100 suggestions from server](https://trello.com/c/0K6Cwb5J/5658-return-up-to-100-suggestions-from-server)

lizkrznarich:
[UI is not loading Works](https://trello.com/c/bSRILf3B/5672-ui-is-not-loading-works)

PR #5252, opened 2019-02-08
Frontend obo

leomendoza123:


leomendoza123:
[OBO UI display](https://trello.com/c/V3R3VuNB/5482-obo-ui-display)

PR #5255, opened 2019-02-11
Print view hide maintenance msg

lizkrznarich:


lizkrznarich:
[Note about survey appears when you print your record](https://trello.com/c/VpUu90Sp/5681-note-about-survey-appears-when-you-print-your-record)

PR #5256, opened 2019-02-11
Remove session and security roles from freemarker vars

amontenegro:


amontenegro:
[Remove freemarker session and security code from the menu and the id banner](https://trello.com/c/VxKxQxQI/5682-remove-freemarker-session-and-security-code-from-the-menu-and-the-id-banner)

amontenegro:
Hi @lizkrznarich could you please take a look at this PR or assign it to @leomendoza123 for review?

Thanks

orcid-web/src/main/webapp/static/javascript/ng1Orcid/app/shared/person.service.ts
```diff
@@ -1,30 +1,43 @@
...
-        });
-      this.endpointWasCall = true;
-      return this.response.asObservable();
+  getPerson() {      
+    if (!this.endpointWasCall) {        
+        this.commonSrvc.userInfo$.subscribe(
+                data => {
+                    this.userInfo = data;
+                    this.personEndpoint = this.http
+                    .get(this.url + '/' + this.userInfo['REAL_USER_ORCID'] + this.path)
```
leomendoza123: This might work fine. The problems is that two serial calls https (meaning the second one waiting for the first one) is going to add some time before the first contentful paint. 

On the prototype I am directly reading the url to quickly get the Orcid id

https://github.com/ORCID/orcid-angular/blob/master/src/app/profile/pages/profile-page/profile-page.component.ts#L25-L32

orcid-web/src/main/webapp/static/javascript/ng1Orcid/app/shared/common.service.ts
```diff
@@ -6,23 +6,31 @@ import { Injectable } from '@angular/core';
...
 
-    public orgDisambiguatedDetails: any;
-
+    public orgDisambiguatedDetails: any;    
+    public userInfo$: Observable<any>;
+    
     constructor(
         private http: HttpClient
     ) {
+        console.log('common.service constructor')
```
leomendoza123: console.log 😄

orcid-web/src/main/webapp/static/javascript/ng1Orcid/app/shared/person.service.ts
leomendoza123: Seems like you are doing something similar here https://github.com/ORCID/ORCID-Source/pull/5256/files#diff-19fa4feaac15184e044ab74e4e7ea3fdR73

amontenegro: Hey @leomendoza123, the problem on the person.service.ts is that we don't have the iD in the url, because those are the pages for a logged in user: /my-orcid, /account, etc; so, not sure how to fix that, any suggestion?

orcid-web/src/main/webapp/static/javascript/ng1Orcid/app/shared/common.service.ts
amontenegro: Removed!

orcid-web/src/main/java/org/orcid/frontend/web/controllers/HomeController.java
```diff
@@ -129,5 +139,64 @@ Object getUserStatusJson(HttpServletRequest request, HttpServletResponse respons
...
+            info.put("CLAIMED", String.valueOf(p.getClaimed()));
+            if(p.getPrimaryRecord() != null) {
+                info.put("PRIMARY_RECORD", p.getPrimaryRecord().getId());                
+            }
+            if(!PojoUtil.isEmpty(p.getGroupType())) {
+                info.put("MEMBER_TYPE", p.getGroupType());
+            }
+            
+        } else {
+            info.put("SIGN_IN_MENU", String.valueOf(true));
```
lizkrznarich: @amontenegro not sure we need SIGN_IN_MENU ...we could just check if userInfo.json is empty

orcid-web/src/main/resources/orcid-frontend-security.xml
```diff
@@ -313,6 +315,8 @@
 			access="IS_AUTHENTICATED_ANONYMOUSLY" />
 		<sec:intercept-url pattern="/userStatus.json(\?.*)?"
 			access="IS_AUTHENTICATED_ANONYMOUSLY" />
+		<sec:intercept-url pattern="/userInfo.json(\?.*)?"
+			access="IS_AUTHENTICATED_ANONYMOUSLY" />
```
lizkrznarich: @amontenegro What's the reason for making this public? Is it just to make SIGN_IN_MENU available if the user isn't logged in? If that's the case, I'd suggest just removing SIGN_IN_MENU and making /userInfo.json accessible only for logged in users.

amontenegro: Hi @lizkrznarich! resolved in the last push, now /userInfo.json requires an authenticated user and, the code checks if userInfo map is empty, if so, it shows the login menu.

orcid-web/src/main/webapp/static/javascript/ng1Orcid/app/shared/person.service.ts
amontenegro: Hey @leomendoza123,

I created a new endpoint for logged in users: /person.json

so: /<ORCID>/person.json will fetch the data in the public page while /person.json will fetch the data when the user is logged in.

Please take another look.

Thanks

PR #5257, opened 2019-02-12
Grouping suggestions now only removed from cache when accepted or rej…

bobcaprice:
…ected by user


bobcaprice:
[Return up to 100 suggestions from server](https://trello.com/c/0K6Cwb5J/5658-return-up-to-100-suggestions-from-server)

PR #5258, opened 2019-02-12
Sort works within group on display index and created date

bobcaprice:


bobcaprice:
[Order works in groups with the same display index with earliest create date first](https://trello.com/c/ncskycOX/5624-order-works-in-groups-with-the-same-display-index-with-earliest-create-date-first)

PR #5261, opened 2019-02-14
Remove2 fa and notifications when deactivating

bobcaprice:


bobcaprice:
[Inbox, trusted individuals, and 2FA aren't removed when an account is deactivated (3)](https://trello.com/c/VIJYJyRR/5662-inbox-trusted-individuals-and-2fa-arent-removed-when-an-account-is-deactivated-3)

PR #5263, opened 2019-02-14
Statistics data endpoint

bobcaprice:


bobcaprice:
[Create new endpoint for statistics data currently passed via model attributes](https://trello.com/c/69mK7d4d/5677-create-new-endpoint-for-statistics-data-currently-passed-via-model-attributes)

PR #5264, opened 2019-02-15
Changed method of deleting all notifications for profile; deletes in …

bobcaprice:
…batches from three notification tables


bobcaprice:
[Deactivate doesn't work](https://trello.com/c/VpjqB9qc/5707-deactivate-doesnt-work)

bobcaprice:
[Inbox, trusted individuals, and 2FA aren't removed when an account is deactivated (3)](https://trello.com/c/VIJYJyRR/5662-inbox-trusted-individuals-and-2fa-arent-removed-when-an-account-is-deactivated-3)

PR #5266, opened 2019-02-18
Replaced model data for /unsubscribe with pojo containing data, avail…

bobcaprice:
…able from new endpoint


bobcaprice:
[Create new endpoint for unsubscribe data currently passed via model attributes](https://trello.com/c/8ER0aI2r/5689-create-new-endpoint-for-unsubscribe-data-currently-passed-via-model-attributes)

PR #5268, opened 2019-02-19
Reactivation data now comes from endpoint rather than model attributes

bobcaprice:


bobcaprice:
[Create new endpoint for reactivation data currently passed via model attributes](https://trello.com/c/UgclEWhQ/5687-create-new-endpoint-for-reactivation-data-currently-passed-via-model-attributes)

orcid-web/src/main/webapp/static/javascript/ng1Orcid/app/modules/reactivation/reactivation.component.ts
```diff
@@ -223,6 +240,7 @@ export class ReactivationComponent implements AfterViewInit, OnDestroy, OnInit {
 
     ngOnInit() {
         this.getReactivation(orcidVar.resetParams, '');
+        this.getReactivationData(orcidVar.resetParams);
```
lizkrznarich: @bobcaprice With the changes to PasswordResetController, [orcidVar.resetParams](https://github.com/ORCID/ORCID-Source/blob/master/orcid-web/src/main/resources/freemarker/common/html-head.ftl#L46) is no longer set. Need to do this without using orcidVar.resetParams, and ```orcidVar.resetParams = '${(resetParams)!}';
        orcidVar.emailToReactivate = '${(email)!}';``` should be removed from https://github.com/ORCID/ORCID-Source/blob/master/orcid-web/src/main/resources/freemarker/common/html-head.ftl#L46

bobcaprice: Done

PR #5271, opened 2019-02-20
Token expired no longer used as model attribute

bobcaprice:


bobcaprice:
[Pass reset password tokenExpired state in URL param](https://trello.com/c/K1V0lWOG/5686-pass-reset-password-tokenexpired-state-in-url-param)

PR #5272, opened 2019-02-20
Errors doc

dreamofaredbird:
This is the initial errors doc. I aim to add some more examples now and as we go along.


caoyler:
Looks good. Please remove the errors that are around creating records. Also, can you reproduce each error to verify it is still applicable- there are a few there that I think we no longer use, for example you have three different possible errors when a file can't be found/read, do we still return all of them?

caoyler:
Also, in the intro comments please mention json as well as XML (or just refer to "files" to cover both).

lizkrznarich:
@dreamofaredbird This one has been hanging out for a while. Can you please review and fix so we can merge?

dreamofaredbird:
@lizkrznarich @caoyler I have included some of the changes in this pr https://github.com/ORCID/ORCID-Source/pull/5288 so this can be closed I will create a new pr for the rest of changes requested by Cat but priotised 3.0 docs as this seemed to have urgency..

PR #5275, opened 2019-02-21
Remove delegate state from model

bobcaprice:


bobcaprice:
[Pass admin delegate state in URL param](https://trello.com/c/1HGswUt9/5685-pass-admin-delegate-state-in-url-param)

PR #5276, opened 2019-02-21
Account page not loading

bobcaprice:


bobcaprice:
[/account does not load](https://trello.com/c/NXHHO450/5729-account-does-not-load)

bobcaprice:
[Create new endpoint for unsubscribe data currently passed via model attributes](https://trello.com/c/8ER0aI2r/5689-create-new-endpoint-for-unsubscribe-data-currently-passed-via-model-attributes)

PR #5277, opened 2019-02-21
Left nav was not displaying data

amontenegro:


amontenegro:
[personal info (left column) is missing on public view of records](https://trello.com/c/FZfcV5RQ/5730-personal-info-left-column-is-missing-on-public-view-of-records)

PR #5286, opened 2019-02-25
get member slug in a way that works in all environments

lizkrznarich:


lizkrznarich:
[Get memberSlug from URL path](https://trello.com/c/hr4az5dq/5698-get-memberslug-from-url-path)

PR #5287, opened 2019-02-26
Fix num found in V3+ search

amontenegro:


amontenegro:
[v3.0+ search returns incorrect num-found](https://trello.com/c/XlWDCFvR/5752-v30-search-returns-incorrect-num-found)

PR #5288, opened 2019-02-26
Correcting links, typos etc

dreamofaredbird:
@caoyler please consider this initial commit. I want to review again but I wanted to push my changes so far so you can see working. Thanks.
Mostly I have just corrected typos and added missing 3.0 links. You will see lots of stuff for group-id. This is because I have redirected each readme to the peer review and group id tutorial for consistency.


dreamofaredbird:
@caoyler I reviewed it again and tidyed up a few more bits and removed the record creation stuff from errors. This section (errors) needs more work but the changes can be merged with everything else if you think it is ok.

PR #5290, opened 2019-02-26
Remove signin model attrs

lizkrznarich:


lizkrznarich:
[Remove model attributes from /siginin and /register](https://trello.com/c/mjo01mEN/5692-remove-model-attributes-from-siginin-and-register)

PR #5292, opened 2019-02-27
Remove security questions

bobcaprice:


bobcaprice:
[Remove the Security Question Feature from /Account and also from password reset (3)](https://trello.com/c/HFm4v4wF/5736-remove-the-security-question-feature-from-account-and-also-from-password-reset-3)

PR #5295, opened 2019-02-28
Changed v2 and v3 security managers to allow viewing of private bulk …

bobcaprice:
…work items with read-public token where client is the source


bobcaprice:
[Read multiple works endpoint returns 403 error when reading private works the client is the source of (3)](https://trello.com/c/hKJBhnTp/5628-read-multiple-works-endpoint-returns-403-error-when-reading-private-works-the-client-is-the-source-of-3)

PR #5297, opened 2019-02-28
Add subject-id as one of the possible headers that contains the remot…

amontenegro:
…e user name


amontenegro:
Should move along with https://github.com/ORCID/registry_vagrant/pull/335

This PR will allow us to use https://samltest.id

amontenegro:
[Enable https://samltest.id as an alternative for http://www.testshib.org/ which is disabled](https://trello.com/c/anhVs5yl/5762-enable-https-samltestid-as-an-alternative-for-http-wwwtestshiborg-which-is-disabled)

amontenegro:
[Enable https://samltest.id as an alternative for http://www.testshib.org/ which is disabled](https://trello.com/c/anhVs5yl/5762-enable-https-samltestid-as-an-alternative-for-http-wwwtestshiborg-which-is-disabled)

PR #5299, opened 2019-03-01
Researcher url name now optional for v3 rc1 and rc2

bobcaprice:


bobcaprice:
[3.0 Make researcher-url:url-name field optional](https://trello.com/c/Uqyh8Wdj/5735-30-make-researcher-urlurl-name-field-optional)

PR #5300, opened 2019-03-01
Encoded fields rather than encoding in getter

bobcaprice:


bobcaprice:
[Institutional sign in link account screen doesn't load](https://trello.com/c/SHZbWGWC/5765-institutional-sign-in-link-account-screen-doesnt-load)

PR #5301, opened 2019-03-01
Fix oauth authorize

lizkrznarich:


lizkrznarich:
[OAuth authorize screen is broken](https://trello.com/c/t9vBbIaJ/5763-oauth-authorize-screen-is-broken)

PR #5302, opened 2019-03-01
reset password strip trailing whitespace from email

lizkrznarich:


lizkrznarich:
[Emails with trailing spaces fail to send reset password email](https://trello.com/c/5uml00R0/5764-emails-with-trailing-spaces-fail-to-send-reset-password-email)

PR #5303, opened 2019-03-01
fix issue with brackets in properties on /account

lizkrznarich:


lizkrznarich:
[Account Settings page doesn't load in Japanese or Korean](https://trello.com/c/RwsR77ER/5768-account-settings-page-doesnt-load-in-japanese-or-korean)

PR #5304, opened 2019-03-04
Check for null to avoid 500 error when suggestions toggled on after w…

bobcaprice:
…ork groups generated


bobcaprice:
[handle 500 error when turning grouping suggestions toggle off](https://trello.com/c/fDoT1Ed6/5767-handle-500-error-when-turning-grouping-suggestions-toggle-off)

PR #5306, opened 2019-03-04
Removed unused model attributes

bobcaprice:


bobcaprice:
[Remove unused model attributes from controllers](https://trello.com/c/CBqkrwFj/5699-remove-unused-model-attributes-from-controllers)

PR #5308, opened 2019-03-04
Update to angular 6 services provider

leomendoza123:


leomendoza123:
[Register Angular services in root injector to avoid creating duplicates](https://trello.com/c/4JKCK6WW/5691-register-angular-services-in-root-injector-to-avoid-creating-duplicates)

PR #5309, opened 2019-03-04
Remove .ts extension for imports

leomendoza123:
Remove .ts extension for imports. 

This was a change introduced in 2016 
https://github.com/Microsoft/TypeScript/issues/11901

Do not represent a big improvement but makes easier the coding process in VS code with autocompletion and import suggestions. 

image


leomendoza123:
[Register Angular services in root injector to avoid creating duplicates](https://trello.com/c/4JKCK6WW/5691-register-angular-services-in-root-injector-to-avoid-creating-duplicates)

PR #5311, opened 2019-03-06
Fixed history source element for api versions 2.x and 3.x

bobcaprice:


bobcaprice:
[Fix how we return the source of records created via the API in 3.0](https://trello.com/c/nsuQMgwd/5733-fix-how-we-return-the-source-of-records-created-via-the-api-in-30)

PR #5315, opened 2019-03-07
Correct client source bad data

bobcaprice:


bobcaprice:
[script for correcting data regarding client sources](https://trello.com/c/jeUYuQdx/5777-script-for-correcting-data-regarding-client-sources)

PR #5316, opened 2019-03-07
Send all requets to userStatus.json in https, prevent delegators-and-…

amontenegro:
…me to run in local page


amontenegro:
[http versions of public pages keep reloading](https://trello.com/c/fdhVZZUr/5776-http-versions-of-public-pages-keep-reloading)

PR #5319, opened 2019-03-08
Queries changed to take account of public clients

bobcaprice:


bobcaprice:
[script for correcting data regarding client sources](https://trello.com/c/jeUYuQdx/5777-script-for-correcting-data-regarding-client-sources)

PR #5321, opened 2019-03-11
Param justRegistered no longer added to all redirect urls after regis…

bobcaprice:
…tering


bobcaprice:
[Authorization fails on new registration if code type follows the redirect URI in the authorization URL](https://trello.com/c/STBZEh4x/5785-authorization-fails-on-new-registration-if-code-type-follows-the-redirect-uri-in-the-authorization-url)

PR #5324, opened 2019-03-12
Display my iD in HTTPS

leomendoza123:


leomendoza123:
[Display my iD in HTTPS (0.5)](https://trello.com/c/NwFY52t4/5719-display-my-id-in-https-05)

PR #5326, opened 2019-03-13
Support IE promises

leomendoza123:


leomendoza123:
[Unable to sign in using ie11 on both QA and Sandbox](https://trello.com/c/NdvEePv3/5797-unable-to-sign-in-using-ie11-on-both-qa-and-sandbox)

PR #5327, opened 2019-03-14
Update work identifier relationships helper

leomendoza123:


leomendoza123:
[Update the help text in the registry for work identifier relationships (0.5)](https://trello.com/c/w8JDA0Ci/5720-update-the-help-text-in-the-registry-for-work-identifier-relationships-05)

PR #5331, opened 2019-03-15
Get gaString from ng component

leomendoza123:


leomendoza123:
[Google Analytics OAuth signin event is tracked as Website signin (5)](https://trello.com/c/gJaoro6o/5801-google-analytics-oauth-signin-event-is-tracked-as-website-signin-5)

PR #5333, opened 2019-03-18
Fix public page for locked, disabled and deprecated users

leomendoza123:


leomendoza123:
[On prod and QA, when visiting the public page on a locked, deactivated or deprecated record you get a server error instead of the correct message.](https://trello.com/c/dk4VPHgi/5802-on-prod-and-qa-when-visiting-the-public-page-on-a-locked-deactivated-or-deprecated-record-you-get-a-server-error-instead-of-the)

lizkrznarich:
@leomendoza123 Please check the failing tests

PR #5334, opened 2019-03-19
Add verification email buttons

leomendoza123:


leomendoza123:
[Reminder to verify email, make link to verify button like we do on initial registration email (1)](https://trello.com/c/Ows9tSCS/5806-reminder-to-verify-email-make-link-to-verify-button-like-we-do-on-initial-registration-email-1)

orcid-core/src/main/resources/i18n/email_en.properties
```diff
@@ -88,7 +88,7 @@ email.verify.2=, and your full ORCID iD and the link to your public record is
 email.verify.primary_email_1=(primary email address:
 email.verify.primary_email_2=)
 email.verify.if_you_did_not=If you did not add this email address to your ORCID record, please contact us immediately by replying to this email.
-
+email.verify.button=Verify your email address
```
lizkrznarich: @leomendoza123 Please use existing string [email.button](https://github.com/ORCID/ORCID-Source/blob/d6d586ee699ee9a0cd328182c57045f8195b0715/orcid-core/src/main/resources/i18n/email_en.properties#L498) since it has already been translated. New strings appear in English until translations are updated in Transifex.

PR #5336, opened 2019-03-19
Simplify add works and affiliations button

leomendoza123:


leomendoza123:
[Remove dropdowns where there is only the "add manually" option (1)](https://trello.com/c/TIK0p3ge/5805-remove-dropdowns-where-there-is-only-the-add-manually-option-1)

PR #5337, opened 2019-03-20
Fix2 fa qr code

lizkrznarich:


lizkrznarich:
[2FA Barcode is missing when trying to enable 2FA in /account](https://trello.com/c/LtZmNjYb/5815-2fa-barcode-is-missing-when-trying-to-enable-2fa-in-account)

PR #5343, opened 2019-03-21
hide user dev tools menu link on member api client records

lizkrznarich:


lizkrznarich:
[both developer tools headers are shown on menu when logged into a group id](https://trello.com/c/9xyqcRRg/5782-both-developer-tools-headers-are-shown-on-menu-when-logged-into-a-group-id)

PR #5344, opened 2019-03-23
expand work heading space regardless of togglz state

lizkrznarich:


lizkrznarich:
[Works info question mark icon is not in the right place ( prod)](https://trello.com/c/bRs9S598/5756-works-info-question-mark-icon-is-not-in-the-right-place-prod)

PR #5347, opened 2019-03-24
Verification email buttons

leomendoza123:


leomendoza123:
[Reminder to verify email, make link to verify button like we do on initial registration email (1)](https://trello.com/c/Ows9tSCS/5806-reminder-to-verify-email-make-link-to-verify-button-like-we-do-on-initial-registration-email-1)

orcid-core/src/main/resources/i18n/email_en.properties
```diff
@@ -79,7 +79,7 @@ email.common.reset_password.href=https://orcid.org/reset-password
 email.verify.primary_reminder_v2=This is a reminder that you need to verify your primary email address before you can begin adding information such as your affiliation, biography, or keywords manually to your ORCID record.
 email.verify.primary_reminder=This is a reminder that you need to verify your primary email address before you can begin adding information manually to your ORCID record.
 email.verify.click_link=Click the following link and sign into your ORCID record to verify your email address. If you can't click on the link, copy and paste it into your browser’s address bar:
-email.verify.thank_you=To verify your email address, click the following link and sign into your ORCID record. If you can't click the link, copy and paste it into your browser's address bar:
+email.verify.thank_you=To verify your email address, click the following button and sign into your ORCID record. If you can't click the button, copy and paste the link below into your browser's address bar:
```
lizkrznarich: @leomendoza123 Users don't see the button in the plain text version, see please keep the original email.verify.thank_you property (with 'click the following link') and use it in the [plain text template](https://github.com/ORCID/ORCID-Source/blob/master/orcid-core/src/main/resources/org/orcid/core/template/verification_email_v2.ftl). Add a new property for the 'click the following button' version and use it in the [HTML template]https://github.com/ORCID/ORCID-Source/blob/master/orcid-core/src/main/resources/org/orcid/core/template/verification_email_html_v2.ftl).

Also, please remove the conditional ```<#if isReminder?? && isReminder>                    
                    <@emailMacros.msg "email.verify.click_link" />
                <#else>
                    <@emailMacros.msg "email.verify.thank_you" />
                </#if>``` from both templates, since ```since email.verify.click_link``` and ```email.verify.thank_you``` say almost exactly the same thing neither is specific to the reminder message. Instead, use only the existing ```email.verify.thank_you``` in the plain text template and the new property mentioned above in the HTML template.

PR #5348, opened 2019-03-24
Update singup

leomendoza123:
Currently, the following endpoints keep using the same non-immediate validation for passwords

/change-password.json
/reset-password.json
/reactivationConfirm.json, /shibboleth/reactivationConfirm.json

This is because this endpoint used different validation methods (until they are posted). 
We might add this the "life" validation to this endpoints as well but we would need to create some new endpoints.


leomendoza123:
[Optimize registration form (3)](https://trello.com/c/Rf6GXk2h/5804-optimize-registration-form-3)

lizkrznarich:
@leomendoza123 Please fix merge conflict

PR #5355, opened 2019-03-29
Change url encoder so it doesnt encode path

amontenegro:


amontenegro:
Hi @TomDemeranville, 

Could you please take a look at this small PR? UriUtils.encodePath will encode the elements in the path, but not the path separator '/', so, something like http://orcid.org/this/is/a/path will not be encoded, however: 

http://orcid.org/<alert>this</alert>/other/path 

will be encoded as  

http://orcid.org/%3Calert%3Ethis%3C/alert%3E/other/path

Thanks

amontenegro:
[Unable to import work using specific IDs](https://trello.com/c/uWHBWSep/5787-unable-to-import-work-using-specific-ids)

PR #5356, opened 2019-03-29
Remove arrow function from no transpiled script

leomendoza123:


leomendoza123:
[https://qa.orcid.org/signin on IE 11 does not load](https://trello.com/c/iwCBaLiF/5833-https-qaorcidorg-signin-on-ie-11-does-not-load)

PR #5363, opened 2019-04-04
bibtex import error: make text wrap

lizkrznarich:


lizkrznarich:
[More Characters in the bibtex error description would be extra helpful](https://trello.com/c/19UIwKS2/5832-more-characters-in-the-bibtex-error-description-would-be-extra-helpful)

PR #5366, opened 2019-04-08
Email trickle debug

bobcaprice:


bobcaprice:
[Sending the member email as a trickle fails to send mails successfully](https://trello.com/c/SBGDF7gr/5860-sending-the-member-email-as-a-trickle-fails-to-send-mails-successfully)

PR #5367, opened 2019-04-08
User status reload when logged in false

lizkrznarich:


lizkrznarich:
[Login check no longer redirects to /siginin when loggedIn = false](https://trello.com/c/Hd1L6Jy3/5859-login-check-no-longer-redirects-to-siginin-when-loggedin-false)

PR #5368, opened 2019-04-08
New header

leomendoza123:


leomendoza123:
[Update header layout on Registry (5)](https://trello.com/c/Z6MVdyxE/3461-update-header-layout-on-registry-5)

orcid-web/src/main/resources/freemarker/includes/ng2_templates/header2-ng2-template.ftl
```diff
@@ -0,0 +1,224 @@
+<script type="text/ng-template" id="header2-ng2-template">
+
+<div class="header2">
+    <#if oauthError??>            
+    <ng-container *ngIf="isOauth">
```
leomendoza123: @lizkrznarich I don't believe we are using ftl variable anymore but is used on the current header

https://github.com/ORCID/ORCID-Source/blob/8b2d9092c2ae1064f0ba46dfaadeade080cfac4a/orcid-web/src/main/resources/freemarker/includes/ng2_templates/header-ng2-template.ftl#L3

Can you please confirm if we don't need this header part for Oauth

lizkrznarich: @leomendoza123 Yes, the header is not used on OAuth signin and authorize screens (just the ORCID logo is shown at the top of the screen)

orcid-core/src/main/java/org/orcid/core/togglz/Features.java
```diff
@@ -90,7 +93,6 @@
     @Label("Grouping suggestions")
     GROUPING_SUGGESTIONS,
     
-    @Label("Enable promotion of chosen orgs in search")
```
lizkrznarich: @leomendoza123 Was this label intentionally removed?

leomendoza123: Thanks Liz I add it back

PR #5371, opened 2019-04-09
break long source names at end of word instead of any character

lizkrznarich:


lizkrznarich:
[Don't break source names midword (1)](https://trello.com/c/c3Ybf78H/5861-dont-break-source-names-midword-1)

PR #5372, opened 2019-04-10
Fix wobble loading spinner on firefox for mac

leomendoza123:


leomendoza123:
[FF loading arrows spin slights off center](https://trello.com/c/mBE6Zc9p/5862-ff-loading-arrows-spin-slights-off-center)

PR #5373, opened 2019-04-10
Corrected test queue name

bobcaprice:


bobcaprice:
[ViewMQStatistics is showing stats for broker and not queue](https://trello.com/c/OjGmUT0Z/5863-viewmqstatistics-is-showing-stats-for-broker-and-not-queue)

PR #5376, opened 2019-04-12
Research resource sorting

lizkrznarich:


lizkrznarich:
[Research-resources with start but no end date sort to the bottom of the list chronologically](https://trello.com/c/cqaQTqqV/5834-research-resources-with-start-but-no-end-date-sort-to-the-bottom-of-the-list-chronologically)

PR #5378, opened 2019-04-12
Encapsulate header CSS

leomendoza123:


leomendoza123:
[Something odd is happening when hovering over any of the icons on the page. A small triangle appears.](https://trello.com/c/pwrGFdl2/5865-something-odd-is-happening-when-hovering-over-any-of-the-icons-on-the-page-a-small-triangle-appears)

PR #5379, opened 2019-04-12
Fixed ng2 maintenance message.

wjrsimpson:


lizkrznarich:
[Maintenance message does not appear when set](https://trello.com/c/41NeUWT8/5869-maintenance-message-does-not-appear-when-set)

PR #5382, opened 2019-04-15
We need to cast the authentication object, so, we can detect when it …

amontenegro:
…is a obo request or a token request (UsernamePasswordAuthenticationToken vs OAuth2Authentication)


amontenegro:
[Using the member API endpoint to generate a token returns a 500 error](https://trello.com/c/Cf2GyGyD/5867-using-the-member-api-endpoint-to-generate-a-token-returns-a-500-error)

PR #5383, opened 2019-04-15
Remove 'mailto' from support URLs

leomendoza123:


leomendoza123:
[Link to support from verify pop-up is broken](https://trello.com/c/m0kDmWCx/5877-link-to-support-from-verify-pop-up-is-broken)

PR #5387, opened 2019-04-15
Support email urls

leomendoza123:


leomendoza123:
[Email that your primary email has changed doesn't have clickable links](https://trello.com/c/uHDFck8H/5878-email-that-your-primary-email-has-changed-doesnt-have-clickable-links)

PR #5388, opened 2019-04-16
Removed unverified emails from public page and both APIs, fixed / upd…

bobcaprice:
…ated tests


bobcaprice:
[Unverified emails should not be shown on public page (5)](https://trello.com/c/vVmgmqUk/5715-unverified-emails-should-not-be-shown-on-public-page-5)

PR #5394, opened 2019-04-19
add detail to notes about enums moved from schema to code in v3.0_rc2 and v3.0

lizkrznarich:


lizkrznarich:
[Notes about enums in v3 schema are not correct](https://trello.com/c/crYwHWEP/5885-notes-about-enums-in-v3-schema-are-not-correct)

PR #5395, opened 2019-04-19
change work:contributor-attributes to common:contributor-attributes in v3.0_rc2 and v3.0

lizkrznarich:


lizkrznarich:
[Enum not removed from contributor role in work_3.0.xml](https://trello.com/c/dTZH0Swm/5886-enum-not-removed-from-contributor-role-in-work30xml)

PR #5397, opened 2019-04-22
Screen reader support my orcid

leomendoza123:


leomendoza123:
[Improve accessibility with screen readers (JAWS 2018) (8)](https://trello.com/c/45mrmrfy/5820-improve-accessibility-with-screen-readers-jaws-2018-8)

PR #5398, opened 2019-04-22
Fix research resource endpoint

amontenegro:


amontenegro:
[3.0_rc1 and rc2 Get research-resources, emails and record endpoints return a 500 error](https://trello.com/c/n6dqGyb6/5887-30rc1-and-rc2-get-research-resources-emails-and-record-endpoints-return-a-500-error)

PR #5402, opened 2019-04-23
add cors config to orcid-frontend-servlet to allow requests from orcid-angular

lizkrznarich:


lizkrznarich:
[CORS config for orcid-angular](https://trello.com/c/4fuf9q39/5892-cors-config-for-orcid-angular)

PR #5404, opened 2019-04-23
show error msg if unverified email set to public/limited

lizkrznarich:


lizkrznarich:
[Add warning to email addresses that are unverified and public or limited](https://trello.com/c/G5dWwgJm/5889-add-warning-to-email-addresses-that-are-unverified-and-public-or-limited)

PR #5410, opened 2019-04-24
Fix help icon positions

leomendoza123:


leomendoza123:
[The question mark Icon is in the wrong place on works and research resources.](https://trello.com/c/Fp0jtgYB/5898-the-question-mark-icon-is-in-the-wrong-place-on-works-and-research-resources)

PR #5422, opened 2019-04-26
Added db script to update uri identifier type to be case sensitive

bobcaprice:


bobcaprice:
[URL normalizer, don't make URLs lower case (0.5)](https://trello.com/c/GPxa5DYi/5871-url-normalizer-dont-make-urls-lower-case-05)

PR #5423, opened 2019-04-26
Reason locked now shown with locked account status in admin page

bobcaprice:


bobcaprice:
[Admin: Show reason for locked accounts in admin tool (2)](https://trello.com/c/qOzTSKKe/5884-admin-show-reason-for-locked-accounts-in-admin-tool-2)

PR #5425, opened 2019-04-30
show unverified emails on public page if togglz feature disabled

lizkrznarich:


lizkrznarich:
[Unverified public emails are not showing as public on the public page when they should. (toggled)](https://trello.com/c/Z3OA6r15/5915-unverified-public-emails-are-not-showing-as-public-on-the-public-page-when-they-should-toggled)

PR #5426, opened 2019-04-30
Speed up source correction script

bobcaprice:


bobcaprice:
[Speed up source correction script](https://trello.com/c/UP9nyMV3/5916-speed-up-source-correction-script)

PR #5428, opened 2019-05-01
fix missing var in delegators

lizkrznarich:


lizkrznarich:
[Errors when reading emails with API](https://trello.com/c/TtJldmty/5918-errors-when-reading-emails-with-api)

PR #5429, opened 2019-05-01
fix another missing var in delegators

lizkrznarich:


lizkrznarich:
[Unverified public emails are not showing as public on the public page when they should. (toggled)](https://trello.com/c/Z3OA6r15/5915-unverified-public-emails-are-not-showing-as-public-on-the-public-page-when-they-should-toggled)

lizkrznarich:
[Errors when reading emails with API](https://trello.com/c/TtJldmty/5918-errors-when-reading-emails-with-api)

PR #5432, opened 2019-05-03
Fix reset password error

leomendoza123:


leomendoza123:
[Password reset page is showing ticks for the password even though nothing has been entered](https://trello.com/c/k0kLnYom/5929-password-reset-page-is-showing-ticks-for-the-password-even-though-nothing-has-been-entered)

PR #5434, opened 2019-05-07
Change break-all to break-word and add hyphens when possible

leomendoza123:


leomendoza123:
[Biography section text breaks in the middle of words](https://trello.com/c/iwRLNSsM/5931-biography-section-text-breaks-in-the-middle-of-words)

PR #5439, opened 2019-05-13
Create iss ns according to client requests

bobcaprice:


bobcaprice:
[Use ISSN data for new peer review group IDs (5)](https://trello.com/c/W4InlJ2p/5919-use-issn-data-for-new-peer-review-group-ids-5)

PR #5440, opened 2019-05-14
Fix obo member admin tags

leomendoza123:


leomendoza123:
[Toggling Member OBO in the UI seems to enable user OBO whereas User OBO appears to do nothing](https://trello.com/c/oOaE0JCo/5943-toggling-member-obo-in-the-ui-seems-to-enable-user-obo-whereas-user-obo-appears-to-do-nothing)

PR #5443, opened 2019-05-21
Issn script

bobcaprice:


bobcaprice:
[Ingest ISSN public data for existing peer review group ids (8)](https://trello.com/c/aZxFAD2R/5920-ingest-issn-public-data-for-existing-peer-review-group-ids-8)

PR #5444, opened 2019-05-22
Remove invalid source validations

leomendoza123:


leomendoza123:
[When User OBO is toggled on when I post a work it still shows the client as the source only not 'client' on behalf of 'orcid'.](https://trello.com/c/SWPkFS9Z/5949-when-user-obo-is-toggled-on-when-i-post-a-work-it-still-shows-the-client-as-the-source-only-not-client-on-behalf-of-orcid)

leomendoza123:
This PR should solve recently report problems. But to avoid future problems I want to make sure that the validations on the code to show or hide OBO are ok. 


The following are the responses from the backend that I can see for a source in different scenarios. 

## Scenario 1: 

An APP adds an element but User OBO is disable
or 
An APP adds an element while user OBO was enabled but it was disabled again

```
          "source": "APP-OQYKBIN9CXD9U3SH",
          "sourceName": "Demo Org QA ",
          "assertionOriginOrcid": null,
          "assertionOriginClientId": null,
          "assertionOriginName": null
``` 
	  

## Scenario 2:

An APP adds an element while user OBO was enabled and is currently enabled

```
          "source": "APP-OQYKBIN9CXD9U3SH",
          "sourceName": "Demo Org QA ",
          "assertionOriginOrcid": "0000-0002-4575-651X",
          "assertionOriginClientId": null,
          "assertionOriginName": "Rob Blackburn"
```

## Scenario 3:

A "slave" app adds an element through a "master" app token 
(this one won't change even if the slave or master app have OBO flags enable or disable after added).


```
          "source": "APP-HAYMC6PX5GI1YSM6",
          "sourceName": "Token exchange client",
          "assertionOriginOrcid": null,
          "assertionOriginClientId": "APP-CY6IU882C8WLCEVB",
          "assertionOriginName": "Oct 24 member client",
```
  

## Scenario 4:

A user adds an element manually 

```
            "sourceName": "Ang Li",
            "source": "0000-0002-4575-651X",
            "assertionOriginOrcid": null,
            "assertionOriginClientId": null,
            "assertionOriginName": null,
```

## The fourth scenarios in one element (after this PR) look like this: 

image


So to me, it makes sense if some `assertion..` parameter is not null the OBO should be shown on the UI no matter what. 

But currently on the UI, is set to do the following validations, which I am beginning to believe are not necessary: 


#### Validation 1
If the `source === assertionOriginOrcid` the OBO is not shown, else `"<sourceName> on behalf of <assertionOriginName>"` is shown

This would mean that an OrcidId can be equal to an app id

#### Validation 2 

If the `source === assertionOriginClientId`  the OBO is not shown, else `"<sourceName> on behalf of <assertionOriginName>"` is shown

This would mean that a slave API is adding an element using itself as a master API 



@lizkrznarich @bobcaprice Do you see a scenario where we need those validations?

lizkrznarich:
@leomendoza123 Scenarios 1-4 are correct. Validation 1 is needed, because there was a time when a user's ORCID iD was used as their public client iD (client iD format was changed for all new clients in 2014 https://trello.com/c/TSSw8mPl/698-done-here-as-a-reminder-all-new-clientids-are-non-orcid-alphanumeric-ids , but existing client ids were not changed). Validation 2 should not be needed; if that case occurs, it would likely be due to a mistake made by an admin.

PR #5451, opened 2019-05-27
Clear null fields on mapperfactory

amontenegro:


amontenegro:
[When updating an affiliation with the api using PUT, if something is removed then the corresponding work does not get updated.](https://trello.com/c/LkBZqGtM/5947-when-updating-an-affiliation-with-the-api-using-put-if-something-is-removed-then-the-corresponding-work-does-not-get-updated)

PR #5453, opened 2019-05-30
Loosened regex that recognises group ids that are meant to be issn gr…

bobcaprice:
…oups


bobcaprice:
[Use ISSN data for new peer review group IDs (5)](https://trello.com/c/W4InlJ2p/5919-use-issn-data-for-new-peer-review-group-ids-5)

PR #5462, opened 2019-06-10
Break duplicateGroup filtering if the current item has been filtered …

amontenegro:
…already


amontenegro:
[Peer review filtering on V3 causes illegal state exception errors](https://trello.com/c/yyxAVW6W/5993-peer-review-filtering-on-v3-causes-illegal-state-exception-errors)

PR #5463, opened 2019-06-11
OBO tokens filtered out from trusted apps

bobcaprice:


bobcaprice:
[Prevent OBO client 2 from appearing in the trusted source list (3)](https://trello.com/c/Pba8GzlT/5965-prevent-obo-client-2-from-appearing-in-the-trusted-source-list-3)

PR #5464, opened 2019-06-11
Added prefix to handle identifier type

bobcaprice:


bobcaprice:
[Works: Present HANDLE as links in the user interface (even when user doesn't include identifier URL) (2)](https://trello.com/c/S59vOPZd/5967-works-present-handle-as-links-in-the-user-interface-even-when-user-doesnt-include-identifier-url-2)

PR #5467, opened 2019-06-11
Fix typeahead not showing same name organizations

leomendoza123:


leomendoza123:
[University not showing in dropdown](https://trello.com/c/fGuPCvqG/5983-university-not-showing-in-dropdown)

PR #5468, opened 2019-06-13
Funding sub types should be html scaped before going to solr

amontenegro:


amontenegro:
[Weird error when trying to add a funding subtype](https://trello.com/c/GkbtYLhe/5999-weird-error-when-trying-to-add-a-funding-subtype)

PR #5469, opened 2019-06-14
User menu

leomendoza123:


leomendoza123:
[Move Sign in/Register and signed in options out of the second level navigation and in a user menu in the top right corner (13)](https://trello.com/c/ax7A6H2Q/5956-move-sign-in-register-and-signed-in-options-out-of-the-second-level-navigation-and-in-a-user-menu-in-the-top-right-corner-13)

PR #5473, opened 2019-06-14
Activities now hidden for locked records

bobcaprice:


bobcaprice:
[Don't display activities section on locked records](https://trello.com/c/594dMtvE/6001-dont-display-activities-section-on-locked-records)

PR #5476, opened 2019-06-18
Peer review subject type is not a required field, so, it could be null

amontenegro:


amontenegro:
[3.0_rc1 peer-review json put 500 error](https://trello.com/c/uMGBcpaP/6012-30rc1-peer-review-json-put-500-error)

PR #5477, opened 2019-06-19
Fix social sign in link for accounts that already exists

amontenegro:


amontenegro:
[WHen Signing in Via Google I can connect my account but when I go to sign in with google again I have to reconnect.](https://trello.com/c/xesmUyFL/6014-when-signing-in-via-google-i-can-connect-my-account-but-when-i-go-to-sign-in-with-google-again-i-have-to-reconnect)

PR #5479, opened 2019-06-20
New header translation sanitization

leomendoza123:


leomendoza123:
[The new menu is a bit broken en Español ( see screenshot)](https://trello.com/c/PXT9hprM/6022-the-new-menu-is-a-bit-broken-en-espa%C3%B1ol-see-screenshot)

PR #5481, opened 2019-06-24
Update header second level button hover state color

leomendoza123:


leomendoza123:
[Hover Colour is not Gray rgba(98, 2, 238, 0) it seems to be a lot lighter. See screenshot.](https://trello.com/c/RDMIDY2a/6026-hover-colour-is-not-gray-rgba98-2-238-0-it-seems-to-be-a-lot-lighter-see-screenshot)

PR #5482, opened 2019-06-24
Fix wrong uppercase function scope

leomendoza123:


leomendoza123:
[Membership has a lower case 'p' on third level menu item](https://trello.com/c/Mb3UD0ze/6025-membership-has-a-lower-case-p-on-third-level-menu-item)

PR #5489, opened 2019-07-01
Fix hidden advanced search button

leomendoza123:


leomendoza123:
[Advanced search button is partially hidden](https://trello.com/c/sodC7Yiq/6042-advanced-search-button-is-partially-hidden)

PR #5490, opened 2019-07-01
Add wider grid togglz option

leomendoza123:


leomendoza123:
[Improve UI similitude between apps](https://trello.com/c/bhC9bqMn/6043-improve-ui-similitude-between-apps)

PR #5491, opened 2019-07-02
Null ext ids grouping suggestion bug

bobcaprice:


bobcaprice:
[Works not shown if work with null external ids is included in grouping suggestions](https://trello.com/c/EdmjeGlV/6044-works-not-shown-if-work-with-null-external-ids-is-included-in-grouping-suggestions)

lizkrznarich:
@bobcaprice Can you please have a look at the conflict?

PR #5495, opened 2019-07-04
Fix italian translation

gmejias:
Remove "..." on both strings
@leomendoza123


lizkrznarich:
[other languages break the header at both 16px and 18px](https://trello.com/c/p0vPmm3M/6024-other-languages-break-the-header-at-both-16px-and-18px)

PR #5500, opened 2019-07-08
Sign in hidden or activated depending on user menu togglz state

leomendoza123:


leomendoza123:
[Remove Sign In button when user menu is activated](https://trello.com/c/ipWrr0HZ/6036-remove-sign-in-button-when-user-menu-is-activated)

PR #5501, opened 2019-07-08
Fix lower case mobile menu items

leomendoza123:


leomendoza123:
[The mobile menu has a stray {{ in it.](https://trello.com/c/W0m8xUpQ/6053-the-mobile-menu-has-a-stray-in-it)

PR #5502, opened 2019-07-08
Add transifex single source push docs

leomendoza123:


leomendoza123:
[other languages break the header at both 16px and 18px](https://trello.com/c/p0vPmm3M/6024-other-languages-break-the-header-at-both-16px-and-18px)

PR #5506, opened 2019-07-09
Fix test translations values

leomendoza123:


leomendoza123:
[Add language X on ngOrcid](https://trello.com/c/8cV5iSYG/6056-add-language-x-on-ngorcid)

PR #5507, opened 2019-07-09
Fix loggin width

leomendoza123:


leomendoza123:
[Signin form is too wide with increased grid size](https://trello.com/c/h7PrJi6c/6060-signin-form-is-too-wide-with-increased-grid-size)

PR #5508, opened 2019-07-10
Fix spanish home page translation

leomendoza123:
@gmejias Hello, I am proposing this fix on Spanish since it is currently missing the number of steps 

image


lizkrznarich:
@leomendoza123 @gmejias This should be updated in Transifex, not in Github - if it's only updated in Github, the changes will not be reflected in Transifex.

PR #5509, opened 2019-07-11
Removed integration url from public member details page.

wjrsimpson:


wjrsimpson:
[Member List - Missing integration URL](https://trello.com/c/qZFMI8WS/6079-member-list-missing-integration-url)

PR #5516, opened 2019-07-12
set reviewed=true for new member orcid records

lizkrznarich:


lizkrznarich:
[Mark all new client groups as reviewed (1)](https://trello.com/c/ARaUcDUK/6010-mark-all-new-client-groups-as-reviewed-1)

PR #5521, opened 2019-07-15
add noindex flag to qa and sandbox records

lizkrznarich:


lizkrznarich:
[Don't allow Google to index QA and Sandbox (3)](https://trello.com/c/QpdibNAD/6041-dont-allow-google-to-index-qa-and-sandbox-3)

PR #5522, opened 2019-07-16
Fix sign in width

leomendoza123:


leomendoza123:
[Sign-in and register form are mis-aligned on mobile](https://trello.com/c/ZIiyyla7/6089-sign-in-and-register-form-are-mis-aligned-on-mobile)

PR #5523, opened 2019-07-17
Fix mobile login buttons border

leomendoza123:


leomendoza123:
[Styling on the Instititutional and Personal buttons is slightly off](https://trello.com/c/QnOqQxwk/6095-styling-on-the-instititutional-and-personal-buttons-is-slightly-off)

PR #5524, opened 2019-07-17
Fix privacy update for emails

leomendoza123:


leomendoza123:
[Update email visibility doesn't reflect until you reload the page or do it twice](https://trello.com/c/r6R7ZEAh/6096-update-email-visibility-doesnt-reflect-until-you-reload-the-page-or-do-it-twice)

PR #5528, opened 2019-07-19
Check nameform existence before name display

leomendoza123:


leomendoza123:
[New User login: There is no sign out or other user options](https://trello.com/c/gnZTD4FE/6101-new-user-login-there-is-no-sign-out-or-other-user-options)

PR #5532, opened 2019-07-23
Correct bad org data

bobcaprice:


bobcaprice:
[Fix bad org data](https://trello.com/c/V9xJtlJZ/6115-fix-bad-org-data)

PR #5535, opened 2019-07-24
Fix admin delegation

leomendoza123:


leomendoza123:
[User menu doesn't work on admin group delegation](https://trello.com/c/DI8PvYJ9/6116-user-menu-doesnt-work-on-admin-group-delegation)

PR #5540, opened 2019-07-26
Add back the worksCount var

amontenegro:


amontenegro:
[Works list doesn't appear in permission notification on /inbox](https://trello.com/c/DwF3BcDT/6127-works-list-doesnt-appear-in-permission-notification-on-inbox)

PR #5541, opened 2019-07-29
Flip client names for obo source label

leomendoza123:


leomendoza123:
[Clients are in the wrong order in UI with via/OBO](https://trello.com/c/Z5CnWMup/6128-clients-are-in-the-wrong-order-in-ui-with-via-obo)

PR #5542, opened 2019-07-30
Remove tools tips

leomendoza123:


leomendoza123:
[Delete/actions on tablets/phones does not work! (5)](https://trello.com/c/0lzrbO9o/5992-delete-actions-on-tablets-phones-does-not-work-5)

PR #5549, opened 2019-08-06
New work type for physical-object

bobcaprice:


bobcaprice:
[Add "physical-object" as a work type in v3.0 (?) (2)](https://trello.com/c/9HNoXd8c/6111-add-physical-object-as-a-work-type-in-v30-2)

PR #5551, opened 2019-08-06
Hide name for members on the public profile page

leomendoza123:


leomendoza123:
[Newly created client groups should not have the group name display on the public record (1)](https://trello.com/c/s0XvdS4H/4312-newly-created-client-groups-should-not-have-the-group-name-display-on-the-public-record-1)

PR #5552, opened 2019-08-07
Switch account now available on all pages

bobcaprice:


bobcaprice:
[Admin screen: Add switch user option in all pages (3)](https://trello.com/c/bfkY3K6g/6147-admin-screen-add-switch-user-option-in-all-pages-3)

PR #5553, opened 2019-08-08
Accounts can no longer be deprecated into each other

bobcaprice:


bobcaprice:
[Don't allow records to be deprecated into each other (2)](https://trello.com/c/IWqS2tl1/4592-dont-allow-records-to-be-deprecated-into-each-other-2)

PR #5557, opened 2019-08-08
Do not allow same email as password

leomendoza123:


leomendoza123:
[Account Settings: Disallow users from setting their email address as their password (2)](https://trello.com/c/nPjHioZz/6006-account-settings-disallow-users-from-setting-their-email-address-as-their-password-2)

PR #5558, opened 2019-08-09
Restore cron tasks

amontenegro:


amontenegro:
[Verbose notification emails are not being sent](https://trello.com/c/lmkz0bYm/6155-verbose-notification-emails-are-not-being-sent)

PR #5559, opened 2019-08-09
Ignore case when comparing email to reactivate

amontenegro:


amontenegro:
[Reactivate record: compare existing email with provided email case insensitive](https://trello.com/c/b2hflsno/6027-reactivate-record-compare-existing-email-with-provided-email-case-insensitive)

PR #5560, opened 2019-08-09
Ignore counting duplicates, for example, if the member edit an element twice, count it just once

amontenegro:
…t, and then update it, count the element just once


amontenegro:
[[HIDDEN] The Notifications email numbers do not quite work](https://trello.com/c/aoSZNBjJ/6142-hidden-the-notifications-email-numbers-do-not-quite-work)

PR #5561, opened 2019-08-12
Use group name on notification

amontenegro:


amontenegro:
[[HIDDEN] Verbose notification on peer-reviews use group name instead of subject name](https://trello.com/c/UZnNmyIX/6146-hidden-verbose-notification-on-peer-reviews-use-group-name-instead-of-subject-name)

PR #5563, opened 2019-08-14
Fix footer cc0 link accessibility

leomendoza123:


leomendoza123:
[[HIDDEN] Accesibility contrast on search bar and footer](https://trello.com/c/KjcJiDWQ/6102-hidden-accesibility-contrast-on-search-bar-and-footer)

PR #5564, opened 2019-08-14
Fixed error with fundings containing cents in amounts

bobcaprice:


bobcaprice:
[Can't save a funding with an amount that includes cents (5)](https://trello.com/c/OkLTjkM5/6087-cant-save-a-funding-with-an-amount-that-includes-cents-5)

PR #5565, opened 2019-08-14
Self service exclude group ids

lizkrznarich:


lizkrznarich:
[Remove Group/client IDs from displaying in self service (2)](https://trello.com/c/eNPBIy58/6007-remove-group-client-ids-from-displaying-in-self-service-2)

PR #5566, opened 2019-08-15
Fix 4 level menu hover

leomendoza123:


leomendoza123:
[[HIDDEN] Registry 4th level mobile menu hover state is incorrect](https://trello.com/c/TEQM9ROQ/6159-hidden-registry-4th-level-mobile-menu-hover-state-is-incorrect)

PR #5567, opened 2019-08-15
Fix icon role and dropdown structure

leomendoza123:


leomendoza123:
[The AXE plugin is highlighting a missing alt tag on the login icon](https://trello.com/c/Cj1exVdz/6199-the-axe-plugin-is-highlighting-a-missing-alt-tag-on-the-login-icon)

PR #5568, opened 2019-08-15
ORCID iD converted to upper case before seach submission, stopped emp…

bobcaprice:
…ty search each time page renders


bobcaprice:
[Make ORCID iDs case insensitive on website search (1)](https://trello.com/c/9zKuO3lE/6152-make-orcid-ids-case-insensitive-on-website-search-1)

PR #5572, opened 2019-08-19
Add new footer translations

leomendoza123:


leomendoza123:
[In the Firebase app Italian 'About' is not translated.](https://trello.com/c/xYOtiHmX/6201-in-the-firebase-app-italian-about-is-not-translated)

PR #5573, opened 2019-08-19
New field for orgs searching

amontenegro:


amontenegro:
[Organisation typeahead doesn't like non roman characters](https://trello.com/c/TOtJ7R0O/4627-organisation-typeahead-doesnt-like-non-roman-characters)

PR #5575, opened 2019-08-20
Fix password generator

leomendoza123:


leomendoza123:
["Generate password" functionality under /admin-actions sometimes generates invalid passwords (due to not including at least one number) (2)](https://trello.com/c/aSbuD2Ec/6168-generate-password-functionality-under-admin-actions-sometimes-generates-invalid-passwords-due-to-not-including-at-least-one-numb)

PR #5581, opened 2019-08-21
Save the url as a url

leomendoza123:


leomendoza123:
[BibTeX import: if there is a URL but no identifier, save the URL as a URI identifier (2)](https://trello.com/c/x8G40xhy/6217-bibtex-import-if-there-is-a-url-but-no-identifier-save-the-url-as-a-uri-identifier-2)

PR #5582, opened 2019-08-22
Add new stats

lizkrznarich:


lizkrznarich:
[Display all sections (works, resources, etc) on https://orcid.org/statistics (5)](https://trello.com/c/we9mKQvv/6228-display-all-sections-works-resources-etc-on-https-orcidorg-statistics-5)

PR #5584, opened 2019-08-23
Use only one user menu

leomendoza123:


leomendoza123:
[[HIDDEN] Hide 2nd level Registry menu when ENABLE_USER_MENU feature is enabled](https://trello.com/c/PrfNTXJR/6230-hidden-hide-2nd-level-registry-menu-when-enableusermenu-feature-is-enabled)

PR #5585, opened 2019-08-23
Fix hover for open menu items

leomendoza123:


leomendoza123:
[[HIDDEN] Wrong hover state color on active menu items](https://trello.com/c/w9BfIepp/6231-hidden-wrong-hover-state-color-on-active-menu-items)

PR #5592, opened 2019-08-30
Log4j no longer outputs TrickleTooHeavyExceptions, thrown thousands o…

bobcaprice:
…f times by email trickle mechanism


bobcaprice:
[Email trickle logging errors even when there's no active email schedule](https://trello.com/c/csgOj7yN/6246-email-trickle-logging-errors-even-when-theres-no-active-email-schedule)

PR #5598, opened 2019-09-05
PeerReviewSummary now OrganizationHolder, org address fields cleaned …

bobcaprice:
…before return. ActivitiesSummary calls now clean peer review and funding org address fields


bobcaprice:

PR #5599, opened 2019-09-05
Solr search

dreamofaredbird:
I think the examples here already are sufficientbut can add more if you think. 

@amontenegro I have linked to solr 8 docs is that correct?


caoyler:
Maybe add an example for the external identifier type and value search, that's a little more complicated syntax.

dreamofaredbird:
ok thanks @caoyler I am going to delete this branch and do a new one because it has got mixed up with the fundref

PR #5600, opened 2019-09-05
New affiliations now implement organization holder interface and have…

bobcaprice:
… org address fields cleaned on read


bobcaprice:

PR #5602, opened 2019-09-05
Outline user menu button while user is logged

leomendoza123:


leomendoza123:
[New sign in: accessibility, The name of the user is not highlighted when tabbing through the page.](https://trello.com/c/zx4YUfpL/6259-new-sign-in-accessibility-the-name-of-the-user-is-not-highlighted-when-tabbing-through-the-page)

PR #5605, opened 2019-09-09
Org nodes detached from document to improve performance

bobcaprice:


bobcaprice:
[investigate why fundref import is slow](https://trello.com/c/LuViqS4q/6261-investigate-why-fundref-import-is-slow)

PR #5607, opened 2019-09-10
Works pagination now accepts pageSize param from ui

bobcaprice:


bobcaprice:
[Allow specifying works page size](https://trello.com/c/gONYaULv/6269-allow-specifying-works-page-size)

PR #5609, opened 2019-09-10
Remove contributor edit options from work form

leomendoza123:


leomendoza123:
[Role and Credited options show but can't be saved on old manually added works (2)](https://trello.com/c/u1xE9f2w/6274-role-and-credited-options-show-but-cant-be-saved-on-old-manually-added-works-2)

PR #5614, opened 2019-09-16
Make the open_id token to expire in the same time as the access token…

amontenegro:
… that generate it


amontenegro:
[Raise id_token lifetime to 20 years (1)](https://trello.com/c/m9TwqxUm/6253-raise-idtoken-lifetime-to-20-years-1)

PR #5615, opened 2019-09-16
We dont need to upercase the field before searching anymore

amontenegro:


amontenegro:
[advanced Orcid search fails to return anything 'type error...' when you type in the fields. The search for an orcid is ok](https://trello.com/c/x4GYzt3b/6277-advanced-orcid-search-fails-to-return-anything-type-error-when-you-type-in-the-fields-the-search-for-an-orcid-is-ok)

PR #5616, opened 2019-09-16
fix self service org id query

lizkrznarich:


lizkrznarich:
[Unable to add a consortium organization ID in Self Service If I add an LEI nothing happens](https://trello.com/c/SmaJpIvO/6288-unable-to-add-a-consortium-organization-id-in-self-service-if-i-add-an-lei-nothing-happens)

lizkrznarich:
[No org Ids shown in Self service for Org even though they are in Salesforce](https://trello.com/c/ekw8xtE0/6289-no-org-ids-shown-in-self-service-for-org-even-though-they-are-in-salesforce)

PR #5617, opened 2019-09-16
API now handles invalid amounts gracefully

bobcaprice:


bobcaprice:
[Funding amount lets you save invalid values (including text) in the amount field (QA)](https://trello.com/c/b10xt0Ek/6279-funding-amount-lets-you-save-invalid-values-including-text-in-the-amount-field-qa)

PR #5618, opened 2019-09-16
clear funding details UI cache when groups are reloaded

lizkrznarich:


lizkrznarich:
[Funding amount lets you save invalid values (including text) in the amount field (QA)](https://trello.com/c/b10xt0Ek/6279-funding-amount-lets-you-save-invalid-values-including-text-in-the-amount-field-qa)

PR #5619, opened 2019-09-17
Activities buckets prefix should be the last 3 digits of the iD / the…

amontenegro:
… full id


amontenegro:
[Clear activities on S3 V2 and V3 buckets when the record is deactivated, locked, deprecated, etc](https://trello.com/c/zmGyzvLc/6291-clear-activities-on-s3-v2-and-v3-buckets-when-the-record-is-deactivated-locked-deprecated-etc)

PR #5621, opened 2019-09-17
migrate search code from script.js to angular and update to v3.0

lizkrznarich:


lizkrznarich:
[Advanced search is still using v2.1](https://trello.com/c/MPjuT6kx/6278-advanced-search-is-still-using-v21)

PR #5622, opened 2019-09-19
adding text about diacritics and example search for external Ids

dreamofaredbird:


caoyler:
@dreamofaredbird Thanks for working on this. Does only names handle diacritics? I thought it was any named field and only a general text search didn't handle them? Please double check and update if needed.

Please make the example for the identifier search match the other examples. (Use a sandbox URL and don't provide a curl statement). Also make sure the url on sandbox actually returns a search result if someone tests it as is.

I'd missed the peer-review additional fields before- that seems worth adding an example for, at least the searching by goup-id as that is something we get requests for.

I noticed you updated the version on this which is great- are our other tutorials still using 2.0 examples? If so let's get a card up to get everything updated to 3.0.

dreamofaredbird:
@caoyler thanks for the feedback. I put in there pretty much exactly what @amontenegro told me when he explained how it worked to me. @amontenegro if you have a minute could you see if I have got it right? 

I will push a commit with the changes you suggest. Including the group id search. 

I think all the tuts are already 3.0 but I will double check

PR #5627, opened 2019-09-23
Solr search docs

dreamofaredbird:
Removed the incorrect researcherID example (searching with external IDs is already covered) and putting a correct ISSN in the peer-review-group-id example


caoyler:
@dreamofaredbird I did another pr with some final copy-edits. Please review and if it looks good you can merge this.

PR #5628, opened 2019-09-23
add default pageSize to worksPage.json

lizkrznarich:


lizkrznarich:
[Works are not showing in the UI](https://trello.com/c/fj8zTi5u/6297-works-are-not-showing-in-the-ui)

PR #5631, opened 2019-09-24
Add server-side open graph tags and remove JS added tags

leomendoza123:


leomendoza123:
[When posting my ORCID on Facebook it is displayed as deprecated even though it is not. (5)](https://trello.com/c/o4WlKHQ4/6273-when-posting-my-orcid-on-facebook-it-is-displayed-as-deprecated-even-though-it-is-not-5)

PR #5632, opened 2019-09-25
Toggle for making v3.0 default version of both APIs. If not switched …

bobcaprice:
…on, 2.0 is default version.


bobcaprice:
[[HIDDEN] Make v3 the default version of the API](https://trello.com/c/LopgOC1Z/6302-hidden-make-v3-the-default-version-of-the-api)

PR #5633, opened 2019-09-25
Added extra logging for when profile events found

bobcaprice:


bobcaprice:
[Add error handler to email trickle so errors are not output](https://trello.com/c/eKDsGIIi/6300-add-error-handler-to-email-trickle-so-errors-are-not-output)

PR #5634, opened 2019-09-26
fix wrong var in template

lizkrznarich:


lizkrznarich:
[(prod team)Stats on https://orcid.org/statistics seem incorrect](https://trello.com/c/qpYfFPIM/6309-prod-teamstats-on-https-orcidorg-statistics-seem-incorrect)

PR #5636, opened 2019-09-27
Corrected column name in query

bobcaprice:


bobcaprice:
[user obo historical data script fails on keywords](https://trello.com/c/IcMHbm3L/6318-user-obo-historical-data-script-fails-on-keywords)

PR #5637, opened 2019-09-27
Fixed copy paste error

bobcaprice:


bobcaprice:
[Keyword not showing Source when posting via the APi](https://trello.com/c/PA4iN1j5/6319-keyword-not-showing-source-when-posting-via-the-api)

PR #5638, opened 2019-09-27
Fix search query builder and centralize the logic in the search service

leomendoza123:


leomendoza123:
[(QA) Trusted individual fails to load](https://trello.com/c/b5mqaz0U/6316-qa-trusted-individual-fails-to-load)

PR #5639, opened 2019-09-27
Fix advance search base url

leomendoza123:


leomendoza123:
[(qa) Advance search fails](https://trello.com/c/6XU2Qvng/6323-qa-advance-search-fails)

PR #5640, opened 2019-09-29
User obo writes for v2

bobcaprice:


bobcaprice:
[S&L wizards are not posting OBO correctly. (v2 api does not include the correct metadata.)](https://trello.com/c/deVBm5OU/6321-sl-wizards-are-not-posting-obo-correctly-v2-api-does-not-include-the-correct-metadata)

PR #5643, opened 2019-09-30
Fix obo for publicGroupedKeywords

leomendoza123:


leomendoza123:
[UserOBO keyword source wrong on public page](https://trello.com/c/eJ5Pd9ey/6327-userobo-keyword-source-wrong-on-public-page)

PR #5644, opened 2019-10-01
Remove strange user menu logic in the public page

leomendoza123:


leomendoza123:
[(prod)No sign in option when you are visiting a public page (QA)](https://trello.com/c/vI65uZwv/6315-prodno-sign-in-option-when-you-are-visiting-a-public-page-qa)

PR #5645, opened 2019-10-01
Fix padding for non registry pages

leomendoza123:


leomendoza123:
[Left margin is missing on account-settings, inbox and developer tools (QA)](https://trello.com/c/vFQkhjUD/6322-left-margin-is-missing-on-account-settings-inbox-and-developer-tools-qa)

PR #5646, opened 2019-10-01
add spaces to affiliation key so that typeahead generates better results

lizkrznarich:


lizkrznarich:
[Org typeahead returns no relevant results for orgs that definitely exist](https://trello.com/c/NfUuAdjv/6314-org-typeahead-returns-no-relevant-results-for-orgs-that-definitely-exist)

PR #5648, opened 2019-10-02
Revert user obo data

bobcaprice:


bobcaprice:
[Make DataCite added works NOT User OBO](https://trello.com/c/mS2SRJV5/6339-make-datacite-added-works-not-user-obo)

PR #5649, opened 2019-10-02
update search to handle v3 affiliation groups structure

lizkrznarich:


lizkrznarich:
[Prod and QA: Search results do no include affiliations (QA)](https://trello.com/c/b3yorpKv/6341-prod-and-qa-search-results-do-no-include-affiliations-qa)

PR #5650, opened 2019-10-02
Do not allow empty searches 

leomendoza123:


leomendoza123:
[Hitting return in Quicksearch without filling anything in the box results in an endlessly spinning arrow.](https://trello.com/c/31M5ZhKY/6340-hitting-return-in-quicksearch-without-filling-anything-in-the-box-results-in-an-endlessly-spinning-arrow)

PR #5654, opened 2019-10-04
Grouping takes into account null external ids as well as empty list o…

bobcaprice:
…f external ids


bobcaprice:
[Combine works sometimes fails](https://trello.com/c/CfJNMbAW/6342-combine-works-sometimes-fails)

PR #5656, opened 2019-10-07
User obo fields only populated for requests with orcid profile for token

bobcaprice:


bobcaprice:
[When user OBO is enabled- you can't delete a peer-review group you are the source of](https://trello.com/c/lXPFhucs/6325-when-user-obo-is-enabled-you-cant-delete-a-peer-review-group-you-are-the-source-of)

PR #5658, opened 2019-10-08
Prevent NPE to happen when the souce is null

amontenegro:


amontenegro:
[Affiliation issue with 0000-0002-7720-2757](https://trello.com/c/Y2boWgQa/3036-affiliation-issue-with-0000-0002-7720-2757)

PR #5661, opened 2019-10-10
Allow affiliations with end dates only

amontenegro:


amontenegro:
[Allow affiliations with end dates but no start dates (2)](https://trello.com/c/59nkgBLa/6333-allow-affiliations-with-end-dates-but-no-start-dates-2)

PR #5664, opened 2019-10-15
Allow affs with only end date on V3

amontenegro:


amontenegro:
[(QA) When posting with no start date but an end date for affiliations using the 3.0- api I get an error.](https://trello.com/c/HksMmkN5/6360-qa-when-posting-with-no-start-date-but-an-end-date-for-affiliations-using-the-30-api-i-get-an-error)

PR #5667, opened 2019-10-16
Remove solr web

lizkrznarich:


lizkrznarich:
[Check and remove SOLR dependencies that are no longer needed](https://trello.com/c/szeBrrTt/6380-check-and-remove-solr-dependencies-that-are-no-longer-needed)

PR #5668, opened 2019-10-17
Tweak orgs query

lizkrznarich:


lizkrznarich:
[PROD Org type ahead still missing orgs (QA)](https://trello.com/c/Iqse8sPg/6366-prod-org-type-ahead-still-missing-orgs-qa)

PR #5670, opened 2019-10-17
Lets allow the users with no primary email to login, and, if they hav…

amontenegro:
…e any verified email, update their record


amontenegro:
[Allow users with no primary email to login](https://trello.com/c/IkkJyEaR/6373-allow-users-with-no-primary-email-to-login)

PR #5672, opened 2019-10-18
Allow user with no primary email to log in

amontenegro:


amontenegro:
[Allow users with no primary email to login](https://trello.com/c/IkkJyEaR/3041-allow-users-with-no-primary-email-to-login)

PR #5673, opened 2019-10-18
Fix RTL work/affiliation cards

leomendoza123:


leomendoza123:
[Right-to-left languages misalignment on work cards](https://trello.com/c/iu5q7aGj/6368-right-to-left-languages-misalignment-on-work-cards)

PR #5677, opened 2019-10-29
duplicate check populate affiliations in back end instead of using pub api

lizkrznarich:


lizkrznarich:
[Registration duplicate check missing affiliations](https://trello.com/c/lVX5TkwX/6379-registration-duplicate-check-missing-affiliations)

PR #5679, opened 2019-10-31
make disambiguated org query lower case for affilations and funding

lizkrznarich:


lizkrznarich:
[PROD Org type ahead should not be case sensitive (QA)](https://trello.com/c/7dWD6Glb/6367-prod-org-type-ahead-should-not-be-case-sensitive-qa)

PR #5680, opened 2019-11-04
fix sign in link in registration form error msg

lizkrznarich:


lizkrznarich:
[Register form:  sign in link for existing records doesn't work](https://trello.com/c/JMeqXsCR/6384-register-form-sign-in-link-for-existing-records-doesnt-work)

PR #5681, opened 2019-11-04
add toUpperCase transformation back to UI search

lizkrznarich:


lizkrznarich:
[Make ORCID iDs case insensitive on website search (1)](https://trello.com/c/9zKuO3lE/6152-make-orcid-ids-case-insensitive-on-website-search-1)

PR #5691, opened 2019-11-13
Remove sign in url on mobile

leomendoza123:


leomendoza123:
[Sign in/register link on mobile turns into profile icon](https://trello.com/c/x3ihExdg/6397-sign-in-register-link-on-mobile-turns-into-profile-icon)

PR #5699, opened 2019-11-14
Email trickle email exists check

bobcaprice:


bobcaprice:
[Add check in email trickle that emails still exist in system](https://trello.com/c/Tt57SkLC/6403-add-check-in-email-trickle-that-emails-still-exist-in-system)

PR #5701, opened 2019-11-20
change angular build phase from compile to generate-resources

lizkrznarich:


lizkrznarich:
[The deploy process does not copy the most recent orcid-angular build](https://trello.com/c/dZFe4HNQ/6414-the-deploy-process-does-not-copy-the-most-recent-orcid-angular-build)

PR #5702, opened 2019-11-21
Org script fixes

bobcaprice:


bobcaprice:
[Org data correction script falling over due to violation constraint](https://trello.com/c/54sUIR8L/6427-org-data-correction-script-falling-over-due-to-violation-constraint)

PR #5703, opened 2019-11-22
Script for clearing out user obo data from db

bobcaprice:


bobcaprice:
[Clear out user obo columns](https://trello.com/c/dKafVUi9/6351-clear-out-user-obo-columns)

PR #5704, opened 2019-11-22
Stopped empty journal title elements from appearing in xml.

bobcaprice:


bobcaprice:
[Fix most common public profile validation error](https://trello.com/c/n93u27qO/6430-fix-most-common-public-profile-validation-error)

PR #5707, opened 2019-12-04
Fixed problem with following redirects on different protocol for DOIs

bobcaprice:


bobcaprice:
[KISTI add by DOI not working (2)](https://trello.com/c/nba3jyZW/6404-kisti-add-by-doi-not-working-2)

PR #5710, opened 2019-12-10
Adding more logging for institutional sign in

amontenegro:


amontenegro:
[Institutional sign in error when there is more that one id available in the request headers](https://trello.com/c/5hN6Adt8/6446-institutional-sign-in-error-when-there-is-more-that-one-id-available-in-the-request-headers)

PR #5711, opened 2019-12-11
LoadIssnData script can now be run any time and updates records accor…

bobcaprice:
…ding to last modified date. Incoming data from issn portal cleaned to avoid bad unicode characters'


bobcaprice:
[Peer Review Group ID Journal Name displayed with incorrect characters in UI (3)](https://trello.com/c/VR30kowi/6454-peer-review-group-id-journal-name-displayed-with-incorrect-characters-in-ui-3)

PR #5712, opened 2019-12-11
Add "Authorization" to allowed CORS headers

### LINKED ISSUE #5675: Add "authorization" to allowed CORS headers for oauth/userinfo requests
I'm using the [`oidc-client`](https://www.npmjs.com/package/oidc-client) library to authenticate users in a web browser (Chrome 77, in this case), and after successfully authenticating a user with ORCID it makes a cross-domain request to fetch the user's profile information.

Unfortunately an OPTIONS request to https://sandbox.orcid.org/oauth/userinfo fails: the request contains a `Access-Control-Request-Headers: authorization` header but the response has only `access-control-allow-headers: X-Requested-With,Origin,Content-Type,Accept,x-csrf-token` - `authorization` is missing from the allowed headers.

Would it be possible to add `authorization` to the list of allowed headers?

hubgit:
Requests to `/oauth/userinfo` are authenticated using a Bearer token in the `Authorization` header, so this needs to be added to the list of allowed headers in `Access-Control-Allow-Headers`.

fixes #5675

### DIFF
--- a/orcid-core/src/main/java/org/orcid/core/web/filters/CorsFilterWeb.java
+++ b/orcid-core/src/main/java/org/orcid/core/web/filters/CorsFilterWeb.java
@@ -51,7 +51,7 @@ public class CorsFilterWeb extends OncePerRequestFilter {
                 if (request.getHeader("Access-Control-Request-Method") != null && "OPTIONS".equals(request.getMethod())) {
                     // CORS "pre-flight" request
                     response.addHeader("Access-Control-Allow-Methods", "GET, POST, PUT, DELETE");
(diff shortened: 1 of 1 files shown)


amontenegro:
Hi @TomDemeranville, @wjrsimpson 

What do you think about this? 

Thanks

amontenegro:
[Test userinfo endpoints](https://trello.com/c/SYG25jAi/6479-test-userinfo-endpoints)

amontenegro:
Thanks @hubgit.

@wjrsimpson I just created the card and will follow up with QA team.

Thanks

PR #5714, opened 2019-12-12
If only invalid or empty fl args supplied, default fl param of all va…

bobcaprice:
…lid fields is used


bobcaprice:
[Make API search useful: randow fields returned and odd looking results when incorrect fields used to search](https://trello.com/c/F9xOO0SL/6457-make-api-search-useful-randow-fields-returned-and-odd-looking-results-when-incorrect-fields-used-to-search)

PR #5717, opened 2019-12-16
Creating fundings and affiliations in UI no longer result in orgs wit…

bobcaprice:
…h null regions


bobcaprice:
[Orgs still being created with NULL region](https://trello.com/c/qkBEaog3/6461-orgs-still-being-created-with-null-region)

PR #5725, opened 2019-12-19
If a user have more than one primary email, we should set just one as…

amontenegro:
… the primar


amontenegro:
[Automatically set one primary email address on record with multiple primary email addresses (3)](https://trello.com/c/LaoTHkH8/6463-automatically-set-one-primary-email-address-on-record-with-multiple-primary-email-addresses-3)

PR #5726, opened 2019-12-19
When editing emails, trim them before sending them to the database

amontenegro:


amontenegro:
["Add another email" field allows addresses with trailing whitespace](https://trello.com/c/Jmyn8oXv/6471-add-another-email-field-allows-addresses-with-trailing-whitespace)

PR #5737, opened 2020-01-07
Org regions never set with null in entity

bobcaprice:


bobcaprice:
[Orgs still being created with NULL region](https://trello.com/c/qkBEaog3/6461-orgs-still-being-created-with-null-region)

amontenegro:
Hi @bobcaprice, could you please take a look at what failed for this branch?

Thanks

PR #5739, opened 2020-01-09
Add back work count when pagination is enabled

leomendoza123:


leomendoza123:
[Works pagination: work count doesn't show in header on records with more than 50 works](https://trello.com/c/h0tBNnQN/6490-works-pagination-work-count-doesnt-show-in-header-on-records-with-more-than-50-works)

PR #5744, opened 2020-01-15
Update source id references for deprecated profiles

bobcaprice:


bobcaprice:
[Data issue with same ORCID ids in profile and client details tables, where profile.orcid_type is 'CLIENT'](https://trello.com/c/HXGYMamV/6447-data-issue-with-same-orcid-ids-in-profile-and-client-details-tables-where-profileorcidtype-is-client)

PR #5745, opened 2020-01-15
Removed unused model attribute

bobcaprice:


bobcaprice:
[isDelegatedByAdmin model attribute doesn't appear to be used](https://trello.com/c/8RF7whBJ/6465-isdelegatedbyadmin-model-attribute-doesnt-appear-to-be-used)

PR #5752, opened 2020-01-20
Hide new header on print view

leomendoza123:


leomendoza123:
[Public page print view now contains the navigation titles (2)](https://trello.com/c/WDWaDsW1/6406-public-page-print-view-now-contains-the-navigation-titles-2)

PR #5753, opened 2020-01-20
Fix peer review detail date

leomendoza123:


leomendoza123:
[Peer Review Completion date incorrect when year, month and day are added (0.5)](https://trello.com/c/AeqJe694/6434-peer-review-completion-date-incorrect-when-year-month-and-day-are-added-05)

PR #5756, opened 2020-01-27
Is delegated by admin bug

bobcaprice:


bobcaprice:
[isDelegatedByAdmin always returns true once an admin has switched user and back again](https://trello.com/c/7qRwoGbx/6464-isdelegatedbyadmin-always-returns-true-once-an-admin-has-switched-user-and-back-again)

PR #5760, opened 2020-01-31
adding UTF-8 character encoding to display special characters in admi…

DanielPalafox:
https://trello.com/c/Ku7R3y74/6431-admin-lookup-tool-does-not-display-names-with-non-roman-chars


DanielPalafox:
[Admin lookup tool does not display names with non-Roman chars](https://trello.com/c/Ku7R3y74/6431-admin-lookup-tool-does-not-display-names-with-non-roman-chars)

PR #5773, opened 2020-02-10
Script for updating indexing status of all invalid records

bobcaprice:


bobcaprice:
[Reindex all locked / deactivated etc records](https://trello.com/c/oa7GVQao/6539-reindex-all-locked-deactivated-etc-records)

PR #5780, opened 2020-02-12
Force logout funcitonality now removes JSESSIONID cookie

bobcaprice:


bobcaprice:
[Set rackspace cookie samesite flag](https://trello.com/c/jgBq2utG/6538-set-rackspace-cookie-samesite-flag)

PR #5786, opened 2020-02-17
Added liquibase file with changesets for both postgresql and hsqldb f…

bobcaprice:
…or adding not null constraints


bobcaprice:
[Add NOT NULL constraints to org table to enforce unique constraint](https://trello.com/c/sOn23cDx/6559-add-not-null-constraints-to-org-table-to-enforce-unique-constraint)

PR #5790, opened 2020-02-20
Check user status every 30 seconds

amontenegro:


amontenegro:
[UserStatus check every 30 seconds](https://trello.com/c/YpIRb83h/6570-userstatus-check-every-30-seconds)

PR #5796, opened 2020-02-21
On user status check, reload the page only if it is a 205 error, whic…

amontenegro:
…h means a CSRF error


amontenegro:
[On user status, reload page only on CSRF errors](https://trello.com/c/hWlzEfqE/6573-on-user-status-reload-page-only-on-csrf-errors)

PR #5798, opened 2020-02-24
Uncommented out not null constraints liquibase script

bobcaprice:


bobcaprice:
[Add NOT NULL constraints to org table to enforce unique constraint](https://trello.com/c/sOn23cDx/6559-add-not-null-constraints-to-org-table-to-enforce-unique-constraint)

PR #5807, opened 2020-03-02
Fix a11y on mantainance banner

leomendoza123:


leomendoza123:
[Element has insufficient color contrast](https://trello.com/c/QH2knD5C/6595-element-has-insufficient-color-contrast)

PR #5809, opened 2020-03-03
Updates to ma tests pedro

pedrommcosta:
Behold my first PR!


caoyler:
@pedrommcosta Looks good. I had a couple small suggestions:

Cookies banner- we need to tell people to use incognito window for this. Also, can this be done on /register to skip the going to the home screen step?

Reset password
Missing step to go to Visit https://qa.orcid.org/reset-password at the start
Send both emails, then go to mailinator only once

Step 40, We skip the step to click to verify the email

Step 66 plus. We don't need to change the language. Let's just set 0000-0002-0517-4718 to be in Russian and step 66 is only to confirm that the record displays in Russian

PR #5814, opened 2020-03-06
Multiple emails now returned in expanded search to mirror csv search

bobcaprice:


bobcaprice:
[Expanded search only returns one email address CSV search returns all verified, public email addresses](https://trello.com/c/72QvOmjG/6609-expanded-search-only-returns-one-email-address-csv-search-returns-all-verified-public-email-addresses)

PR #5815, opened 2020-03-06
Fix DB change log name

amontenegro:


amontenegro:
[Create views for the biography tables](https://trello.com/c/oXt9gJNg/6594-create-views-for-the-biography-tables)

PR #5816, opened 2020-03-06
 adding validation to work identifiers entered in URL form

DanielPalafox:
https://trello.com/c/lTIF4jIg/6575-on-work-identifiers-we-parse-into-urls-but-do-not-validate-check-if-the-identifier-is-entered-in-url-form-3


DanielPalafox:
[On work identifiers we parse into URLs but do not validate, check if the identifier is entered in URL form (3)](https://trello.com/c/lTIF4jIg/6575-on-work-identifiers-we-parse-into-urls-but-do-not-validate-check-if-the-identifier-is-entered-in-url-form-3)

PR #5818, opened 2020-03-06
Add unsuported banner with regexp

leomendoza123:


leomendoza123:
[Update old browser banner check](https://trello.com/c/7OGCcVT6/6605-update-old-browser-banner-check)

PR #5819, opened 2020-03-09
replacing source id and display only domain from urls

DanielPalafox:
https://trello.com/c/qjpxHfcA/6613-updates-in-the-generated-views


DanielPalafox:
[Updates in the generated views](https://trello.com/c/qjpxHfcA/6613-updates-in-the-generated-views)

PR #5831, opened 2020-03-18
remove empty strings from JA messages file

lizkrznarich:


lizkrznarich:
[Untranslated JA string appear as blank](https://trello.com/c/xbAQj7X0/6623-untranslated-ja-string-appear-as-blank)

PR #5838, opened 2020-03-26
Affiliation dates fix

Camelia-Orcid:
Fix for validation of the date of the Affiliations section.  

( Trello card: Affiliations allow dates with no month if the day is set to 01)


Camelia-Orcid:
[Affiliations allow dates with no month if the day is set to 01](https://trello.com/c/eimHqqZd/6628-affiliations-allow-dates-with-no-month-if-the-day-is-set-to-01)

PR #5839, opened 2020-03-26
Admin tools: Turn off 2FA as an admin without switching to user's record

Camelia-Orcid:
Turn off the Two factor Authentication as admin from Admin Actions page  without the need to switch to the user first


Camelia-Orcid:
[Admin tools: Turn off 2FA as an admin without switching to user's record (3)](https://trello.com/c/fEOhDRls/6627-admin-tools-turn-off-2fa-as-an-admin-without-switching-to-users-record-3)

PR #5842, opened 2020-03-31
Fix invalid urls

amontenegro:


amontenegro:
[Filter invalid URLs from the V2.0 API and allow displaying them on the V2.0 and V3.0 API when there is one](https://trello.com/c/N1At0wz5/6646-filter-invalid-urls-from-the-v20-api-and-allow-displaying-them-on-the-v20-and-v30-api-when-there-is-one)

PR #5847, opened 2020-04-02
Fix for the case when year not provided only month and day

Camelia-Orcid:
I have added the patch for the bug found in QA ( existent in current production release as well) .


Camelia-Orcid:
[[QA] Affiliations allow dates with no year if the day is set to 01](https://trello.com/c/4fkZgqsA/6650-qa-affiliations-allow-dates-with-no-year-if-the-day-is-set-to-01)

PR #5849, opened 2020-04-02
6612 create views for other tables

Camelia-Orcid:
The other views for the data warehouse.


Camelia-Orcid:
[Create views for other tables](https://trello.com/c/F85QY6Iu/6612-create-views-for-other-tables)

PR #5850, opened 2020-04-02
Prevent disabling of members

amontenegro:


amontenegro:
[Prevent disabling members in the admin funcionality](https://trello.com/c/sSh54KiQ/6654-prevent-disabling-members-in-the-admin-funcionality)

PR #5851, opened 2020-04-06
If there is an invalid DOI added through the UI, show an error indica…

amontenegro:
…ting the user the resource is not resolvable


amontenegro:
[[QA] UI does not show error for DOIs that would create invalid URLs](https://trello.com/c/rxWU5NWS/6656-qa-ui-does-not-show-error-for-dois-that-would-create-invalid-urls)

PR #5855, opened 2020-04-09
Added logging to works processing

bobcaprice:


bobcaprice:
[Add logging to script correcting references to profile records of CLIENT type](https://trello.com/c/2OhH5NgR/6667-add-logging-to-script-correcting-references-to-profile-records-of-client-type)

PR #5857, opened 2020-04-12
dw_profile_keyword added rows expanding for keyword column for delimitators

Camelia-Orcid:
Split values  for keyword  column like 'one,two,three' into multiple rows  for  the common delimiters are , ; .


Camelia-Orcid:
[data warehouse dw_profile_keyword view needs rows expanding](https://trello.com/c/USsRe4SP/6625-data-warehouse-dwprofilekeyword-view-needs-rows-expanding)

PR #5862, opened 2020-04-18
6612 create views for other tables changes

Camelia-Orcid:
Added the  changes for dw_userconnection and dw_oauth2_token_detail

dw_oauth2_token_detail adding token_expiration, revocation_date
dw_userconnection table contains only accounts with orcid not null or
not empty


Camelia-Orcid:
[Create views for other tables](https://trello.com/c/F85QY6Iu/6612-create-views-for-other-tables)

PR #5863, opened 2020-04-20
Added the table for active user count

Camelia-Orcid:
Created the table dw_active_users to store active user count for data warehouse


Camelia-Orcid:
[Start tracking active user count (5)](https://trello.com/c/3QBg1qul/6670-start-tracking-active-user-count-5)

PR #5867, opened 2020-04-23
Arm in open id token

amontenegro:


amontenegro:
[Add AMR to id_token so integrations know if someone used 2fa (3)](https://trello.com/c/9LVa7TqK/5302-add-amr-to-idtoken-so-integrations-know-if-someone-used-2fa-3)

PR #5871, opened 2020-04-30
Fix 500 error when the user sends and invalid email address on the de…

amontenegro:
…activated user form


amontenegro:
[Send reactivation email form throw 500 error if you enter invalid email address.](https://trello.com/c/5UVSND07/6717-send-reactivation-email-form-throw-500-error-if-you-enter-invalid-email-address)

PR #5876, opened 2020-05-08
Add given and family names in expanded search

DanielPalafox:
https://trello.com/c/kjrf6tcV/6629-advance-search-doesnt-return-diacritics


DanielPalafox:
[Advance search doesn't return diacritics](https://trello.com/c/kjrf6tcV/6629-advance-search-doesnt-return-diacritics)

PR #5878, opened 2020-05-11
Peer review org now optional in api v3, removed not null db constrain…

bobcaprice:
…t and updated orcid-model dependency version


bobcaprice:
[Make convening org ID optional in 3.0 for peer review (3)](https://trello.com/c/283onKi1/6531-make-convening-org-id-optional-in-30-for-peer-review-3)

PR #5881, opened 2020-05-18
Removed the contributors_json from the view

Camelia-Orcid:
In the future we will create a separate view/table for contributors ( there is a separate trello card in road map for it)


Camelia-Orcid:
[Datawarehouse:  modify the dw_work  view to not include the contributor_json column](https://trello.com/c/AEp1qluS/6752-datawarehouse-modify-the-dwwork-view-to-not-include-the-contributorjson-column)

PR #5887, opened 2020-05-22
Added write rights for dw_user to dw_active_users table

Camelia-Orcid:


Camelia-Orcid:
[DW: grant dw_user write to dw_active_users table](https://trello.com/c/feZGjrGv/6784-dw-grant-dwuser-write-to-dwactiveusers-table)

PR #5889, opened 2020-05-25
6765 drop the unused profile columns table thru liquibase

Camelia-Orcid:


Camelia-Orcid:
[Drop the unused profile columns table  thru liquibase.](https://trello.com/c/CBM0KYst/6765-drop-the-unused-profile-columns-table-thru-liquibase)

PR #5892, opened 2020-05-27
UI allows peer-review orgs to be null

bobcaprice:


bobcaprice:
[[QA] UI unable to show peer reviews if record has a peer review without convening org ID](https://trello.com/c/DJjlhCRJ/6764-qa-ui-unable-to-show-peer-reviews-if-record-has-a-peer-review-without-convening-org-id)

PR #5898, opened 2020-05-28
Fix long texts

DanielPalafox:
https://trello.com/c/8vRb5Os0/6801-peer-review-with-long-source-work-id-breaks-ui


DanielPalafox:
[Peer review with long source-work-id breaks UI](https://trello.com/c/8vRb5Os0/6801-peer-review-with-long-source-work-id-breaks-ui)

PR #5901, opened 2020-06-01
Fixed ringgold importer bug caused by assumption that new orgs don't …

bobcaprice:
…already exist


bobcaprice:
[Fix Ringgold importer bug with deprecated orgs](https://trello.com/c/P1tE8eie/6806-fix-ringgold-importer-bug-with-deprecated-orgs)

PR #5930, opened 2020-06-24
Fix date created issue on V2 and V3

amontenegro:


amontenegro:
[[QA] Updating peer-review group id removes created_date from db](https://trello.com/c/dPk88Ho6/6843-qa-updating-peer-review-group-id-removes-createddate-from-db)

PR #5939, opened 2020-07-06
Modify DOI regex to match Ariti DOI's

amontenegro:


amontenegro:
[Modify DOI validation regex to allow Ariti DOI's](https://trello.com/c/G2D8QaVr/6862-modify-doi-validation-regex-to-allow-ariti-dois)

PR #5943, opened 2020-07-08
Set a timeout of 30 secs in the queries to look for records to index

amontenegro:


amontenegro:
[Scheduler stop feeding the message listener (5)](https://trello.com/c/0yzDJXgk/6256-scheduler-stop-feeding-the-message-listener-5)

PR #5950, opened 2020-07-15
Changes to send welcome emails for users

Camelia-Orcid:


Camelia-Orcid:
[Resend Welcome Emails for the profile created between dates](https://trello.com/c/E8MvR4Qi/6877-resend-welcome-emails-for-the-profile-created-between-dates)

PR #5955, opened 2020-07-16
Fix for Private name not to be displayed in the list of trusted individuals.

Camelia-Orcid:


Camelia-Orcid:
[[QA] Private names show in list of trusted individuals](https://trello.com/c/c8SV0H5Q/6859-qa-private-names-show-in-list-of-trusted-individuals)

PR #5962, opened 2020-07-21
fix return json instead of null

DanielPalafox:
https://trello.com/c/CZZJXgJB/4777-return-json-instead-of-null-from-oauth-custom-authorize-getrequestinfoformjson-1


DanielPalafox:
[Return json instead of null from /oauth/custom/authorize/get_request_info_form.json (1)](https://trello.com/c/CZZJXgJB/4777-return-json-instead-of-null-from-oauth-custom-authorize-getrequestinfoformjson-1)

PR #5964, opened 2020-07-23
On bibtex import: Include issn ext ids and ignore ids with emtpy value

amontenegro:


amontenegro:
[BibTeX unable to import ISSNs and instead imports empty DOI](https://trello.com/c/EGBtzCj6/6888-bibtex-unable-to-import-issns-and-instead-imports-empty-doi)

PR #5965, opened 2020-07-24
Sending message to all users emails when disabling 2FA

Camelia-Orcid:


Camelia-Orcid:
[Send an email to all email addresses on a record when 2FA is turned off (5)](https://trello.com/c/zM77XEPK/6638-send-an-email-to-all-email-addresses-on-a-record-when-2fa-is-turned-off-5)

PR #5966, opened 2020-07-27
CODE OF CONDUCT

gaushikmr:


wjrsimpson:
Thanks for the pull request!

Having a code of code of conduct is great idea. However, we have recently created our own company dignity at work statement. I will review that and see if we can use it for our open source repos.

So, I am closing this pull request but thanks for the push in a good direction.

PR #5974, opened 2020-08-04
Update browserlist regexp

leomendoza123:
update to make orcid source match the same regexp used on orcid angular. This is just a periodical update.


Camelia-Orcid:
[Puppet for DW Production](https://trello.com/c/OFJDVqnR/6674-puppet-for-dw-production)

PR #5975, opened 2020-08-04
Corrected response code for OrcidInvalidScopeException being thrown

bobcaprice:


bobcaprice:
[Token delegation: 400 error instead of 401 when user revokes access](https://trello.com/c/3OZ3ruI0/6883-token-delegation-400-error-instead-of-401-when-user-revokes-access)

PR #5984, opened 2020-08-17
fix return requestInfoForm if user is already logged

DanielPalafox:
https://trello.com/c/PGW2b438/6887-speedup-oath-custom-initjson-already-authorize-support


DanielPalafox:
[[speedup]/oath/custom/init.json already authorize support](https://trello.com/c/PGW2b438/6887-speedup-oath-custom-initjson-already-authorize-support)

PR #5988, opened 2020-08-18
fix open id prompt none and add missing validations

DanielPalafox:
https://trello.com/c/vn59ESpm/6884-speedup-open-id-prompnone-does-not-return-the-interactionrequired-error-code


DanielPalafox:
[[speedup] Open id promp=none does not return the 'interaction_required` error code](https://trello.com/c/vn59ESpm/6884-speedup-open-id-prompnone-does-not-return-the-interactionrequired-error-code)

PR #5990, opened 2020-08-21
fix update attributes if oauth request changed

DanielPalafox:
https://trello.com/c/snLkIRIs/6921-speedupoauth-atributes-do-not-update-when-the-oauth-link-change


DanielPalafox:
[[speedup]Oauth atributes do not update when the Oauth link change](https://trello.com/c/snLkIRIs/6921-speedupoauth-atributes-do-not-update-when-the-oauth-link-change)

PR #5994, opened 2020-08-26
Granted select privilege to dw_user for work table.

Camelia-Orcid:


Camelia-Orcid:
[Create views that extract work contributors and work external identifiers and put one per row (5)](https://trello.com/c/b1mGQz50/6850-create-views-that-extract-work-contributors-and-work-external-identifiers-and-put-one-per-row-5)

PR #6006, opened 2020-09-09
fix missing emailId in shibboleth

DanielPalafox:
https://trello.com/c/W8XJDqsk/6844-register-shibboleth-support


DanielPalafox:
[Register: shibboleth support](https://trello.com/c/W8XJDqsk/6844-register-shibboleth-support)

PR #6011, opened 2020-09-15
ISNI org external ids now normalised for fundref, ringgold, and grid …

bobcaprice:
…imports


bobcaprice:
[Org ID external ID ISNIs are in wrong formats (5)](https://trello.com/c/gTF0XNm4/6919-org-id-external-id-isnis-are-in-wrong-formats-5)

PR #6012, opened 2020-09-16
Member obo id token expiry

bobcaprice:


bobcaprice:
[Reduce id token lifespan to 24hs due to security considerations (3)](https://trello.com/c/byfjid09/6866-reduce-id-token-lifespan-to-24hs-due-to-security-considerations-3)

PR #6015, opened 2020-09-18
Select accounts based on the RecordTypeId

amontenegro:


amontenegro:
[Scoping of impact of new SF Opp and ORG record types on Self-service and Members list - and do the work! (5)](https://trello.com/c/Ee7yVEm9/6911-scoping-of-impact-of-new-sf-opp-and-org-record-types-on-self-service-and-members-list-and-do-the-work-5)

PR #6028, opened 2020-10-06
fix invalid request oauth

DanielPalafox:
fix NPE in oauth request and catch errors when a parameter is empty


leomendoza123:
[Invalid requestInfoForm](https://trello.com/c/kM6aYdoj/7050-invalid-requestinfoform)

PR #6035, opened 2020-10-07
fix invalid redirect URL

DanielPalafox:
Fix if the redirect URL has invalid characters


amontenegro (changes requested):
@DanielPalafox do we really need this?

amontenegro:
This breaks current integrations that needs to store encoded white space in the redirect uri (%20), @DanielPalafox is going to investigate alternatives for this.

PR #6036, opened 2020-10-07
Code compiling with  jdk11

Camelia-Orcid:
Upgraded pom's to compile with java 11, fixed failing tests.


Camelia-Orcid:
all the work to migrate jdk11 is in other branch

PR #6038, opened 2020-10-12
fix warning message and display only in redirect URLs

DanielPalafox:
https://trello.com/c/lLQX6aIW/6962-show-warning-message-to-prevent-http-redirect-uris-being-added-to-clients


DanielPalafox:
[Show warning message to prevent http redirect URIs being added to clients](https://trello.com/c/lLQX6aIW/6962-show-warning-message-to-prevent-http-redirect-uris-being-added-to-clients)

PR #6042, opened 2020-10-13
Fixed publisher label

Camelia-Orcid:


Camelia-Orcid:
[[QA] replace journal title field with publisher field when adding a book DOI](https://trello.com/c/G8BpBl7d/6910-qa-replace-journal-title-field-with-publisher-field-when-adding-a-book-doi)

PR #6043, opened 2020-10-14
display warning error only in production

DanielPalafox:
https://trello.com/c/pPanvRsM/7069-show-warning-message-only-in-production-to-prevent-http-redirect-uris-being-added-to-clients


DanielPalafox:
[Show warning message only in production to prevent http redirect URIs being added to clients](https://trello.com/c/pPanvRsM/7069-show-warning-message-only-in-production-to-prevent-http-redirect-uris-being-added-to-clients)

PR #6044, opened 2020-10-14
Remove user status check from logs

amontenegro:


amontenegro:
[Remove userStatus.json from the access logs](https://trello.com/c/TZ6ZMky8/7062-remove-userstatusjson-from-the-access-logs)

PR #6046, opened 2020-10-20
fix missing getBaseUri

DanielPalafox:
https://trello.com/c/BTh749ad/7088-cant-add-redirect-uris


amontenegro:
[Can't add redirect URIs](https://trello.com/c/BTh749ad/7088-cant-add-redirect-uris)

PR #6048, opened 2020-10-23
Ringgold import fix

bobcaprice:


bobcaprice:
[Ringgold import failure](https://trello.com/c/LRM7wnwG/7094-ringgold-import-failure)

PR #6056, opened 2020-10-28
Added HAL identifier_type

bobcaprice:


bobcaprice:
[Add HAL as a supported work identifier](https://trello.com/c/2tHDdjVn/7102-add-hal-as-a-supported-work-identifier)

PR #6057, opened 2020-10-28
fix oauth session should not continue if user navigates to signin or …

DanielPalafox:
https://trello.com/c/N6WgO5E8/7093-qa-oauth-session-should-not-persist-after-the-user-abandons-the-authorization-screen


DanielPalafox:
[[QA] OAuth session should not persist after the user abandons the authorization screen](https://trello.com/c/N6WgO5E8/7093-qa-oauth-session-should-not-persist-after-the-user-abandons-the-authorization-screen)

PR #6069, opened 2020-11-04
Filter the results before return

amontenegro:


amontenegro:
[Adjustments to Members List on the website to ensure only Consortia Leads that are also Consortium Members show on the All Members tab, while keeping everything else the same.](https://trello.com/c/0auB5F0i/7101-adjustments-to-members-list-on-the-website-to-ensure-only-consortia-leads-that-are-also-consortium-members-show-on-the-all-membe)

PR #6072, opened 2020-11-09
Fix for finding version to use for copy, now checks put code

bobcaprice:


bobcaprice:
[Make copy button loads wrong work](https://trello.com/c/5fHL0768/7128-make-copy-button-loads-wrong-work)

PR #6075, opened 2020-11-10
fix oauth account linking

DanielPalafox:
https://trello.com/c/DIeeZy72/7129-oauth-account-linking-process-is-broken


DanielPalafox:
[Oauth + Account Linking process is broken](https://trello.com/c/DIeeZy72/7129-oauth-account-linking-process-is-broken)

amontenegro (approved):

PR #6077, opened 2020-11-10
Revert "fix oauth session should not continue if user navigates to si…

leomendoza123:
…gnin or register"

This reverts commit 190dff26006ecde473914aa35649ce75fe7c24e6.


leomendoza123:
[Oauth + Account Linking process is broken](https://trello.com/c/DIeeZy72/7129-oauth-account-linking-process-is-broken)

PR #6082, opened 2020-11-13
fix 2fa deprecate

DanielPalafox:


DanielPalafox:
[[QA] 2fa code is not requested when deprecating account with 2fa enabled](https://trello.com/c/ovbLWiSd/3157-qa-2fa-code-is-not-requested-when-deprecating-account-with-2fa-enabled)

PR #6099, opened 2020-11-24
fix error message in manage member

DanielPalafox:
https://trello.com/c/zaDddmRl/7131-manage-member-should-display-a-message-if-member-is-not-found-is-throwing-error-in-the-console-instead


DanielPalafox:
[Manage member should display a message if member is not found is throwing error in the console instead](https://trello.com/c/zaDddmRl/7131-manage-member-should-display-a-message-if-member-is-not-found-is-throwing-error-in-the-console-instead)

PR #6109, opened 2020-11-26
Done

amontenegro:


amontenegro:
[[QA] Change ORCID Swagger test redirect URI in /developer-tools to use v3.0 instead of v2.0](https://trello.com/c/nXDwNczC/7166-qa-change-orcid-swagger-test-redirect-uri-in-developer-tools-to-use-v30-instead-of-v20)

PR #6113, opened 2020-11-30
From 6000 to 4500

amontenegro:


amontenegro:
[Reduce the max number of cached elements(1)](https://trello.com/c/F56eA4eJ/7180-reduce-the-max-number-of-cached-elements1)

PR #6116, opened 2020-12-02
Check user is still a SF contact before allowing him to execute actio…

amontenegro:
…ns on SF


amontenegro:
[Disable access to member self service once the contact is no longer active](https://trello.com/c/gaXnhL8a/7173-disable-access-to-member-self-service-once-the-contact-is-no-longer-active)

PR #6118, opened 2020-12-02
7181 qa separate each notification email into its own property file

leomendoza123:


leomendoza123:
[[QA] Separate each notification email into its own property file](https://trello.com/c/okZLegPb/7181-qa-separate-each-notification-email-into-its-own-property-file)

PR #6122, opened 2020-12-09
Check membership contact role as well

amontenegro:


amontenegro:
[Disable access to member self service once the contact is no longer active](https://trello.com/c/gaXnhL8a/7173-disable-access-to-member-self-service-once-the-contact-is-no-longer-active)

PR #5761, opened 2020-02-03
Reduce analytics sampling from 100 to 70 per cent

leomendoza123:

PR #6127, opened 2020-12-15
Revert "Reduce analytics sampling from 100 to 70 per cent"

leomendoza123:
Reverts ORCID/ORCID-Source#5761


leomendoza123:
[Modify SIGN-IN event to GA](https://trello.com/c/YsTvisp2/7072-modify-sign-in-event-to-ga)

PR #6128, opened 2020-12-16
Transifex update

leomendoza123:


amontenegro:
Hi @leomendoza123 could you please remember me why we need this?

Thanks

PR #6129, opened 2020-12-21
Last modified and indexing status

bobcaprice:


bobcaprice:
[Last updated should only change if something on the record changes.](https://trello.com/c/KVWcPvBH/6961-last-updated-should-only-change-if-something-on-the-record-changes)

PR #6134, opened 2020-12-31
Statistics values will be set by a @PostConstruct and then refreshed …

amontenegro:
…every hour by a cron job


amontenegro:
[Improve the way stats are being cached](https://trello.com/c/chjZc3St/7206-improve-the-way-stats-are-being-cached)

PR #6136, opened 2021-01-04
Register social/shibboleth signins

leomendoza123:


leomendoza123:
[Modify SIGN-IN event to GA](https://trello.com/c/YsTvisp2/7072-modify-sign-in-event-to-ga)

PR #6138, opened 2021-01-05
The member tools menu item should display just if the member is an ac…

amontenegro:
…tive contact in SF


amontenegro:
[[QA] menu option to access member tools stays visible after org contact is deleted from SF](https://trello.com/c/UwZa773r/7182-qa-menu-option-to-access-member-tools-stays-visible-after-org-contact-is-deleted-from-sf)

PR #6143, opened 2021-01-12
adding url validation regex

DanielPalafox:
https://trello.com/c/SjnodmXQ/7213-member-cannot-add-redirect-uris-although-they-are-valid


DanielPalafox:
Adding regex validation to accept underscores in subdomains since commons-validator from apache doesn't support it and it should be accepted according to https://tools.ietf.org/html/rfc2181#section-11

DanielPalafox:
[Member cannot add Redirect URIS although they are valid](https://trello.com/c/SjnodmXQ/7213-member-cannot-add-redirect-uris-although-they-are-valid)

PR #6144, opened 2021-01-18
fix userid is not been collected

DanielPalafox:
https://trello.com/c/qj8NIADs/7211-collect-orcid-id-via-hotjar-identity-api


DanielPalafox:
[Collect ORCID iD via Hotjar identity API](https://trello.com/c/qj8NIADs/7211-collect-orcid-id-via-hotjar-identity-api)

PR #6153, opened 2021-01-22
Convert public client to member

bobcaprice:


bobcaprice:
[Ability to update public API credentials to Member API credentials](https://trello.com/c/PqW8QgDo/7208-ability-to-update-public-api-credentials-to-member-api-credentials)

PR #6155, opened 2021-02-02
Send the amend email when a member is updating a funding

amontenegro:


amontenegro:
[[QA] inbox notification not sent when a funding item is updated](https://trello.com/c/wZjwakLu/7236-qa-inbox-notification-not-sent-when-a-funding-item-is-updated)

PR #6164, opened 2021-02-10
Contributor type works

bobcaprice:


bobcaprice:
Failing due to model not having been pushed to mvn central. 

Currently working on fundings role conversion with branches this branch so will later raise a PR for both

PR #6166, opened 2021-02-15
Remove name max-width

leomendoza123:


leomendoza123:
[Remove maximum width from full name display in current UI](https://trello.com/c/PCGGFf2H/7248-remove-maximum-width-from-full-name-display-in-current-ui)

PR #6168, opened 2021-02-16
Add batch emails edit endpoint

leomendoza123:


leomendoza123:
[[speedup] Bio: Emails (5)](https://trello.com/c/4EOdjRy6/7214-speedup-bio-emails-5)

PR #6170, opened 2021-02-17
Dont send the UserStatus check if we are in an error page

amontenegro:


amontenegro:
[Stop sending the apiStatus check from the error page (2)](https://trello.com/c/muUCbLIf/7249-stop-sending-the-apistatus-check-from-the-error-page-2)

PR #6173, opened 2021-02-23
fix show client name in emails

DanielPalafox:


DanielPalafox:
[[QA] [new inbox] client ID shown instead of client name](https://trello.com/c/KOoYk0SB/7253-qa-new-inbox-client-id-shown-instead-of-client-name)

PR #6175, opened 2021-03-01
Fix french translation

leomendoza123:


leomendoza123:
[Fix French translation for 'Journal Article'](https://trello.com/c/3d1pQZK1/7241-fix-french-translation-for-journal-article)

PR #6176, opened 2021-03-02
Github actions that will fire on tagging the develpment or the main b…

amontenegro:
…ranch

### DIFF
--- a/.github/workflows/release_tag_on_development_branch.yml
+++ b/.github/workflows/release_tag_on_development_branch.yml
@@ -1,0 +1,16 @@
+name: release_tag_on_development_branch
+
+on:
+  push:
+    tags:
+      - 'release-*'
+
+jobs:
+  check-bats-version:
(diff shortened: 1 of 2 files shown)


amontenegro:
@gilesw Please see the two actions attached to this PR:

- **release_tag_on_development_branch.yml** will be fired when a new tag with the prefix 'release-' is pushed in the _development_branch_.
- **release_tag_on_main_branch.yml** will be fired when a new tag with the prefix 'release-' is pushed in the _main_ branch_.

This should be the starting point of our new deployment process, so, lets talk more about it in our next call.

Thanks

gilesw:
Heya bud it looks good. However I'm nervous about 2 things:-

1. testing things out on the main repo
2. using github actions with public repos without understanding the security concerns


I've started work on a page.. https://github.com/ORCID/orcid-devops/wiki/CI-Github-actions

PR #6185, opened 2021-03-19
Contributor roles with null values no longer persisted for works or f…

bobcaprice:
…undings


bobcaprice:
[[QA] enum validation for contributor roles is broken in API v3.0](https://trello.com/c/Jz3IqSsP/7277-qa-enum-validation-for-contributor-roles-is-broken-in-api-v30)

PR #6186, opened 2021-03-19
Corrected property names

bobcaprice:


bobcaprice:
[Correct ringgold org import property names](https://trello.com/c/m7isMtSN/7278-correct-ringgold-org-import-property-names)

PR #6187, opened 2021-03-23
FundingGroup will now include the prefered funding so it is easier to…

amontenegro:
… navigate in the UI


amontenegro:
[[speedup] Activities: Funding (5)](https://trello.com/c/BspBYUaL/7122-speedup-activities-funding-5)

PR #6198, opened 2021-04-13
Fix spanish

leomendoza123:


leomendoza123:
[Welcome email in spanish say "ORDIC" instead of "ORCID"](https://trello.com/c/1pDApP91/7320-welcome-email-in-spanish-say-ordic-instead-of-orcid)

PR #6204, opened 2021-04-21
Peer review clear subject name on put

amontenegro:


amontenegro:
[Peer Review item not being updated properly using PUT call.](https://trello.com/c/C7DMlSa6/7324-peer-review-item-not-being-updated-properly-using-put-call)

PR #6205, opened 2021-04-22
fix missing `a` in footer

DanielPalafox:


DanielPalafox:
[Missing 'a' from notification footer](https://trello.com/c/yn6gtv2g/7325-missing-a-from-notification-footer)

PR #6208, opened 2021-04-23
Check if the value is null

amontenegro:


amontenegro:
[Update casrai credit roles to use URIs in the metadata](https://trello.com/c/7tIPTMr8/7330-update-casrai-credit-roles-to-use-uris-in-the-metadata)

PR #6213, opened 2021-04-29
Fix reset-password for unclaim emails 

leomendoza123:


leomendoza123:
[Unable to send claim emails](https://trello.com/c/nnSHFcO7/7335-unable-to-send-claim-emails)

PR #6217, opened 2021-05-05
Stop indexing records with no token

amontenegro:


amontenegro:
[Reduce indexing frequency to 1 per hour (2)](https://trello.com/c/zEVW157W/7327-reduce-indexing-frequency-to-1-per-hour-2)

amontenegro:
[Exclude records from our own search which we exclude from Google indexing [private boards only] (5)](https://trello.com/c/4gpjf0O2/9-exclude-records-from-our-own-search-which-we-exclude-from-google-indexing-private-boards-only-5)

PR #6218, opened 2021-05-05
Fix claim email on reset-password

leomendoza123:


leomendoza123:
[Unable to send claim emails from reset password or admin page](https://trello.com/c/nnSHFcO7/7335-unable-to-send-claim-emails-from-reset-password-or-admin-page)

PR #6230, opened 2021-05-17
Fix bug that cause an NPE when trying to process FUNDREF disambiguate…

amontenegro:
…d orgs on research resources


amontenegro:
[Records with research resources where the proposal host org comes from FUNDREF fails to be indexed](https://trello.com/c/nlxhLjgV/7351-records-with-research-resources-where-the-proposal-host-org-comes-from-fundref-fails-to-be-indexed)

PR #6231, opened 2021-05-18
Fix grid importer, make the ringgold importer really noisy so we can …

amontenegro:
…spot the problem


amontenegro:
[Fix GRID importer](https://trello.com/c/VWVPJKUj/7353-fix-grid-importer)

PR #6241, opened 2021-06-01
fix return emypty objects instead of throwing 500

DanielPalafox:


DanielPalafox:
[Work info details endpoint is returning a server error instead of an empty JSON](https://trello.com/c/irRFgU2G/7359-work-info-details-endpoint-is-returning-a-server-error-instead-of-an-empty-json)

PR #6243, opened 2021-06-02
fix research resources updates doesnt send notifications

DanielPalafox:


DanielPalafox:
[[QA] Inbox notification not sent when API client updates research resource](https://trello.com/c/uRFBPJWt/7295-qa-inbox-notification-not-sent-when-api-client-updates-research-resource)

PR #6250, opened 2021-06-09
Swagger move static content to each api project

amontenegro:


amontenegro:
[Move swagger assets to each app so they dont have to use assets from orcid-web](https://trello.com/c/uJxWIEZT/7364-move-swagger-assets-to-each-app-so-they-dont-have-to-use-assets-from-orcid-web)

PR #6252, opened 2021-06-14
sort researches by affiliations

DanielPalafox:


DanielPalafox:
[Boost records with affiliations in search results (which pushes spam down) [Private boards only] (3)](https://trello.com/c/VPSJWUDc/8-boost-records-with-affiliations-in-search-results-which-pushes-spam-down-private-boards-only-3)

DanielPalafox:
[Boost records with affiliations in the "Could this be you?" modal](https://trello.com/c/XrHvdPDG/14-boost-records-with-affiliations-in-the-could-this-be-you-modal)

PR #6258, opened 2021-06-17
Adde Liquibase extension, which replaces the StandardLockService, by using database locks

Camelia-Orcid:


Camelia-Orcid:
[Investigate and fix liquibase lock when multiple apps deployed at the same time on QA ( jdk11)](https://trello.com/c/HcHfOVD5/7377-investigate-and-fix-liquibase-lock-when-multiple-apps-deployed-at-the-same-time-on-qa-jdk11)

PR #6272, opened 2021-07-08
Changed GET to POST for delegators in account

Camelia-Orcid:


Camelia-Orcid:
["Sign in as" function at https://orcid.org/account is broken.](https://trello.com/c/8Dwpzx7O/7406-sign-in-as-function-at-https-orcidorg-account-is-broken)

PR #6274, opened 2021-07-09
fix source name in email

DanielPalafox:


DanielPalafox:
[[QA][public page] Display the name as source of email addresses when name is public](https://trello.com/c/ffPkK4bS/7409-qapublic-page-display-the-name-as-source-of-email-addresses-when-name-is-public)

PR #6275, opened 2021-07-12
Indexing all records regardles if they have a third party or not

amontenegro:


amontenegro:
[Index new records regardless if they have an access token](https://trello.com/c/pPF6NCrx/17-index-new-records-regardless-if-they-have-an-access-token)

DanielPalafox (approved):

PR #6283, opened 2021-07-22
Bibtext export endpoint

amontenegro:


amontenegro:
[Make export BibTex behave the same way as other bulk actions](https://trello.com/c/PD79me9v/7339-make-export-bibtex-behave-the-same-way-as-other-bulk-actions)

PR #6288, opened 2021-07-29
Unified the key names for ringold

Camelia-Orcid:
It goes hand in hand with the orcid-ansilble_inventory-private key changes commit


Camelia-Orcid:
[Unified key names for Ringgold properties](https://trello.com/c/PfjNNNec/7458-unified-key-names-for-ringgold-properties)

PR #6298, opened 2021-08-10
fix research resource pagination

DanielPalafox:


DanielPalafox:
[[QA][public page] Item counter not accurate when there are more than 50 research resources](https://trello.com/c/P6bHgrmT/7467-qapublic-page-item-counter-not-accurate-when-there-are-more-than-50-research-resources)

PR #6324, opened 2021-08-30
Webhooks should be retried 25 times, with an exponential delay of 10 …

amontenegro:
…mintes by default. Max retries and delay must be configurable


amontenegro:
[OBO source is missing from the new public page.](https://trello.com/c/R0Z2JvR7/7525-obo-source-is-missing-from-the-new-public-page)

amontenegro:
[Only retry the sending of webhooks 25 times max](https://trello.com/c/hBIkvmKt/7515-only-retry-the-sending-of-webhooks-25-times-max)

DanielPalafox (approved):

PR #6340, opened 2021-09-29
Add ISMN external identifier

auumgn:


auumgn:
[Add ISMN as a new work identifier](https://trello.com/c/j5ELn8BM/7611-add-ismn-as-a-new-work-identifier)

PR #6362, opened 2021-10-28
Fix social signin with 2FA enabled

auumgn:


orcid-web/src/main/java/org/orcid/frontend/web/controllers/LoginController.java
```diff
@@ -336,7 +336,7 @@ private ModelAndView processSocialLogin(HttpServletRequest request, HttpServletR
         userCookieGenerator.addCookie(userConnectionId, response);
         
         if ("social_2FA".equals(view.getViewName())) {
-            return new ModelAndView("redirect:"+ orcidUrlManager.getBaseUrl() +"/2fa-signin?social=true");
+            new ModelAndView("redirect:"+ orcidUrlManager.getBaseUrl() +"/2fa-signin?social=true");
```
amontenegro, line 339: Hey @auumgn, I think you should make the `view` equals to the new ModelAndView right? So the code should be like this:

```
view = new ModelAndView("redirect:"+ orcidUrlManager.getBaseUrl() +"/2fa-signin?social=true");
```

amontenegro (changes requested):

amontenegro (approved):

PR #6368, opened 2021-11-15
Added the view corresponding to work external IDs object

Camelia-Orcid:
@amontenegro  is to be merged after the DB migration to fix the external ids sync in panoply


Camelia-Orcid:
[Add the view for work external identifiers](https://trello.com/c/F6GvQmxF/7811-add-the-view-for-work-external-identifiers)

PR #6369, opened 2021-11-18
Remove bit dependency from oldapp

auumgn:


auumgn:
[Remove bitdev dependency from the old registration page (2)](https://trello.com/c/Bm4ZRkVp/7573-remove-bitdev-dependency-from-the-old-registration-page-2)

PR #6375, opened 2021-11-22
orcid-persistence/src/main/resources/db/updates/identifier-types/iden…

amontenegro:
…tifier-type-48-emdb.xml


amontenegro:
[Add EMDB as a new identifier type (2)](https://trello.com/c/UVfgVjm8/7689-add-emdb-as-a-new-identifier-type-2)

PR #5999, opened 2020-09-01
add redirect to new 2fa sign in

DanielPalafox:
https://trello.com/c/L1WNUVfT/6758-add-new-page-link-shibboleth-signin-for-new-institutional-linking-orcid-angular

PR #6381, opened 2021-11-30
Reinstate angular 2FA screen

auumgn:


orcid-web/src/main/java/org/orcid/frontend/web/controllers/ShibbolethController.java
```diff
@@ -159,7 +159,7 @@ public ModelAndView signinHandler(HttpServletRequest request, HttpServletRespons
             
             ProfileEntity profile = profileEntityCacheManager.retrieve(userConnectionEntity.getOrcid());
             if (profile.getUsing2FA()) {
-                return new ModelAndView("redirect:" + orcidUrlManager.getBaseUrl() + "/2fa-signin");
+                return new ModelAndView("institutional_2FA");
```
DanielPalafox, line 162: @auumgn this code will redirect to the old 2FA page https://github.com/ORCID/ORCID-Source/pull/5999/files. I don't think that's the expected behavior. Could you give it another look, please

DanielPalafox (changes requested):

auumgn:
@amontenegro @DanielPalafox reverted previously done changes so it redirects to the angular 2FA screen again.

Related pr: https://github.com/ORCID/orcid-angular/pull/1447

PR #6383, opened 2021-12-01
Fix dup check

amontenegro:


amontenegro:
[Duplicates check for updating fundings is disabled for the API and enabled for the UI](https://trello.com/c/d7DnB00q/7692-duplicates-check-for-updating-fundings-is-disabled-for-the-api-and-enabled-for-the-ui)

DanielPalafox (approved):

PR #6387, opened 2021-12-09
Added the ignore unknown props

Camelia-Orcid:


Camelia-Orcid:
[GRID Uploader - Fails  because of unrecognised  property "path"](https://trello.com/c/pZAUrpeD/7843-grid-uploader-fails-because-of-unrecognised-property-path)

PR #6396, opened 2021-12-15
fix: Refactor plaint text notification email

DanielPalafox:


DanielPalafox:
[Some Crossref notifications appear incorrectly in Zendesk](https://trello.com/c/oKexuzKx/3231-some-crossref-notifications-appear-incorrectly-in-zendesk)

PR #6402, opened 2021-12-21
Google signin - catch missing family name

auumgn:


auumgn:
[Sign in with google is broken if the user doesn't have a last name](https://trello.com/c/lOfIQ6c0/7858-sign-in-with-google-is-broken-if-the-user-doesnt-have-a-last-name)

PR #6404, opened 2021-12-23
Wrong if condition, changed to an OR

amontenegro:


auumgn (approved):
Looks good

amontenegro:
[Provide the ability for member integrators to delete items which they have added to records even when permission has been revoked by the user](https://trello.com/c/zRaHIaFr/7354-provide-the-ability-for-member-integrators-to-delete-items-which-they-have-added-to-records-even-when-permission-has-been-revoke)

PR #6407, opened 2021-12-28
fix: Add missing property in grouping error message

DanielPalafox:


DanielPalafox:
[[QA] Improve error message shown when user tries to combine works that have no identifiers](https://trello.com/c/rSJ3frzC/7872-qa-improve-error-message-shown-when-user-tries-to-combine-works-that-have-no-identifiers)

DanielPalafox:
[DOI normalizer doesn't work with 10.24192/2386-7027(2015)(v3)(09)](https://trello.com/c/VT9VTx3k/7861-doi-normalizer-doesnt-work-with-1024192-2386-70272015v309)

PR #6408, opened 2021-12-29
Catch npe on non existing tokens

amontenegro:


amontenegro:
[[QA] 500 Internal Server Error response when using invalid access token in delete request](https://trello.com/c/76APELVy/7876-qa-500-internal-server-error-response-when-using-invalid-access-token-in-delete-request)

DanielPalafox (approved):

PR #6413, opened 2022-01-05
fix: Keep url in external identifier if its provided

DanielPalafox:


DanielPalafox:
[[QA] Identifier URLs are auto-generated when identifier is not resolvable](https://trello.com/c/1ULRdvOm/7873-qa-identifier-urls-are-auto-generated-when-identifier-is-not-resolvable)

PR #6426, opened 2022-01-25
fix: Add only resolved auto generated external identifiers url

DanielPalafox:


DanielPalafox:
[[QA] Identifier URLs are auto-generated for identifiers the system cannot resolve](https://trello.com/c/bsvtzv1M/7882-qa-identifier-urls-are-auto-generated-for-identifiers-the-system-cannot-resolve)

orcid-core/src/main/java/org/orcid/core/adapter/jsonidentifier/converter/JSONWorkExternalIdentifiersConverterV3.java
```diff
@@ -86,8 +90,17 @@ public ExternalIDs convertFrom(String source, Type<ExternalIDs> destinationType)
...
                     }                    
+                } else {
+                    try {
+                        PIDResolutionResult pdr = resolverService.resolve(id.getType(), workExternalIdentifier.getWorkExternalIdentifierId().content); 
+                        if (pdr.isResolved()) {
+                            id.setNormalizedUrl(new TransientNonEmptyString(pdr.getGeneratedUrl()));
+                        }                       
+                    } catch (IllegalArgumentException iae) {
+
+                    }
```
amontenegro, line 101: @DanielPalafox looks like this will change the API behavior, it will add the normalized url to the response which is a change in which the API works; could you please check that in the member API and let me know the result?

PR #6430, opened 2022-01-27
fix: Update default end date to the latest possible

DanielPalafox:


DanielPalafox:
[[QA][my-orcid] When start and end year are the same, end date cannot have month/day](https://trello.com/c/J9QdeJcv/7851-qamy-orcid-when-start-and-end-year-are-the-same-end-date-cannot-have-month-day)

PR #6431, opened 2022-01-28
Filter invalid chars in group id record names

auumgn:


auumgn:
[Remove strange unicode characters from journal names](https://trello.com/c/Toknvbbg/7860-remove-strange-unicode-characters-from-journal-names)

PR #6434, opened 2022-01-31
7889 qa contributor count in summary view not right

DanielPalafox:


DanielPalafox:
[[QA] Contributor count in summary view not right](https://trello.com/c/G2bySwqN/7889-qa-contributor-count-in-summary-view-not-right)

PR #6445, opened 2022-02-10
Add maximum length to passwords

auumgn:


auumgn:
[Set a maximum length for passwords (3)](https://trello.com/c/oT9yFDcx/22-set-a-maximum-length-for-passwords-3)

PR #6452, opened 2022-02-22
Added the ROR org grouping

Camelia-Orcid:


Camelia-Orcid:
[Organization typeahead to only show each organization once using ROR where available](https://trello.com/c/7Fg5THGe/7875-organization-typeahead-to-only-show-each-organization-once-using-ror-where-available)

Camelia-Orcid:
[Cleanup non ROR orgs from SOLR](https://trello.com/c/5hDlczq2/7913-cleanup-non-ror-orgs-from-solr)

orcid-persistence/src/main/java/org/orcid/persistence/dao/impl/OrgDisambiguatedDaoImpl.java
```diff
@@ -87,6 +85,19 @@ public OrgDisambiguatedEntity findByNameCityRegionCountryAndSourceType(String na
         return query.getResultList();
     }
 
+    @Override
+    public List<OrgDisambiguatedEntity> findOrgsToGroup(int firstResult, int maxResult) {
+        TypedQuery<OrgDisambiguatedEntity> query = entityManager.createQuery("from OrgDisambiguatedEntity where source_type=:ROR or source_type=:FUNDREF or source_type=:RINGGOLD",
+                OrgDisambiguatedEntity.class);
+        //group by ROR
```
amontenegro, line 92: Why GRID orgs are not included in this query?

orcid-persistence/src/main/java/org/orcid/persistence/dao/impl/OrgDisambiguatedDaoImpl.java
```diff
@@ -87,6 +85,19 @@ public OrgDisambiguatedEntity findByNameCityRegionCountryAndSourceType(String na
         return query.getResultList();
     }
 
+    @Override
+    public List<OrgDisambiguatedEntity> findOrgsToGroup(int firstResult, int maxResult) {
+        TypedQuery<OrgDisambiguatedEntity> query = entityManager.createQuery("from OrgDisambiguatedEntity where source_type=:ROR or source_type=:FUNDREF or source_type=:RINGGOLD",
```
amontenegro, line 90: It looks like we should include another where clause so we don't try to regroup orgs that has been already grouped.

> and status !='PART_OF_GROUP'

Does that makes sense?

amontenegro (changes requested):
Please run the formatting rules in all the modified files.

orcid-persistence/src/main/java/org/orcid/persistence/dao/impl/OrgDisambiguatedDaoImpl.java
Camelia-Orcid: @angel  as per card https://trello.com/c/8kwEskUt/7825-include-ror-orgs-in-typeahead  and part of cleaning GRID I removed the GRID from the query as other organisations types that are "retired". If it was a misunderstanding I can easily add it back.

orcid-persistence/src/main/java/org/orcid/persistence/dao/impl/OrgDisambiguatedDaoImpl.java
Camelia-Orcid: I was thinking of a scenario when we want to regroup for some and re-run the cli cleaning script for everything, as this function is solely used for it.  I guess if that situation happens we can update the status to empty and re-run the script after. Changing it.

amontenegro (approved):

PR #6454, opened 2022-02-23
fix: Catch exception to return null instead of throwing an exception

DanielPalafox:


DanielPalafox:
[[QA] 500 error response when removing duplicate record that doesn't exist](https://trello.com/c/8LgHh5Cj/7960-qa-500-error-response-when-removing-duplicate-record-that-doesnt-exist)

PR #6455, opened 2022-02-24
Add xx/lr/rl translation keys

auumgn:


auumgn:
[Password length new keys are hardcoded](https://trello.com/c/N2Z0jpel/29-password-length-error-message-is-hardcoded)

PR #6464, opened 2022-03-03
New salesforce micro service

amontenegro:


orcid-core/src/main/java/org/orcid/core/manager/impl/SalesforceManagerImpl.java
```diff
@@ -0,0 +1,124 @@
...
+    public List<Member> retrieveMembers() {
+        List<Member> members = new ArrayList<Member>();
+
+        try {
+            String membersListString = client.retrieveMembers();
+            JSONObject membersListJsonObject = new JSONObject(membersListString);
+            members = salesForceAdapter.createMembersListFromJson(membersListJsonObject);
+        } catch (IOException | InterruptedException e) {
+            // TODO Auto-generated catch block
+            e.printStackTrace();
```
Camelia-Orcid, line 42: I think it makes sense to use log4j logger to log errors as we do through our source code ( lines 41, 45, 116 and 119)

orcid-core/src/main/java/org/orcid/core/salesforce/dao/impl/SalesforceMicroserviceClientImpl.java
```diff
@@ -0,0 +1,52 @@
...
+    public SalesforceMicroserviceClientImpl(@Value("${org.orcid.microservice.gateway.url}") String gatewayUrl,
+            @Value("${org.orcid.salesforce.api.timeout:5000}") Integer timeout) {
+        membersListEndpoint = gatewayUrl + "/members/list";
+        memberDetailsEndpoint = gatewayUrl + "/member/%s/details";
+        httpClient = HttpClient.newBuilder().version(HttpClient.Version.HTTP_2).connectTimeout(Duration.ofSeconds(timeout)).build();
+    }
+
+    @Override
+    public String retrieveMembers() throws IOException, InterruptedException {
+        // TODO: Handle exceptions
```
Camelia-Orcid, line 34: If the TODO is not required, maybe is a good idea to remove the comments in the code?

PR #6471, opened 2022-03-14
fix: add put codes list to peer reviews minimized

DanielPalafox:


DanielPalafox:
[[QA] Peer review visibility selector disappears after page reload](https://trello.com/c/8ar1gcbL/7985-qa-peer-review-visibility-selector-disappears-after-page-reload)

PR #6473, opened 2022-03-16
Make orcid case insensitive  when retrieving user details

Camelia-Orcid:


Camelia-Orcid:
[[QA] Lower case x in an ORCID iD leads to unsuccessful sign in](https://trello.com/c/DGi9mcMb/7998-qa-lower-case-x-in-an-orcid-id-leads-to-unsuccessful-sign-in)

PR #6491, opened 2022-03-31
Fixed the startIndex , added additional logs

Camelia-Orcid:


Camelia-Orcid:
[[QA] - Group organizations with same external identifier e.g ISNI](https://trello.com/c/naf1PcNu/8001-qa-group-organizations-with-same-external-identifier-eg-isni)

Camelia-Orcid:
[Organization typeahead to only show each organization once using ROR where available](https://trello.com/c/7Fg5THGe/7875-organization-typeahead-to-only-show-each-organization-once-using-ror-where-available)

PR #6506, opened 2022-04-20
feature: Add new internal API endpoint that find ORCID iD by email

DanielPalafox:


DanielPalafox:
[Create an API endpoint that takes email and returns ORCID iD. for use by ORCID API clients](https://trello.com/c/BGb9P5jt/8072-create-an-api-endpoint-that-takes-email-and-returns-orcid-id-for-use-by-orcid-api-clients)

PR #6507, opened 2022-04-21
fix: Retrieve only contributors with name or orcid

DanielPalafox:


DanielPalafox:
[[QA] Empty contributor section in production record](https://trello.com/c/8uhJERg7/8082-qa-empty-contributor-section-in-production-record)

PR #6512, opened 2022-04-26
Fixes for Trusted Parties  User menu

Camelia-Orcid:
@leomendoza123 can you add to the PR a the icon image ( I believe you already have it but is not commited) the src I see is /img/svg/vpn_key_FILL1_wght400_GRAD0_opsz20.svg therefore the image should be /orcid-web/src/main/webapp/static/img/svg/ along with others . Thanks.


Camelia-Orcid:
[[QA] Trusted parties page cannot be accessed from old app pages](https://trello.com/c/nfaKd15j/8086-qa-trusted-parties-page-cannot-be-accessed-from-old-app-pages)

PR #6513, opened 2022-04-26
Fix trusted parties icon

leomendoza123:


leomendoza123:
[[QA] Trusted parties page cannot be accessed from old app pages](https://trello.com/c/nfaKd15j/8086-qa-trusted-parties-page-cannot-be-accessed-from-old-app-pages)

PR #6515, opened 2022-04-27
fix/enable-togglz-for-trusted-indviduals

leomendoza123:


orcid-web/src/main/java/org/orcid/frontend/web/controllers/ManageProfileController.java
```diff
@@ -174,17 +174,7 @@ public ModelAndView manageProfile() {
     
     @RequestMapping(value = "/addDelegate.json")
     public @ResponseBody ManageDelegate addDelegate(@RequestBody ManageDelegate addDelegate) {
-        // Check password
-        String password = addDelegate.getPassword();
-        ProfileEntity profile = profileEntityCacheManager.retrieve(getCurrentUserOrcid());
-        if (orcidSecurityManager.isPasswordConfirmationRequired()
```
leomendoza123, line 180: Not require to call `isPasswordConfirmationRequired` since this function will be allow for account delegates :

"Delegates can now add trusted individuals to a managed record. This is a change in behaviour from the current version of the UI." https://trello.com/c/iKBm6d0V/7903-make-it-obvious-what-you-can-and-cant-do-as-an-account-delegate-in-another-users-record

leomendoza123:
[Make it obvious what you can and can't do as an account delegate in another users record](https://trello.com/c/iKBm6d0V/7903-make-it-obvious-what-you-can-and-cant-do-as-an-account-delegate-in-another-users-record)

PR #6516, opened 2022-04-27
Fix restricted delegators togglz

leomendoza123:


leomendoza123:
A small fix to the previous https://github.com/ORCID/ORCID-Source/pull/6515 RP

PR #6522, opened 2022-05-03
8085 contributors store top 50 contributors in a different field

DanielPalafox:


orcid-persistence/src/main/java/org/orcid/persistence/dao/impl/WorkDaoImpl.java
```diff
@@ -397,7 +399,24 @@ public List<Object[]> getWorksByOrcid(String orcid) {
                 .addScalar("assertion_origin_client_source_id", StringType.INSTANCE)
                 .addScalar("date_created", TimestampType.INSTANCE)
                 .addScalar("last_modified", TimestampType.INSTANCE)
-                .addScalar("top_contributors", StringType.INSTANCE);
+                .addScalar("contributors", StringType.INSTANCE)
```
amontenegro, line 402: Please add a //TODO so in the future, after the CLI finishes, we fetch only the top_contributors_json

orcid-core/src/main/java/org/orcid/core/manager/v3/impl/WorkManagerImpl.java
```diff
@@ -494,5 +518,15 @@ private WorkEntity createCopyOfUserPreferredWork(MinimizedWorkEntity preferred)
         workEntity.setWorkUrl(preferredFullData.getWorkUrl());
         return workEntity;
     }
+
```
amontenegro, line 521: Please add a unit test that ensures that we store the top 50 contributors when an API request comes with more than 50; please make it a mock test so you dont need database here becase hsqldb doesnt work well with json fields

amontenegro (changes requested):

amontenegro (approved):

PR #6530, opened 2022-05-05
fix: Oauth request and prompt login if parameter is present

DanielPalafox:


DanielPalafox:
[[QA] User that has already granted permission is not taken to signin when visiting prompt=login OAuth link](https://trello.com/c/mEnTE96x/8081-qa-user-that-has-already-granted-permission-is-not-taken-to-signin-when-visiting-promptlogin-oauth-link)

PR #6533, opened 2022-05-10
fix: Return `Name is private` instead of empty CreditName. Translatio…

DanielPalafox:
…n already exists in orcid-angular


DanielPalafox:
[Placeholder for private names in contributor lists](https://trello.com/c/rQnsZQbc/8094-placeholder-for-private-names-in-contributor-lists)

PR #6538, opened 2022-05-13
Catch the invalid redirect uri exception serverside

amontenegro:


amontenegro:
[[QA] 500 response from init.json when using unregistered redirect_uri](https://trello.com/c/iOoD4UI3/8128-qa-500-response-from-initjson-when-using-unregistered-redirecturi)

PR #6545, opened 2022-05-17
fix: Display UI message more contributors

DanielPalafox:


DanielPalafox:
[[QA] Contributor count is missing from contributor list in works](https://trello.com/c/CTYt6vf5/8136-qa-contributor-count-is-missing-from-contributor-list-in-works)

PR #6546, opened 2022-05-17
fix: Add missing number of contributors when top contributors togglz …

DanielPalafox:
…is enabled


DanielPalafox:
[[QA] Contributor count is missing from contributor list in works](https://trello.com/c/CTYt6vf5/8136-qa-contributor-count-is-missing-from-contributor-list-in-works)

PR #6551, opened 2022-05-20
Trim the redirect URI

amontenegro:


amontenegro:
[[QA] Whitespace in registered redirect URIs has to be ignored](https://trello.com/c/DruCIumW/8148-qa-whitespace-in-registered-redirect-uris-has-to-be-ignored)

PR #6552, opened 2022-05-24
reverted pom to the release 2.13.16

Camelia-Orcid:
Jersey library we use is not compatible at runtime with jdk11 config in pom. In order to successfully do the jackson+ swagger lib upgrade and set the maven-compiler-plugin configuration to jdk11, there is the need to upgrade jersey to jersey 2 (https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest/migration.html#mig-1.x ) most probably we need to rework some of the code will not work out of the box upgrading to jersey 2. I will create a separate card for the upgrade.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@ the software.
         <security.version>5.6.2</security.version>
         <liquibase.version>4.9.1</liquibase.version>
         <jersey.version>1.19.1</jersey.version>
-        <swagger.version>1.6.6</swagger.version>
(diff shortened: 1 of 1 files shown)


Camelia-Orcid:
[[QA]- papi, mapi, iapi not working with the latest pom](https://trello.com/c/5zU4dr48/8158-qa-papi-mapi-iapi-not-working-with-the-latest-pom)

amontenegro:
Thanks a lot @Camelia-Orcid !

PR #6567, opened 2022-06-06
Fixed the lockout check when the lock window exceeded as well as reset password

Camelia-Orcid:


Camelia-Orcid:
[[QA] Clear lockout after succesfully reseting password](https://trello.com/c/tnwstWeK/37-qa-clear-lockout-after-succesfully-reseting-password)

PR #6569, opened 2022-06-08
fix: Update orcid-model to support affiliation types v3 in notificati…

DanielPalafox:
Do not merge until [orcid-model](https://oss.sonatype.org/#nexus-search;quick~org.orcid) 3.0.11  is published


DanielPalafox:
[Allow the new affiliation types to be used for permission notifications](https://trello.com/c/QoVoNjnO/8150-allow-the-new-affiliation-types-to-be-used-for-permission-notifications)

PR #6577, opened 2022-06-16
8172-update-the-prefix-identifier-for-cstr

leomendoza123:


leomendoza123:
[Update the prefix identifier for CSTR](https://trello.com/c/E9Sw9ckk/8172-update-the-prefix-identifier-for-cstr)

amontenegro (changes requested):
@leomendoza123 please change the name of the file, it should be something like fix-cstr-id_resolution_prefix.xml and it should be added to the db-master.xml file so liquibase can pick it.

PR #6582, opened 2022-06-24
Log error only  if not  NoResultException

Camelia-Orcid:


Camelia-Orcid:
[Add catch for NoEntity Exception to reduce the noise](https://trello.com/c/COBUF3ai/8191-add-catch-for-noentity-exception-to-reduce-the-noise)

PR #6583, opened 2022-06-27
8188-return-a-406-when-calling-unsupported-media-types

leomendoza123:


leomendoza123:
[Return a 406 when calling unsupported media types](https://trello.com/c/MkuED0Wm/8188-return-a-406-when-calling-unsupported-media-types)

PR #6584, opened 2022-06-28
Added the configurable queries for bulk organisation grouping

Camelia-Orcid:
This assures all RORs  that have a fundref are being grouped and we have customisable query to feed the organisations to  ProcessDisambiguatedOrgsAsPartOfRorGroupToReindex cli


Camelia-Orcid:
The query in staging.properties has been already ran in QA env.

PR #6587, opened 2022-06-30
build: Rollback spring security

DanielPalafox:
The latest version of spring security is accepting only readable ASCII codes in the URL which causes a problem when you are looking for an element with an unicode i.e `Université de Montréal`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,7 +26,7 @@ the software.
 
     <properties>
         <spring.version>5.3.20</spring.version>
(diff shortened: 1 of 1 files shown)


DanielPalafox:
[[QA] 500 error response when entering non-English characters in organization typeahead](https://trello.com/c/g5GFXPLl/8197-qa-500-error-response-when-entering-non-english-characters-in-organization-typeahead)

amontenegro (approved):

PR #6589, opened 2022-07-05
fix: Update last modified when token is revoked

DanielPalafox:


DanielPalafox:
[Last updated is not being updated when tokens are revoked](https://trello.com/c/ZWzCz8KQ/8175-last-updated-is-not-being-updated-when-tokens-are-revoked)

amontenegro (approved):

PR #6593, opened 2022-07-05
Rolling back the token expiration check so the member portal can work…

amontenegro:
… with implicit flow


amontenegro:
[OBO tokens should work even if they are expired.](https://trello.com/c/mx0RctKu/8208-obo-tokens-should-work-even-if-they-are-expired)

PR #6598, opened 2022-07-13
fix: Add missing record locked description

DanielPalafox:


DanielPalafox:
[[QA] Lookup admin function, lock description is missing for locked records](https://trello.com/c/OIc6mZuN/8219-qa-lookup-admin-function-lock-description-is-missing-for-locked-records)

PR #6603, opened 2022-07-20
fix: Remove `unspecified` contributor role

DanielPalafox:


DanielPalafox:
[Allow users to set their own contributor roles](https://trello.com/c/0PMTWQS4/7321-allow-users-to-set-their-own-contributor-roles)

amontenegro (approved):

PR #6605, opened 2022-07-25
fix: Add contributor even if does not have roles

DanielPalafox:
[[QA] Malformed API response when reading work created in UI containing contributor without role](https://trello.com/c/7KRPyUR3/8234-qa-malformed-api-response-when-reading-work-created-in-ui-containing-contributor-without-role)


DanielPalafox:
[[QA] Malformed API response when reading work created in UI containing contributor without role](https://trello.com/c/7KRPyUR3/8234-qa-malformed-api-response-when-reading-work-created-in-ui-containing-contributor-without-role)

PR #6610, opened 2022-08-08
Adding a header to each call the ML does to our apis

amontenegro:


amontenegro:
[Add headers to the message listener requests so we know it is the message listener](https://trello.com/c/Hj6wMoAP/8241-add-headers-to-the-message-listener-requests-so-we-know-it-is-the-message-listener)

leomendoza123 (approved):

PR #6611, opened 2022-08-11
feature: Add new endpoint to retrieve only employment affiliations

DanielPalafox:


DanielPalafox:
[Only show current employment affiliations in contributor listing](https://trello.com/c/IpDuPKh4/8240-only-show-current-employment-affiliations-in-contributor-listing)

PR #6628, opened 2022-09-08
Get the last ROR dump available , after  ROR metadata and versioning  change.

Camelia-Orcid:
As per ROR recommendation now we need to do the following to retrieve the latest file (https://ror.readme.io/docs/data-dump#download-ror-data-dumps-programmatically-with-the-zenodo-api) https://zenodo.org/api/records/?communities=ror-data&sort=mostrecent&size=1
In the response, the most recent record will be hits.hits[0]. Find the latest file attached to this record by checking the last item in hits.hits[0].files. The file URL is in links.self within the file object

I added a cli to be able to load orgs for a type at any time manually  without the need to wait for the weekly schedule. To trigger a load  organization for a specific org type  ( -o can have the supported values FUNDREF, ROR or RINGGOLD)

mvn exec:java -Dexec.args="-o ROR" -Djava.io.tmpdir="/Users/camelia/OrcidDev/github/jdk11/ORCID-Source/orcid-core/tmp" -Dexec.mainClass="org.orcid.core.cli.LoadDataForOrganizationSource" -Dorg.orcid.config.file="file:///Users/camelia/OrcidDev/github/jdk11/ORCID-Source/orcid-core/tmp/orcid.properties" -Dorg.orcid.persistence.messaging.brokerURL=tcp://127.0.0.1:61616?jms.useAsyncSend=true -Dorg.orcid.persistence.messaging.enabled=true

### DIFF
--- a/orcid-core/src/main/java/org/orcid/core/cli/LoadDataForOrganizationSource.java
+++ b/orcid-core/src/main/java/org/orcid/core/cli/LoadDataForOrganizationSource.java
@@ -1,0 +1,87 @@
+package org.orcid.core.cli;
+
+
+import org.apache.commons.lang.StringUtils;
+import org.kohsuke.args4j.CmdLineException;
+import org.kohsuke.args4j.CmdLineParser;
+import org.kohsuke.args4j.Option;
+import org.orcid.core.orgs.load.manager.OrgLoadManager;
+import org.orcid.core.orgs.load.source.OrgLoadSource;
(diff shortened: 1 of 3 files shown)


Camelia-Orcid:
[ROR importer is not picking the last available dump](https://trello.com/c/ZsqPWmqM/8269-ror-importer-is-not-picking-the-last-available-dump)

amontenegro (approved):

PR #6630, opened 2022-09-12
orcid-model to 3.0.12

amontenegro:


amontenegro:
[Add new scope /member-list/full-access](https://trello.com/c/GI8G7sdD/8281-add-new-scope-member-list-full-access)

PR #6635, opened 2022-09-15
fix: Set publication date to null if there is no publication year to …

DanielPalafox:
…fix publication date comparator


DanielPalafox:
[Works without a publication date should show after works with a publication date](https://trello.com/c/zGg64vhg/8277-works-without-a-publication-date-should-show-after-works-with-a-publication-date)

amontenegro (approved):

PR #6637, opened 2022-09-20
Anonymizing works

Camelia-Orcid:


Camelia-Orcid:
[Script to copy and anonymize records](https://trello.com/c/Hh9tBjJa/8255-script-to-copy-and-anonymize-records)

Camelia-Orcid:
[[QA] When anonymize the external id it needs to be kept the same across all the works](https://trello.com/c/msksNHLZ/8301-qa-when-anonymize-the-external-id-it-needs-to-be-kept-the-same-across-all-the-works)

amontenegro (approved):

PR #6638, opened 2022-09-20
fix: NPE roles and sequences array list and add missing converter test

DanielPalafox:


DanielPalafox:
[[QA] Works imported from BibTeX not visible, 500 error response](https://trello.com/c/LpADtJWR/8294-qa-works-imported-from-bibtex-not-visible-500-error-response)

amontenegro (approved):

PR #6648, opened 2022-10-17
orcid-core and orcid-persistence must not use the orcid-utils package…

amontenegro:
…, so there are no dependencies with the new jersey client


amontenegro:
[[QA] Website search not working](https://trello.com/c/Z2jIFyf3/8324-qa-website-search-not-working)

PR #6650, opened 2022-10-18
We dont send emails from the message listener, so, we can ignore the …

amontenegro:
…bean creation


amontenegro:
[[QA] Indexing to Solr not working](https://trello.com/c/hVwKKVln/8337-qa-indexing-to-solr-not-working)

PR #6659, opened 2022-10-25
Disable using subdomains as redirect uris

amontenegro:


amontenegro:
[Reinstate matching against subdomains when validating redirect URIs](https://trello.com/c/2ORgqS1N/8330-reinstate-matching-against-subdomains-when-validating-redirect-uris)

Camelia-Orcid (approved):

PR #6673, opened 2022-11-11
If the work was created through bibtex we dont care if it contains th…

amontenegro:
…e record owner as contributor


amontenegro:
[Bibtex import not working when togglz is on](https://trello.com/c/8FZdXx1Z/8367-bibtex-import-not-working-when-togglz-is-on)

leomendoza123 (approved):

PR #6675, opened 2022-11-15
Fixed the locale to match the list available in the  UI dropdown

Camelia-Orcid:


Camelia-Orcid:
[Users with CS local cant reset their password](https://trello.com/c/oui1KcV0/8368-users-with-cs-local-cant-reset-their-password)

Camelia-Orcid:
[Reset password fails for records with `š` in the name](https://trello.com/c/RWrLABgC/8369-reset-password-fails-for-records-with-%C5%A1-in-the-name)

amontenegro (approved):

PR #6677, opened 2022-11-17
8271-create-an-admin-function-to-move-a-client-from-one-group-id-to-a…

leomendoza123:
…nother

### DIFF
--- a/orcid-core/src/main/java/org/orcid/core/manager/v3/impl/ClientDetailsManagerImpl.java
+++ b/orcid-core/src/main/java/org/orcid/core/manager/v3/impl/ClientDetailsManagerImpl.java
@@ -455,7 +455,7 @@ public class ClientDetailsManagerImpl extends ClientDetailsManagerReadOnlyImpl i
                     clientScopeDao.insertClientScope(clientId, newScope);
                 }
             }
-            LOGGER.info("Client {} was succesfully move to the groupId {} with the type {}", clientId, groupId, clientType);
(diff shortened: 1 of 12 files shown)


leomendoza123:
[Create an admin function to move a client from one group ID to another](https://trello.com/c/bW0ct0qN/8271-create-an-admin-function-to-move-a-client-from-one-group-id-to-another)

orcid-core/src/main/java/org/orcid/core/manager/v3/impl/ClientDetailsManagerImpl.java
```diff
@@ -421,5 +421,45 @@ public void convertPublicClientToMember(String clientId, String groupId) {
...
+            // clientDetailsDao.updateClientGrantedAuthority(clientId, "ROLE_CLIENT");
+            // Assign scopes to client
+            List<String> clientScopes = clientScopeDao.getActiveScopes(clientId);
+            Set<String> newScopes = null;
+            if (clientType.equals(ClientType.PREMIUM_UPDATER)) {
+                newScopes = ClientType.getScopes(ClientType.PREMIUM_UPDATER);
+            } else {
+                newScopes = ClientType.getScopes(ClientType.UPDATER);
+            }
+
```
amontenegro, line 446: Lets add some comments here explaining why you should delete some scopes and then add others .... Because you and me understand it but maybe not others

amontenegro (changes requested):

amontenegro (approved):

PR #6679, opened 2022-11-18
Fix bug on bulk work create

amontenegro:


amontenegro:
[Bulk works endpoint doesn't allow posting a work with a "self" ISBN if the identifier is already used in another work as "part-of"](https://trello.com/c/vT6t4vVG/8358-bulk-works-endpoint-doesnt-allow-posting-a-work-with-a-self-isbn-if-the-identifier-is-already-used-in-another-work-as-part-of)

PR #6696, opened 2022-12-21
8419-qa-no-error-message-when-trying-to-move-client-id-to-invalid-gro…

leomendoza123:
…up-id


leomendoza123:
[[QA] No error message when trying to move client ID to invalid group ID](https://trello.com/c/kYkulncR/8419-qa-no-error-message-when-trying-to-move-client-id-to-invalid-group-id)

PR #6697, opened 2022-12-22
fix: Set work type to `dissertation_thesis` when `dissertation` or `t…

DanielPalafox:
…hesis` values are passed and fix JSONException in issued


amontenegro:
[[QA] Work cannot be added via `Add DOI` with issued value as null](https://trello.com/c/1BgNsfkM/8423-qa-work-cannot-be-added-via-add-doi-with-issued-value-as-null)

amontenegro (changes requested):

amontenegro (approved):

PR #6701, opened 2022-12-26
Fix move not active client

leomendoza123:


leomendoza123:
[[QA] Do not allow moving deactivated Client IDs from one Group ID to another](https://trello.com/c/egjjCgHq/8424-qa-do-not-allow-moving-deactivated-client-ids-from-one-group-id-to-another)

amontenegro (approved):

PR #6706, opened 2023-01-03
Community type enum is going back to the core package because there a…

amontenegro:
…re i18n properties that will be hard to update if we change the package name


amontenegro:
[Remove old salesforce integration code.](https://trello.com/c/0YvDkloZ/8428-remove-old-salesforce-integration-code)

PR #6714, opened 2023-01-13
More reliable delete process that will sleep for 10 secs, retry and m…

amontenegro:
…ark as failed only after retry


amontenegro:
[[QA] Activities deleted in UI not deleted in AWS S3](https://trello.com/c/SDflsG9y/8449-qa-activities-deleted-in-ui-not-deleted-in-aws-s3)

leomendoza123 (approved):

PR #6728, opened 2023-02-08
feature: Compare work to be updated with the one stored in the database

DanielPalafox:


amontenegro (changes requested):
@DanielPalafox please do the following changes: 

- Update the code so it also perform the same validation for the API 2.0 and 2.1 
- Update the code so the validation is in the manager level, so, you can do it for the UI and the API in the same place

amontenegro (approved):

PR #6731, opened 2023-02-10
Remove invalid profileEntity.id reference

amontenegro:


amontenegro:
[[QA] OBO and openId are broken](https://trello.com/c/fm0fINWn/8486-qa-obo-and-openid-are-broken)

PR #6734, opened 2023-02-16
fix: Compare order of contributors and external identifiers

DanielPalafox:


DanielPalafox:
[FOR WORKS: Stop sending notifications when no updates to the works has actually been made](https://trello.com/c/PiSAzXoZ/8437-for-works-stop-sending-notifications-when-no-updates-to-the-works-has-actually-been-made)

PR #6735, opened 2023-02-16
Code that compiles with jersey 2.27  and loads the jersey context

Camelia-Orcid:
Known  Issues: Response from jersey server not returned properly due to InjectParam change. Some test cases fail because the Spring context gets loaded before some listeners. Swagger commented out until APIs are working ( jersey 2.27 is not working with swagger 1.5.16; need to find the 2.x swagger that works with jersey 2.27)


Camelia-Orcid:
[Update jersey server to 3.0.5](https://trello.com/c/qNy3Usbd/8362-update-jersey-server-to-305)

amontenegro (approved):

PR #6738, opened 2023-02-23
Fixed the relationship for the same external identifiers with different relationships for various works

Camelia-Orcid:


Camelia-Orcid:
[[Anonymize][QA] Script to anonymize data - identifier relationship types should persist as per original record](https://trello.com/c/6471jUd3/8378-anonymizeqa-script-to-anonymize-data-identifier-relationship-types-should-persist-as-per-original-record)

amontenegro (approved):

PR #6744, opened 2023-03-15
Changed the config of jersey to get it from filter, to match the mapi and public apis

Camelia-Orcid:


Camelia-Orcid:
[[QA] iapi is not starting in QA](https://trello.com/c/6JSjcIxp/8531-qa-iapi-is-not-starting-in-qa)

amontenegro (approved):

PR #6747, opened 2023-03-17
qs changed to be in the range of 0 to 1 as in specs

Camelia-Orcid:
qs values can vary in the range 0.000 to 1.000. Note that any variant with a qs value of 0.000 will never be chosen. Variants with no 'qs' parameter value are given a qs factor of 1.0. The qs parameter indicates the relative 'quality' of this variant compared to the other available variants, independent of the client's capabilities. For example, a JPEG file is usually of higher source quality than an ASCII file if it is attempting to represent a photograph. However, if the resource being represented is an original ASCII art, then an ASCII representation would have a higher source quality than a JPEG representation. A qs value is therefore specific to a given variant depending on the nature of the resource it represents.


Camelia-Orcid:
[[QA] API cannot handle some XML and JSON content types](https://trello.com/c/DCN0bsbO/8530-qa-api-cannot-handle-some-xml-and-json-content-types)

PR #6749, opened 2023-03-17
feature: Compare names, other names and biography with stored in db a…

DanielPalafox:
…nd only trigger update if there is changes


DanielPalafox:
[Record last_modified date is updated if I save names modal without making changes](https://trello.com/c/1cyXdn6V/8484-record-lastmodified-date-is-updated-if-i-save-names-modal-without-making-changes)

amontenegro (approved):

PR #6755, opened 2023-03-21
Copy the immutable params map so we can modify it

amontenegro:


amontenegro:
[[QA] [JERSEY] Search is not working](https://trello.com/c/NMSKp2Pt/8544-qa-jersey-search-is-not-working)

PR #6766, opened 2023-04-05
Added the xslt header

Camelia-Orcid:
Still not working cause it returns a 404 for the https://pub.dev.orcid.org/static/identifierTypes.xsl  . It looks like is not picking up jersey.config.servlet.filter.staticContentRegex


Camelia-Orcid:
After talking to Angel, as the table is showing and the xsl resolved when using local host, it might be just a problem with the dev ngnix setup.

amontenegro (approved):

PR #6769, opened 2023-04-11
ROR uploader changes to take in account the external Identifiers updates

Camelia-Orcid:


Camelia-Orcid:
[Fix fundref->ROR matching (e.g. Duplicate organization: "Deutsches Museum")](https://trello.com/c/SRtNe0ah/8563-fix-fundref-ror-matching-eg-duplicate-organization-deutsches-museum)

PR #6770, opened 2023-04-11
fix: Reverse list to compare other names stored in the database

DanielPalafox:


DanielPalafox:
[[QA] Record last_modified date updated if names modal is saved without changes on records with other names/published name](https://trello.com/c/1gxiThqw/8568-qa-record-lastmodified-date-updated-if-names-modal-is-saved-without-changes-on-records-with-other-names-published-name)

amontenegro (approved):

PR #6789, opened 2023-05-09
Check if the scopes are empty and throw an exception

amontenegro:


amontenegro:
[[QA] 500 error response from authorize.json for OAuth requests containing empty scope parameter](https://trello.com/c/IAWfEJD7/8600-qa-500-error-response-from-authorizejson-for-oauth-requests-containing-empty-scope-parameter)

PR #6800, opened 2023-05-24
Remove all togglz that have to do with the angular pages

amontenegro:


amontenegro:
[[Tech][Togglz] Remove togglz for angular pages](https://trello.com/c/GVK0z0Sb/8630-techtogglz-remove-togglz-for-angular-pages)

PR #6822, opened 2023-06-19
Add a property to indicate if a record should have the noindex metatag

amontenegro:


amontenegro:
[no-index is not being set for records without an OAuth token](https://trello.com/c/5z6Nv0KH/51-no-index-is-not-being-set-for-records-without-an-oauth-token)

leomendoza123 (approved):

leomendoza123:
[Account: Developer tools](https://trello.com/c/1lf82w2p/8661-account-developer-tools)

PR #6831, opened 2023-07-14
8682 trust summary json format

DanielPalafox:


DanielPalafox:
[Trust summary - JSON format](https://trello.com/c/ZZG8OhW9/8682-trust-summary-json-format)

orcid-web/src/main/java/org/orcid/frontend/web/controllers/PublicRecordController.java
```diff
@@ -258,6 +315,203 @@ PublicRecord getRecord(String orcid) {
...
+            recordSummary.setName(localeManager.resolveMessage("orcid.reserved_for_claim"));
+            return recordSummary;
+        } catch (OrcidDeprecatedException e) {
+            isDeprecated = true;
+        } catch (OrcidNoResultException e) {
+            return recordSummary;
+        }
+
+        recordSummary = getSummary(orcid);
+        if (isDeprecated) {
```
amontenegro, line 341: Hey @DanielPalafox, lets make this check before the `getSummary(orcid)` so we dont waste resources fetching the info if the record is deprecated

amontenegro: And add an else clause where you call the getSummary in case it is not deprecated

amontenegro (changes requested):

orcid-web/src/main/java/org/orcid/frontend/web/controllers/PublicRecordController.java
DanielPalafox: Done!

PR #6836, opened 2023-07-25
fix: One trust style on print preview

DanielPalafox:


DanielPalafox:
[Onetrust styling broken on print preview](https://trello.com/c/FItB1dlt/8736-onetrust-styling-broken-on-print-preview)

PR #6846, opened 2023-08-03
Ror to fundref ringgold mapping

Camelia-Orcid:


Camelia-Orcid:
[Produce a one off CSV file of the mapping between ROR and Funder Registry IDs](https://trello.com/c/FnA6cJbI/8740-produce-a-one-off-csv-file-of-the-mapping-between-ror-and-funder-registry-ids)

PR #6847, opened 2023-08-03
notify.orcid.org and verify.orcid.org should be not reply addresses

amontenegro:


amontenegro:
[Update email addresses used by the registry to be no reply email addresses](https://trello.com/c/yfK6k1cL/8735-update-email-addresses-used-by-the-registry-to-be-no-reply-email-addresses)

Camelia-Orcid (approved):

PR #6848, opened 2023-08-03
Check the source of the element based on the source ids and not the name

amontenegro:


amontenegro:
[[QA] Validated activities are shown as self-asserted in the summary page](https://trello.com/c/8k1N6VXM/8783-qa-validated-activities-are-shown-as-self-asserted-in-the-summary-page)

leomendoza123 (approved):

PR #6856, opened 2023-08-09
fix: Remove field from response if its private

DanielPalafox:


DanielPalafox:
[[QA] Return 'name' as undefined when the name is private or limited](https://trello.com/c/SLVLVSFq/8797-qa-return-name-as-undefined-when-the-name-is-private-or-limited)

amontenegro (approved):

PR #6863, opened 2023-08-10
fix: Add commonName and reference in ExternalIdentifiersSummary

DanielPalafox:


DanielPalafox:
[[QA] External identifier URL is shown in summary page instead of external identifier value](https://trello.com/c/djZ73BEb/8790-qa-external-identifier-url-is-shown-in-summary-page-instead-of-external-identifier-value)

amontenegro (approved):

PR #6874, opened 2023-08-18
Trim org data

amontenegro:


amontenegro:
[Add "; charset=utf8" to content type in http headers to prevent ISO text being posted to the server](https://trello.com/c/LZVOoi2c/8801-add-charsetutf8-to-content-type-in-http-headers-to-prevent-iso-text-being-posted-to-the-server)

leomendoza123 (approved):

PR #6876, opened 2023-08-22
Remove token from cache when it is revoked

amontenegro:


amontenegro:
[[QA] Remove tokens from cache when they are evicted](https://trello.com/c/w0k3DuUz/8826-qa-remove-tokens-from-cache-when-they-are-evicted)

PR #6887, opened 2023-09-04
If a record was member created and the member added activities to it …

amontenegro:
…on creation time, the record should be indexed


amontenegro:
[[Tech][R&D] Google indexing improvements](https://trello.com/c/bOfeGoF9/8805-techrd-google-indexing-improvements)

leomendoza123 (approved):

PR #6891, opened 2023-09-14
If the sent date is null, we should not care about webhook creation d…

amontenegro:
…ate vs profile last modified


amontenegro:
[[PROD] [Webhooks] The webhook process is not able to cope with the current rate and we are queueing webhooks](https://trello.com/c/udELvo9C/920-prod-webhooks-the-webhook-process-is-not-able-to-cope-with-the-current-rate-and-we-are-queueing-webhooks)

amontenegro:
There is no webhooks queue, it was a confusion with the webhooks that have never been sent before that made us to think that there was a growing queue.

PR #6892, opened 2023-09-20
fix: Validate that availability and documentStyle exist

DanielPalafox:


orcid-core/src/main/java/org/orcid/core/utils/v3/identifiers/resolvers/PubMedResolver.java
```diff
@@ -206,8 +206,15 @@ private WorkExtended getWork(JSONObject json) throws JSONException, ParseExcepti
...
+                                urlType = url.getString("documentStyle");
+                            }
+                            String availability = null;
+                            if (url.has("availability")) {
+                                availability = url.getString("availability");
+                            }
+
                             // If we find the html link, use it and stop
                             // searching
                             if (urlType.equals("html")) {
```
amontenegro, line 220: @DanielPalafox we should add a null check on the urlType

amontenegro (changes requested):

orcid-core/src/main/java/org/orcid/core/utils/v3/identifiers/resolvers/PubMedResolver.java
DanielPalafox: Done

amontenegro (approved):

PR #6901, opened 2023-10-04
URL comparison should be case sensitive

amontenegro:


amontenegro:
[[QA] Items in the websites section are converted to lowercase when editing](https://trello.com/c/4w3HGRGC/8830-qa-items-in-the-websites-section-are-converted-to-lowercase-when-editing)

amontenegro:
Lets leave this one for the next sprint.

PR #6918, opened 2023-10-27
8889 prod looks like the ror loader is broken

Camelia-Orcid:


Camelia-Orcid:
[[PROD] Looks like the ROR loader is broken?](https://trello.com/c/1kKNhTk7/8889-prod-looks-like-the-ror-loader-is-broken)

amontenegro (approved):

PR #6923, opened 2023-10-31
Welcome email ready to go

amontenegro:


amontenegro:
[Updates to the email verification email text](https://trello.com/c/1dKzcgO8/8897-updates-to-the-email-verification-email-text)

orcid-core/src/main/resources/i18n/email_subject_en.properties
```diff
@@ -2,7 +2,7 @@
 # NotificaionManager.java
 
 email.subject.verify_reminder=[ORCID] Verify your email address
-email.subject.verify_reminder_primary=[ORCID] Reminder to verify your primary email address
+email.subject.verify_reminder_primary=[ORCID] Don't forget to verify your email address
```
leomendoza123, line 5: Please note that since `email.subject.verify_reminder_primary` has been modified, other languages will retain the translations for the previous string until we complete a translation cycle.

This is why I generally recommend not modifying existing keys. Instead, I prefer to delete old keys and create new ones.

This might not be a significant issue right now since the changes are minor.

leomendoza123 (approved):

PR #6925, opened 2023-11-03
Is there is more than one org with no parent, we should not suggest a…

amontenegro:
…ny org for that domain


amontenegro:
[[QA] /email-domain endpoint returns org details when it shouldn't for email domain with multiple orgs](https://trello.com/c/6msI20S2/8899-qa-email-domain-endpoint-returns-org-details-when-it-shouldnt-for-email-domain-with-multiple-orgs)

PR #6934, opened 2023-11-16
Fix NPE on social sign in

amontenegro:


amontenegro:
[[QA] User lands in the old app doing social signin after a specific set of steps](https://trello.com/c/hP09irOD/8928-qa-user-lands-in-the-old-app-doing-social-signin-after-a-specific-set-of-steps)

DanielPalafox (approved):

PR #6942, opened 2023-12-03
feature: Add affiliation when a new user registers

DanielPalafox:


orcid-core/src/main/java/org/orcid/core/manager/impl/RegistrationManagerImpl.java
```diff
@@ -290,7 +297,12 @@ private String createMinimalProfile(Registration registration, boolean usedCaptc
             name.setGivenNames(new GivenNames(registration.getGivenNames().getValue()));
         }
         recordNameManager.createRecordName(orcid, name);
-                
+
+        if (Features.EVENTS.isActive() && registration.getAffiliationForm() != null) {
```
amontenegro, line 301: @DanielPalafox is this the right feature for this `Features.EVENTS`? I think this one is to track user events right?

orcid-web/src/main/java/org/orcid/frontend/web/controllers/RegistrationController.java
```diff
@@ -321,6 +329,28 @@ public void validateRegistrationFields(HttpServletRequest request, Registration
         regEmailValidate(request, reg, false, false);
         registerTermsOfUseValidate(reg);
 
+        if (Features.EVENTS.isActive() && reg.getAffiliationForm() != null) {
```
amontenegro, line 332: @DanielPalafox again, I think the feature `Features.EVENTS` is not the one we want to use here, could you please create one, something like `Features.AFFILIATIONS_ON_REGISTRATION`

amontenegro (changes requested):

orcid-core/src/main/java/org/orcid/core/manager/impl/RegistrationManagerImpl.java
DanielPalafox: Yes, you are right I will put the correct togglz. Thanks

orcid-web/src/main/java/org/orcid/frontend/web/controllers/RegistrationController.java
DanielPalafox: Done

amontenegro (approved):

PR #6946, opened 2023-12-07
Allow backend validation but not resolution

amontenegro:


amontenegro:
[Disable the WorldCat ISBN resolver from the works identifier section](https://trello.com/c/mK7azlDw/8939-disable-the-worldcat-isbn-resolver-from-the-works-identifier-section)

leomendoza123 (approved):

PR #6962, opened 2024-01-02
feat: Repurposed plain text status checks for the APIs into simple json check.

wjrsimpson:


wjrsimpson:
[[Tech] Update status checks so they work for monitoring tools](https://trello.com/c/XzdI7Gpw/1029-tech-update-status-checks-so-they-work-for-monitoring-tools)

amontenegro (approved):

PR #6965, opened 2024-01-03
fix: Add OL identifier case sensitive and RRID regex validator

DanielPalafox:


DanielPalafox:
[Fix the outstanding issues that we have with a small number of work identifiers](https://trello.com/c/UWCTuXqW/8984-fix-the-outstanding-issues-that-we-have-with-a-small-number-of-work-identifiers)

amontenegro (approved):

PR #6966, opened 2024-01-05
Remove few togglz from print view

amontenegro:


amontenegro:
[[PROD] Printable version page show a spinning wheel and never finish loading](https://trello.com/c/QNEtz5Ss/9012-prod-printable-version-page-show-a-spinning-wheel-and-never-finish-loading)

PR #6976, opened 2024-01-19
Allow to filter orgs that are not fundref

amontenegro:


amontenegro:
[[QA] Fundref org ids available in registration form typeahead](https://trello.com/c/cbLEqg7R/9032-qa-fundref-org-ids-available-in-registration-form-typeahead)

PR #6979, opened 2024-01-31
Filter by is-funding-org only on funding requests, and, filter by ROR…

amontenegro:
… in all affiliation requests


amontenegro:
[[QA] Fundref org ids available in registration form typeahead](https://trello.com/c/cbLEqg7R/9032-qa-fundref-org-ids-available-in-registration-form-typeahead)

leomendoza123 (approved):

PR #6987, opened 2024-02-07
fix: Remove global navigation from header

DanielPalafox:


DanielPalafox:
[[QA] Global navigation not removed from old app pages](https://trello.com/c/IIZrEmuP/9045-qa-global-navigation-not-removed-from-old-app-pages)

amontenegro (approved):

PR #6996, opened 2024-02-22
Added the check for put requests when the source is not the same

Camelia-Orcid:


Camelia-Orcid:
[[PROD] No-change PUT request behaviour](https://trello.com/c/UkMsfwDe/60-prod-no-change-put-request-behaviour)

amontenegro (approved):

PR #7000, opened 2024-02-27
Deactivated records should get 409 on GET requests

amontenegro:


amontenegro:
[Return a 409 error code for GET calls for deactivated iDs instead of a 200](https://trello.com/c/AWIytVnH/9016-return-a-409-error-code-for-get-calls-for-deactivated-ids-instead-of-a-200)

leomendoza123 (approved):

PR #7016, opened 2024-04-02
feature: Start collection daily pub api stats

DanielPalafox:


orcid-pub-web/src/main/java/org/orcid/api/publicV3/server/delegator/impl/PublicV3ApiServiceDelegatorImpl.java
```diff
@@ -961,4 +967,15 @@ public Response viewResearchResourceSummary(String orcid, Long putCode) {
         return Response.ok(e).build();
     }
 
+    @Override
+    public void trackEvents(HttpServletRequest httpRequest) {
+        boolean anonymous = false;
+        String clientId = orcidSecurityManager.getClientIdFromAPIRequest();
+        if (clientId == null) {
+            clientId = OrcidRequestUtil.getIpAddress(httpRequest);
```
amontenegro, line 975: @DanielPalafox the IP address should be stored in a different field, please create another DB field `ip` and store that value there; the IP should be stored even if the client id is provided

amontenegro (changes requested):

orcid-pub-web/src/main/java/org/orcid/api/publicV3/server/delegator/impl/PublicV3ApiServiceDelegatorImpl.java
DanielPalafox: Done

PR #7024, opened 2024-04-08
fix: Update cron job time and fix issue with select in postgres

DanielPalafox:


DanielPalafox:
[[QA] Aggregate stats not collected for Public API usage](https://trello.com/c/triKXB0z/9189-qa-aggregate-stats-not-collected-for-public-api-usage)

PR #7029, opened 2024-05-07
fix: Return attempted resolution as false if resolver is disabled

DanielPalafox:


DanielPalafox:
[Remove WorldCat warning message when adding ISBN to works](https://trello.com/c/5eb32S46/9227-remove-worldcat-warning-message-when-adding-isbn-to-works)

amontenegro (approved):

PR #7034, opened 2024-05-29
fix: Only collect public api anonymouse request over a 1000

DanielPalafox:


DanielPalafox:
[Only collect Public API anonymous client requests stats for clients that are making over 1000 requests each day](https://trello.com/c/PvABwmyT/9255-only-collect-public-api-anonymous-client-requests-stats-for-clients-that-are-making-over-1000-requests-each-day)

amontenegro (approved):

PR #7039, opened 2024-06-06
fix: Make validated items to be default preferred source for grouping

DanielPalafox:


DanielPalafox:
[Make validated items to be default preferred source for grouping](https://trello.com/c/apZ2fULk/9262-make-validated-items-to-be-default-preferred-source-for-grouping)

orcid-core/src/main/java/org/orcid/core/manager/v3/impl/WorkManagerImpl.java
```diff
@@ -477,9 +477,12 @@ public void createNewWorkGroup(List<Long> workIds, String orcid) throws MissingG
         MinimizedWorkEntity userPreferred = null;
         
         boolean groupableExternalIdFound = false;
-        for (MinimizedWorkEntity work : works) {
+        for (MinimizedWorkEntity work : works) {
```
amontenegro, line 480: @DanielPalafox could you please update this so it uses the [DisplayIndexCalculatorHelper.java](https://github.com/ORCID/ORCID-Source/pull/7039/files#diff-aeecd639e3870358cd80c013203b6999ee1c45e3a2865b29b160823e22baa6c7)? Thanks

amontenegro (changes requested):

orcid-core/src/main/java/org/orcid/core/manager/v3/impl/WorkManagerImpl.java
DanielPalafox: We only use the [DisplayIndexCalculatorHelper.java](https://github.com/ORCID/ORCID-Source/pull/7039/files#diff-aeecd639e3870358cd80c013203b6999ee1c45e3a2865b29b160823e22baa6c7) for new works and we assign the index depending on if the request was made via the member API or self-asserted, and this block of code it's for legacy works that have the indexing with the preferred source as the self-asserted so if a user wants to group works previously created the new preferred source will be the ones created by the member API

PR #7043, opened 2024-06-18
Changes for v2 fields stored in SOLR

Camelia-Orcid:


Camelia-Orcid:
[Update SOLR to include new data from the ROR importer now using V2.0](https://trello.com/c/LuPeudCr/9269-update-solr-to-include-new-data-from-the-ror-importer-now-using-v20)

amontenegro (approved):

orcid-persistence/src/main/resources/db/updates/org_disambiguated_extra_columns.xml
```diff
@@ -2,6 +2,14 @@
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
 
     <changeSet id="ADD-ORG-EXTRA-COLUMNS" author="Camelia Dumitru" dbms="postgresql">
+		 <preConditions onFail="MARK_RAN">
```
amontenegro, line 5: @Camelia-Orcid just a comment here, this will change the hash of the liquibase file, so, liquibase will complain about it the next time it runs in QA/Sbox and PROD. Do we really need to add this at this point? It already ran in all environments.

amontenegro (changes requested):

orcid-persistence/src/main/resources/db/updates/org_disambiguated_extra_columns.xml
Camelia-Orcid: Removed, this change was needed in dev only and committed by mistake.

amontenegro (approved):

PR #7050, opened 2024-07-13
Added the support for multiple org names  in the type ahead

Camelia-Orcid:


Camelia-Orcid:
[Update the affiliation typeahead to include 'Other Names' from ROR data](https://trello.com/c/SESCCSIa/9287-update-the-affiliation-typeahead-to-include-other-names-from-ror-data)

PR #7058, opened 2024-07-24
Update refresh_tokens.md

iamzoltan:
Looks like the urls did not match from the description and the curl commands. Does this resolve the issue?


amontenegro (approved):

amontenegro:
Good catch @iamzoltan, thanks a lot!

PR #7068, opened 2024-08-13
Set default visibility when more than one email have same domain

amontenegro:


amontenegro:
[[QA] Email Verify Flow - Domain visibility is not the less restrictive when there are more email siwht the same domain](https://trello.com/c/oGQ65BSK/9349-qa-email-verify-flow-domain-visibility-is-not-the-less-restrictive-when-there-are-more-email-siwht-the-same-domain)

PR #7074, opened 2024-08-27
Added unique constraint for external id identifier for disambiguated org

Camelia-Orcid:
This pool request has to be merged/deployed only after the coresponding database has been cleaned for duplicates.


Camelia-Orcid:
[Add Db constraint so only one external ID for the same type and identifier can be added for a disambiguated organization](https://trello.com/c/DaRrzHbR/9363-add-db-constraint-so-only-one-external-id-for-the-same-type-and-identifier-can-be-added-for-a-disambiguated-organization)

amontenegro:
@Camelia-Orcid this should be a draft right? because we should not merge it until we remove the duplicates from all envs right?

amontenegro (approved):

PR #7088, opened 2024-09-17
new email domains should have a default profile visibility

auumgn:


auumgn:
[New email domains should inherit record level default visibility](https://trello.com/c/ndYIyDoy/9399-new-email-domains-should-inherit-record-level-default-visibility)

amontenegro (approved):

PR #7091, opened 2024-09-27
Added the code to store deleted items in panoply dw_deleted_items table

Camelia-Orcid:
The items will be processed by the script that runs daily and cleaned up after the item is deleted from the corresponding view


Camelia-Orcid:
[Delete org_affiliation_relation rows in panoply for affiliations deleted in the registry](https://trello.com/c/F35BAqlO/9339-delete-orgaffiliationrelation-rows-in-panoply-for-affiliations-deleted-in-the-registry)

PR #7118, opened 2024-10-30
add source name and or id to emails missing both

auumgn:


auumgn:
[Some self asserted email addresses are not showing a source](https://trello.com/c/qiIq8UrO/9460-some-self-asserted-email-addresses-are-not-showing-a-source)

amontenegro:
Hey @auumgn please add some tests for the PublicRecordController. Only tests that have to do with your changes, and, add a TODO comment to remember we need to add more tests there.
Thanks

amontenegro (approved):

PR #7122, opened 2024-11-05
Papi limit filter NPE when no Authorization header

Camelia-Orcid:


Camelia-Orcid:
[[QA] Null pointer ex on api rate limiting filter](https://trello.com/c/wIqw5dDI/9470-qa-null-pointer-ex-on-api-rate-limiting-filter)

amontenegro (approved):

PR #7123, opened 2024-11-05
add null check when fetching emails

auumgn:


auumgn:
[[QA] 500 error response from public-record.json](https://trello.com/c/K2QW4e7L/9468-qa-500-error-response-from-public-recordjson)

PR #7139, opened 2024-11-15
Base uri property should be org.orcid.core.baseUri

amontenegro:


amontenegro:
https://trello.com/c/PHpVIhoG/70-prod-authorizing-email-addresses-from-mobile-devices-android-is-broken

PR #7143, opened 2024-11-20
Added the client Id white list

Camelia-Orcid:


Camelia-Orcid:
[[API Usage] - Implement whitelist by client id for the public API rate limiting](https://trello.com/c/y06mJXY8/9511-api-usage-implement-whitelist-by-client-id-for-the-public-api-rate-limiting)

orcid-pub-web/src/main/java/org/orcid/api/filters/ApiRateLimitFilter.java
```diff
@@ -98,6 +98,9 @@ public class ApiRateLimitFilter extends OncePerRequestFilter {
 
     @Value("${org.orcid.papi.rate.limit.ip.whiteSpaceSeparatedWhiteList:127.0.0.1}")
     private String papiWhiteSpaceSeparatedWhiteList;
+    
+    @Value("${org.orcid.papi.rate.limit.clientId.whiteSpaceSeparatedWhiteList}")
+    private String papiClientIdWhiteSpaceSeparatedWhiteList;
```
amontenegro, line 103: Lets make this a List and create a constructor that receive the whiteSpaceSeparatedWhiteList param and transform it into a list.

orcid-pub-web/src/main/java/org/orcid/api/filters/ApiRateLimitFilter.java
```diff
@@ -297,5 +302,18 @@ private boolean isWhiteListed(String ipAddress) {
         }
         return false;
     }
+    
+    private boolean isClientIdWhiteListed(String clientId) {
+        List<String> papiClientIdWhiteList = null;
+        if (StringUtils.isNotBlank(papiClientIdWhiteSpaceSeparatedWhiteList)) {
```
amontenegro, line 308: Lets change this so the list is generated on constructor time and here we only need to check if the id is in the list

amontenegro (changes requested):

orcid-pub-web/src/main/java/org/orcid/api/filters/ApiRateLimitFilter.java
Camelia-Orcid: Done in the afterProperties method that gets called just once in the lifecycle of the GenericFilterBean.

PR #7145, opened 2024-11-25
fix email source npe

auumgn:


auumgn:
[[QA] NPE returned from endpoints containing email data when record name is non-public](https://trello.com/c/H3d875ku/9515-qa-npe-returned-from-endpoints-containing-email-data-when-record-name-is-non-public)

amontenegro (approved):

PR #7151, opened 2024-12-09
If the source is null lets add one

amontenegro:


auumgn (approved):

amontenegro:
[[PROD] 500 error while reading records on 2.0 and 2.1](https://trello.com/c/NlC3jnsf/71-prod-500-error-while-reading-records-on-20-and-21)

PR #7152, opened 2024-12-13
Coding done, working on some unit tests

amontenegro:


amontenegro:
[Include membersupport@orcid.org as a cc for emails we send to PAPI clients that have exceeded the limit, update text and sender email](https://trello.com/c/WfJxBrwb/9536-include-membersupportorcidorg-as-a-cc-for-emails-we-send-to-papi-clients-that-have-exceeded-the-limit-update-text-and-sender-ema)

PR #7156, opened 2024-12-17
Map new v3 work types to v2 type other

auumgn:


auumgn:
https://trello.com/c/HCSBBu3W/9528-map-new-v3-work-types-to-other-in-v2

PR #7196, opened 2025-01-29
Update the cached token info so it indicate if the token is an OBO one

amontenegro:


amontenegro:
https://trello.com/c/QcpzXR7C/9608-prod-oboapiversioncheckfilter-is-not-using-the-token-cache

PR #7225, opened 2025-02-18
Allow token exchange if at least one token is not revoked

auumgn:


auumgn:
[[QA] Member OBO token exchange error](https://trello.com/c/SvHqqeoE/9642-qa-member-obo-token-exchange-error)

amontenegro (approved):

PR #7235, opened 2025-03-03
Fix switch user role

amontenegro:


amontenegro:
[[QA] 500 server response when using switch account functionality](https://trello.com/c/leftyWIt/9675-qa-500-server-response-when-using-switch-account-functionality)

leomendoza123 (approved):

PR #7239, opened 2025-03-05
Added the subdomain fix when processing email domains

Camelia-Orcid:


Camelia-Orcid:
[Investigate and fix why some emails addresses are flagged as professional but no email domain added](https://trello.com/c/DhhGt5ug/9664-investigate-and-fix-why-some-emails-addresses-are-flagged-as-professional-but-no-email-domain-added)

amontenegro (approved):

PR #7246, opened 2025-03-10
We dont need flash attributes, they were used only in the old UI

amontenegro:


amontenegro:
[[QA] Confirm deactivate - Using the email on a different account break the session](https://trello.com/c/NpCdHL6U/9689-qa-confirm-deactivate-using-the-email-on-a-different-account-break-the-session)

PR #7253, opened 2025-03-17
When counting the peer reviews summary, consider groups

amontenegro:


amontenegro:
[Incorrect count of peer review groups when a item is grouped](https://trello.com/c/BF7ziTgR/9658-incorrect-count-of-peer-review-groups-when-a-item-is-grouped)

leomendoza123 (approved):

PR #7254, opened 2025-03-18
rollback professional email source override

auumgn:


auumgn:
[Professional verified emails should be self asserted in API](https://trello.com/c/E8KGy7mr/9695-professional-verified-emails-should-be-self-asserted-in-api)

PR #7266, opened 2025-03-27
Add get interstitial flags endpoint

auumgn:


amontenegro (approved):

auumgn:
[Create an interstitial flag in the db so we know which interstitials have been shown to which users.](https://trello.com/c/CdNFAdLe/9706-create-an-interstitial-flag-in-the-db-so-we-know-which-interstitials-have-been-shown-to-which-users)

PR #7267, opened 2025-03-28
add path to mapi works endpoint

auumgn:


auumgn:
[[PROD] works endpoint does not contain the path with the member API](https://trello.com/c/Qn5emgAu/9700-prod-works-endpoint-does-not-contain-the-path-with-the-member-api)

amontenegro (approved):

PR #7274, opened 2025-04-03
On switch user, when the user comes from institutional sign in its au…

amontenegro:
…thority is a PreAuthenticatedAuthenticationToken


amontenegro:
[[PROD] 500 error on switch user](https://trello.com/c/5cMe0NoG/9731-prod-500-error-on-switch-user)

PR #7279, opened 2025-04-22
Add cache to email domain list

auumgn:


auumgn:
[[QA][Tech] Move the email_domains to a cache so they are not read from the DB on every call](https://trello.com/c/rnnBGdX6/9701-qatech-move-the-emaildomains-to-a-cache-so-they-are-not-read-from-the-db-on-every-call)

amontenegro (approved):

PR #7284, opened 2025-05-05
Remove the lowercase of the issn, the real issn number always in uppercase

Camelia-Orcid:
Apart of this change there is the need for data correction of the records in prod env. ( 16 entries in group_id_record and 312 in peer_review)


Camelia-Orcid:
[Peer review empty array](https://trello.com/c/VRn28s2D/9762-peer-review-empty-array)

amontenegro (approved):

PR #7307, opened 2025-06-03
Send notification for premium updater clients and also send notification if the last one created was ealier than 10 days ago

Camelia-Orcid:


Camelia-Orcid:
[[QA] Integration notification feature only available for premium member clients](https://trello.com/c/UyQIBfQi/9823-qa-integration-notification-feature-only-available-for-premium-member-clients)

PR #7314, opened 2025-06-08
Mvp notifications for UI fixes

Camelia-Orcid:
Fixes for:

1. Escape the single quote char so it it gets displayed after translation in english properties file
2. MVP notifications are not send to a user that has a token for the client id ( regardless if is valid or not)
3. Only 2 notifications are sent for the combination orcid + client id  , the second one after 10 days regardless if we have more profesional email domains for the orcid client id


Camelia-Orcid:
[[QA] Send integration notification a second time after 10 days](https://trello.com/c/bPmHPVjJ/9822-qa-send-integration-notification-a-second-time-after-10-days)

Camelia-Orcid:
[[QA] Send integration notification a second time after 10 days](https://trello.com/c/bPmHPVjJ/9822-qa-send-integration-notification-a-second-time-after-10-days)

amontenegro (approved):

PR #7315, opened 2025-06-10
Escape org solr search term

Camelia-Orcid:


Camelia-Orcid:
[[PROD] - SOLR escape all search terms in Profile and Org clients](https://trello.com/c/zSvgqoYR/9856-prod-solr-escape-all-search-terms-in-profile-and-org-clients)

amontenegro (approved):

PR #7333, opened 2025-07-09
Member name is used instead of client name for sending mvp

Camelia-Orcid:


Camelia-Orcid:
@leomendoza123 can you please review and merge?

PR #7339, opened 2025-07-28
City cannot be added without a country for the PA

Camelia-Orcid:


Camelia-Orcid:
https://app.clickup.com/t/9014437828/PD-1925  The clickup task.

PR #7341, opened 2025-07-30
${PD-1967} Do not throw exception if the revoked token doesnt have /activites/up…

amontenegro:
…date scope


amontenegro:
Task linked: [PD-1967 [Oauth] Id token exchange fails because of deactivated tokens](https://app.clickup.com/t/9014437828/PD-1967)

PR #7348, opened 2025-08-11
Revert "${PD-1970} getNewScopes was causing a NPE on the reused auth code request becaus…"

Austindgksdb:
Reverts ORCID/ORCID-Source#7347


amontenegro:
Task linked: [PD-1970 Reusing auth code returns 500 server error](https://app.clickup.com/t/9014437828/PD-1970)

PR #7405, opened 2025-12-09
PD-3704

cryptalith:


orcid-persistence/src/main/java/org/orcid/persistence/dao/impl/OrgAffiliationRelationDaoImpl.java
```diff
@@ -443,6 +443,27 @@ public void revertUserOBODetails(List<BigInteger> ids) {
         query.executeUpdate();
     }
 
+    @Override
+    @Transactional
+    @UpdateProfileLastModifiedAndIndexingStatus
+    public void clearFeatured(String orcid) {
+        Query query = entityManager.createNativeQuery("UPDATE org_affiliation_relation SET featured = NULL, last_modified = now() WHERE orcid = :orcid");
```
amontenegro: @cryptalith this is wrong, you cannot update the last modified on all the affiliations when one is not featured, you can include an extra condition in the where "and featured is not null" or something like that.

amontenegro:
Task linked: [PD-3704 Highlight affiliation in record header](https://app.clickup.com/t/9014437828/PD-3704)

PR #7412, opened 2025-12-17
${PD-3891} Manually check if the client have user obo enabled

amontenegro:


amontenegro:
Task linked: [PD-3891 Server responds 403 Forbidden when clients with obo enabled try to execute a PUT to an item they are in fact the source of](https://app.clickup.com/t/9014437828/PD-3891)

PR #7426, opened 2026-01-09
feat/simple-ping-healthcheck-mapi

gilesw:
.


gilesw:
I don't have a working docker setup for the mapi to test this on my machine but this is a clone of the papi /ping work that is now live.

amontenegro (approved):

PR #7431, opened 2026-01-15
${PD-3845} update email list change message logic

aromanovv:
send on adding new (unverified) emails
do not send on verification
do not send on deprecation


amontenegro:
@cryptalith this is ready to go to QA, please confirm with Will when we can deploy there before merging it.
Thanks

cryptalith (approved):

PR #7432, opened 2026-01-22
PD-3874  Null check for sourcePath

Camelia-Orcid:


amontenegro:
Task linked: [PD-3874 [PROD] Source validation on works V2.0 and V2.1 cannot handle null sources](https://app.clickup.com/t/9014437828/PD-3874)

PR #7437, opened 2026-01-28
${PD-4761} General improvements for work summaries rendering

amontenegro:


orcid-core/src/main/java/org/orcid/core/adapter/jsonidentifier/converter/JSONWorkExternalIdentifiersConverterV3.java
```diff
@@ -23,12 +23,15 @@
...
-    private LocaleManager localeManager;
-    
+    private final PIDNormalizationService norm;
+    private final String normalizationFailedErrorCode;
+    private final String normalizationFailedErrorDescription;
+
     public JSONWorkExternalIdentifiersConverterV3(PIDNormalizationService norm, PIDResolverService resolverService, LocaleManager localeManager) {
         this.norm = norm;
-        this.localeManager = localeManager;
+        // API errors are not localized
```
amontenegro: The API errors are not translated and it looks like they never will, so, it will save some time if we fetch the two strings from the localeManager on init, instead of having to call the localeManager everytime there is an error.

Camelia-Orcid (approved):

PR #7499, opened 2026-04-10
${PD-5409} Dummy liquibase as the update have already been executed in all the envs

amontenegro:


amontenegro:
Task linked: [PD-5409 [Oauth] Increase the attributes lenght to 8000 characters](https://app.clickup.com/t/9014437828/PD-5409)

PR #7513, opened 2026-05-11
${PD-5579} Set the max number of allowed works per page, if that is exceeded, th…

amontenegro:
…en we return an empty page


cryptalith (approved):

amontenegro:
Task linked: [PD-5579 Limit the max number of works that can be return per page in the public page](https://app.clickup.com/t/9014437828/PD-5579)

PR #7519, opened 2026-05-18
PD-5602 Remove token from cache if the user revoke permissions

amontenegro:


amontenegro:
Task linked: [PD-5602 Token stays cached after user revokes access](https://app.clickup.com/t/9014437828/PD-5602)

cryptalith (approved):

PR #7520, opened 2026-05-19
Fix the italic

amontenegro:


amontenegro:
Task linked: [PD-5602 Token stays cached after user revokes access](https://app.clickup.com/t/9014437828/PD-5602)

PR #7534, opened 2026-06-11
Unit tests

amontenegro:


amontenegro:
Task linked: [PD-5689 The source is showing incorrectly for works added by a client with User OBO](https://app.clickup.com/t/9014437828/PD-5689)

PR #7577, opened 2026-07-08
PD-5891 Add a RequestRejectedHandler so instead of generating a 500 it return…

amontenegro:
…s a 404


amontenegro:
Task linked: [PD-5891 Investigate what is causing the drop in the number of indexed pages in google](https://app.clickup.com/t/9014437828/PD-5891)

PR #7596, opened 2026-07-28
PD-6015 Changed it to return the token

Camelia-Orcid:
The /csrf.json endpoint was returning an empty map, so Cypress tests had no way to extract the CSRF token value to include in POST request headers.


amontenegro:
Task linked: [PD-6015 Various cypress tests that pass manually are failing in cypress](https://app.clickup.com/t/9014437828/PD-6015)

PR #7614, opened 2026-08-11
PD-6102 Fix for JDK 17+ forbids objectFieldOffset on hidden classes →throws UnsupportedOperationException

Camelia-Orcid:
Both MapperFacadeFactory files now create the Orika factory with .dumpStateOnException(false) Now, when a mapping fails, the real MappingException (with the actual mapping error as its cause) will propagate instead.


amontenegro:
Task linked: [PD-6102 java.lang.UnsupportedOperationException Full validation error: can't get field offset on a hidden class: private final java.lang.String org.apache.logging.log4j.core.pattern.ThrowablePatternConverter$$Lambda/0x000073057c686000.arg$1](https://app.clickup.com/t/9014437828/PD-6102)

amontenegro (approved):

PR #7632, opened 2026-08-17
PD-6145 upgrade to MapStruct

Camelia-Orcid:


amontenegro:
Task linked: [PD-6145 Orika and jdk21 performance degradation](https://app.clickup.com/t/9014437828/PD-6145)

orcid-activities-indexer/src/main/resources/orcid-activities-indexer-web-context.xml
```diff
@@ -38,7 +38,9 @@
 	
     <context:property-placeholder location="${org.orcid.message-listener.properties}" />
     
-	<context:component-scan base-package="org.orcid" />
+	<context:component-scan base-package="org.orcid">
+		<context:exclude-filter type="regex" expression=".*AdapterImplImpl" />
+	</context:component-scan>
```
amontenegro, line 43: Is this ok? AdapterImplImpl?

Camelia-Orcid: Yes it is. MapStruct generates its implementations at compile-time using Java's Pluggable Annotation Processing API (JSR 269). It does not use reflection at runtime.

We have for example public abstract class JpaJaxbEmailAdapterImpl implements JpaJaxbEmailAdapter ( the mapstruct Impl of this adapter)   MapStruct's default naming convention is to take the exact name of the @Mapper annotated class and append Impl to it and doesn't need to be scanned here

orcid-core/src/main/resources/ehcache_default.xml
```diff
@@ -24,7 +24,7 @@
             http://www.ehcache.org/v3 http://www.ehcache.org/schema/ehcache-core-3.1.xsd
             http://www.ehcache.org/v3/jsr107 http://www.ehcache.org/schema/ehcache-107-ext-3.1.xsd">
         
-    <ehcache:persistence directory="${java.io.tmpdir}/ehcache/default/annotations" />
+    <ehcache:persistence directory="${org.orcid.ehcache.dir}/default/annotations" />
```
amontenegro: `java.io.tmpdir` is set by java VM `org.orcid.ehcache.dir`  is not, so, we might need to add it as a vm argument in all machines?

amontenegro (approved):

PR #7649, opened 2026-08-21
PD-6180 Resolve PubMed identifiers via EuropePMC instead of the NCBI landing page

cryptalith:
## Summary

PubMed identifiers (PMID/PMC) are rejected as invalid when adding works, in both QA and Sandbox — surfaced by the Cypress regression run (Qase 44, 45, 78, 86).

Root cause: `PubMedResolver` gated every resolution on a `HEAD` request to `https://pubmed.ncbi.nlm.nih.gov/<id>` returning exactly HTTP 200. NCBI now fronts that page with a cookie/bot challenge that answers **HTTP 203** to non-browser clients regardless of User-Agent, so the check always failed — before EuropePMC, the actual metadata source (which still resolves these identifiers correctly), was ever called. This is an external change, not a regression in this repo: no ORCID commit to the resolver path is implicated.

## Changes

- `PubMedResolver` now uses EuropePMC as the sole authority for whether an identifier exists (`resolve()` and `resolveMetadata()`), and never fetches the NCBI landing page.
- Existence is cached 20 minutes, matching prior behavior, but a failing lookup is no longer cached (a throwing loader isn't cached by Guava), so a transient EuropePMC hiccup doesn't stick for the full window.
- `resolveMetadata()` now returns `null` when EuropePMC has no results, instead of a `WorkExtended` populated with only a work type.
- The previously-silent `UnexpectedResponseCodeException` catch, and other failure paths, are now logged.

## Test plan

- [x] `PubMedResolverTest` extended with regression coverage; new tests were confirmed to fail against the pre-fix resolver before the fix was applied.
- [x] `mvn test -pl orcid-core -Dtest=PubMedResolverTest,DOIResolverTest,ArXivResolverTest,PIDResolverServiceTest` — 17/17 pass.
- [x] Live-verified against real EuropePMC using the three PMIDs from the ticket (10848626, 37098064, 10848627) — all resolve with correct title/contributors; an unknown PMID still correctly reports not-found.
- [ ] Once deployed to QA, run the four affected Cypress specs in `orcid-cypress_tests-private/cypress/e2e/contributors/` (they exercise the real resolver, no stubbing).

### DIFF
--- a/orcid-core/src/main/java/org/orcid/core/utils/v3/identifiers/resolvers/PubMedResolver.java
+++ b/orcid-core/src/main/java/org/orcid/core/utils/v3/identifiers/resolvers/PubMedResolver.java
@@ -13,6 +13,8 @@ import java.util.Calendar;
 import java.util.Date;
 import java.util.List;
 import java.util.Locale;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 
 import jakarta.annotation.Resource;
 import jakarta.ws.rs.core.MediaType;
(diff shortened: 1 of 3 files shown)


amontenegro:
Task linked: [PD-6180 Pubmed identifiers rejected as invalid when adding works](https://app.clickup.com/t/9014437828/PD-6180)

amontenegro (approved):

PR #7651, opened 2026-08-22
PD-6181 Send password reset link to all verified emails on the account

cryptalith:
Sends the password reset link to every verified email address on the account, not only the address typed into the form ([PD-6181](https://orcid.clickup.com/t/9014437828/PD-6181)).

## What changed

`RecordEmailSender.sendPasswordResetEmail` now builds a recipient set and sends one message per recipient. Everything else in the flow is untouched — `PasswordResetController` keeps its param whitelist, email resolution, three-way branch and response shape; only its log wording changed.

- **Recipients** = the submitted address ∪ every verified address on the record, de-duplicated on a normalized key (`filterEmailAddress` + lowercase), since address lookup is case insensitive and would otherwise send two copies. The stored form of the address is what gets mailed, not the casing the user typed. The submitted address is ordered first.
- **Token handling is unchanged and deliberately stays ahead of the loop.** One JWT and one Redis entry per request, written before any send, so the link is redeemable when the first message lands.
- **Per-recipient body.** The template variable `submittedEmail` is renamed `recipientEmail` in `reset_password_email.ftl` / `_html.ftl`, so the line *"The ORCID iD registered to your address …"* is true in every copy and one address is not disclosed to another. Both templates have exactly one caller.
- **One send per recipient, never a multi-address `to`** — that would put the record's whole verified email list in the header of every copy.
- **Failure handling.** `MailGunManager.sendEmail` returns `false` rather than throwing on a non-200 or a QA/sandbox filter miss, so the send is wrapped per recipient and the loop continues; the endpoint still returns 200. Template rendering stays *outside* that catch — a broken template fails for every recipient alike, and swallowing it would turn "nobody got a link" into a warning nobody reads.
- **Logging** is keyed on the orcid and recipient counts, never the address list, to avoid dumping a record's whole email set into application logs on every request.
- **Recipient cap**, `org.orcid.core.mail.passwordReset.maxRecipients` (default 25), warns when it bites. Fan-out multiplies outbound mail per request against a Cloudflare budget of 30 requests / 600s per IP on `/reset-password.json`.

### DIFF
--- a/orcid-web/src/main/java/org/orcid/frontend/email/RecordEmailSender.java
+++ b/orcid-web/src/main/java/org/orcid/frontend/email/RecordEmailSender.java
@@ -43,11 +43,19 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.MessageSource;
 import org.springframework.context.NoSuchMessageException;
 import org.springframework.stereotype.Component;
+import org.springframework.web.util.HtmlUtils;
 
 @Component
 public class RecordEmailSender {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(RecordEmailSender.class);
+
+    /**
+     * Literal text in reset_password_email.ftl / _html.ftl. The reset templates are rendered once
+     * per request and this token is swapped per recipient, rather than re-rendering N times for a
+     * single differing value.
+     */
+    private static final String RECIPIENT_EMAIL_PLACEHOLDER = "{recipientEmail}";
 
     @Value("${org.orcid.core.mail.apiRecordCreationEmailEnabled:true}")
     private boolean apiRecordCreationEmailEnabled;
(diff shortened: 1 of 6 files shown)


amontenegro:
Task linked: [PD-6181 Send password reset emails to all verified email addresses](https://app.clickup.com/t/9014437828/PD-6181)

orcid-web/src/main/java/org/orcid/frontend/email/RecordEmailSender.java
```diff
@@ -279,10 +288,71 @@ public void sendPasswordResetEmail(String submittedEmail, String userOrcid) {
...
+        int accepted = 0;
+        int failed = 0;
+        for (String recipient : recipients) {
+            // One send per recipient on purpose. Putting every address in a single "to" would
+            // disclose the record's whole verified email list to each of them.
+            templateParams.put("recipientEmail", recipient);
+            // Rendering stays outside the catch below. A broken template fails for every
+            // recipient alike, so swallowing it here would turn "nobody got a link" into a
+            // warning nobody reads.
+            String body = templateManager.processTemplate("reset_password_email.ftl", templateParams);
```
amontenegro: Why dont we generate the body just one time? outside the for loop, and you leave a placeholder {recipientEmail} that gets set at this point of the for loop.

orcid-web/src/main/java/org/orcid/frontend/email/RecordEmailSender.java
```diff
@@ -279,10 +288,71 @@ public void sendPasswordResetEmail(String submittedEmail, String userOrcid) {
...
+            String htmlBody = templateManager.processTemplate("reset_password_email_html.ftl", templateParams);
```
amontenegro: Same with the HTML Body, lets generate it just one time

amontenegro (changes requested):

orcid-web/src/main/java/org/orcid/frontend/email/RecordEmailSender.java
cryptalith: Done in 2796a88879 — both templates are now rendered once per request, with `{recipientEmail}` left as literal text in the `.ftl` and substituted per recipient in the loop.

One wrinkle worth flagging: `reset_password_email_html.ftl` renders inside `<#escape x as x?html>`, and the substitution happens *after* FreeMarker has run, so the address would have been the one unescaped value in the document. The HTML copy substitutes `HtmlUtils.htmlEscape(recipient)`; the plain-text copy substitutes the raw address.

Two tests cover it: `testResetEmail_rendersEachTemplateOnceForAllRecipients` spies the `TemplateManager` and asserts one `processTemplate` call per template across three recipients, and `testResetEmail_escapesRecipientAddressInHtmlBody` uses an address containing `&`. Both confirmed red against the previous code.

orcid-web/src/main/java/org/orcid/frontend/email/RecordEmailSender.java
cryptalith: Same change — the HTML template is rendered once alongside the text one, outside the loop. See the reply on the comment above for the escaping detail, since substituting after `<#escape x as x?html>` means the address has to be escaped explicitly for the HTML copy.

cryptalith:
Pushed 2796a88879 addressing the review: both reset templates are rendered once per request and `{recipientEmail}` is substituted per recipient, instead of re-rendering N times for a single differing value.

Note the HTML copy escapes the address on substitution (`HtmlUtils.htmlEscape`), because the placeholder is filled in after FreeMarker's `<#escape x as x?html>` has already run — without that it would be the only unescaped value in the document. Covered by a new test using an address containing `&`.

`RecordEmailSenderTest` 29 and `PasswordResetControllerTest` 26 pass — 55 total. Ready for another look.

amontenegro (approved):

PR #7670, opened 2026-09-04
Set the @Autowired property to the normalization service

amontenegro:


amontenegro:
https://orcid.clickup.com/t/9014437828/PD-6145

PR #7676, opened 2026-09-07
PD-13369 carry Liquibase's history in the database seed

cryptalith:


gilesw (changes requested):
Can you dump the liquibase into a separate file like a 9-liquibase.sql so it's clear what's going on.

gilesw (approved):

PR #7685, opened 2026-09-08
PD-6145 per element source

cryptalith:


orcid-core/src/main/resources/ehcache_default.xml
```diff
@@ -386,7 +390,7 @@
     </ehcache:cache>
     
     <ehcache:cache-template name="defaultTemplate">
-        <ehcache:value-type copier="org.ehcache.impl.copy.IdentityCopier"/>
+        <ehcache:value-type copier="org.ehcache.impl.copy.SerializingCopier"/>
```
Camelia-Orcid: we need here IdentityCopier if not the performance suffer

orcid-core/src/main/resources/ehcache_orcid-api-web.xml
```diff
@@ -229,7 +233,7 @@
     </ehcache:cache>
     
     <ehcache:cache-template name="defaultTemplate">
-        <ehcache:value-type copier="org.ehcache.impl.copy.IdentityCopier"/>
+        <ehcache:value-type copier="org.ehcache.impl.copy.SerializingCopier"/>
```
Camelia-Orcid: We need to keep the IdentiyCopier not to have perf issues

amontenegro (approved):

PR #7692, opened 2026-09-10
PD-13463 fixed so readonly transactions read from readonlydb

Camelia-Orcid:
Implemented the remediation plan for J21-003 to resolve connection misrouting and advisor precedence issues in the persistence and caching layers. Verified that all DAO read operations route to transactionManagerReadOnly and that cache hits are resolved prior to opening transaction boundaries. Changes
Configured advisor ordering in Spring XML contexts by setting <cache:annotation-driven order="1"/> in orcid-core-cache-config.xml and <tx:annotation-driven order="2" proxy-target-class="true"/> in orcid-persistence-context.xml to ensure cache advice executes outermost. Updated @Transactional(readOnly = true) annotations across all DAO interfaces and implementations in orcid-persistence to explicitly specify @Transactional(value = "transactionManagerReadOnly", readOnly = true). Configured read-only test and development properties (test-db.properties and development.properties) to use simpleDataSourceReadOnly / dedicated connection pool properties.

### DIFF
--- a/orcid-core/src/main/resources/orcid-core-cache-config.xml
+++ b/orcid-core/src/main/resources/orcid-core-cache-config.xml
@@ -7,7 +7,7 @@
         http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context.xsd
         http://www.springframework.org/schema/cache http://www.springframework.org/schema/cache/spring-cache.xsd">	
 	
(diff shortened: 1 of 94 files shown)


amontenegro (approved):

Camelia-Orcid:
https://app.clickup.com/t/9014437828/PD-13463

PR #7693, opened 2026-09-10
Remove empty fields from funding after mapping

amontenegro:


amontenegro:
https://orcid.clickup.com/t/9014437828/PD-13465

Camelia-Orcid (approved):

PR #7702, opened 2026-09-11
J21-004 keep the boxed Boolean email flags across the v2 conversion

cryptalith:


orcid-core/src/main/java/org/orcid/core/version/impl/VersionConverterImplV2_0ToV2_1.java
```diff
@@ -72,16 +73,20 @@ private void copyInto(Object source, Object target, boolean downgrade) {
         try {
             BeanInfo beanInfo = Introspector.getBeanInfo(source.getClass());
             for (PropertyDescriptor pd : beanInfo.getPropertyDescriptors()) {
-                if ("class".equals(pd.getName()) || pd.getReadMethod() == null) {
+                if ("class".equals(pd.getName())) {
```
amontenegro, line 76: I dont like this approach, it only hides the problem and could cause other issues in the code because it might be expecting a function `isBoolean()` that doesn't exists anymore. Is there a way to address this without the need of reflection?

amontenegro (changes requested):

orcid-core/src/main/java/org/orcid/core/version/impl/VersionConverterImplV2_0ToV2_1.java
cryptalith: Will be address by https://orcid.clickup.com/t/9014437828/PD-13512

PR #7705, opened 2026-09-11
Error response from /funding endpoint for GET XML request where source is user

Camelia-Orcid:
Fixed HTTP 500 AccessorException: Object must have some value in its @XmlValue field when accessing funding endpoints (e.g., /v3.0/{orcid}/funding/{putCode}).

Prevented MapStruct from generating empty JAXB objects for optional @XmlValue fields when the underlying database entity attributes are null.


amontenegro (approved):

Camelia-Orcid:
https://orcid.clickup.com/t/9014437828/PD-13496

PR #7706, opened 2026-09-11
PD-13495 Fixed null mappings for contributors V2 and V3 both works and fundings

Camelia-Orcid:
Resolved the issue where contributor in works and fundings was missing its host value when serialized or returned via API endpoints.

Fixed null/empty @XmlValue field issues in contributor mappers to ensure contributors serialize cleanly to XML without AccessorException.


Camelia-Orcid:
https://orcid.clickup.com/t/9014437828/PD-13495

amontenegro (approved):

PR #7713, opened 2026-09-14
PD-13531 When the org have a null country, which is invalid, but some orgs in …

amontenegro:
…prod have it, we used to return an empty address in the org element


cryptalith (approved):

amontenegro:
https://orcid.clickup.com/t/9014437828/PD-13531

PR #7719, opened 2026-09-15
Set the transactions back to the manager level for activities

amontenegro:


Camelia-Orcid:
https://orcid.clickup.com/t/9014437828/PD-13532

PR #7721, opened 2026-09-16
PD-13566 Configuration updated to use readonly manager for the cache

Camelia-Orcid:
In orcid-core-context.xml, profileEntityGenericCacheManager is wired to profileEntityManagerReadOnlyV3.

Calls to profileEntityGenericCacheManager.retrieve(...) route through: profileEntityManagerReadOnlyV3 $\rightarrow$ profileLastModifiedAspectReadOnly $\rightarrow$ profileLastModifiedDaoReadOnly $\rightarrow$ entityManagerReadOnly.

Because profileLastModifiedDaoReadOnly holds entityManagerReadOnly, its @Transactional(value = "transactionManagerReadOnly", readOnly = true) annotation runs against the same read-only connection pool (pooledDataSourceReadOnly), resolving the dual-connection starvation issue on the primary write pool.


Camelia-Orcid:
https://orcid.clickup.com/t/9014437828/PD-13566

